PR #62, opened 2015-04-20
Fixed warnings.

natacha-beck:
Issue on Redmine: 5853


natacha-beck:
Pull request on aces:post_public_release instead.

PR #99, opened 2015-05-21
progress on 4.1.0

eafkhami:


BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -649,7 +649,34 @@ def get_type_list #:nodoc:
...
+    
+    if true
+      list = (check_role(:site_manager) || check_role(:admin_user)) ? DataProvider.descendants.map(&:name).delete_if {|elem| elem == "LocalDataProvider" }.sort_by {|elem| elem.downcase} : %w{ SshDataProvider }
+    else 
+      # this is the condition where all portals and bourreaux are on the same machine. check for this.
+      # It was suggested to ask the user if all portal and bourreaux are on the same machine and check that condition
+      (check_role(:site_manager) || check_role(:admin_user)) ? DataProvider.descendants.map(&:name) : %w{ SshDataProvider }     
+    end
+    
+    grouped_options = [
```
prioux: That all hardcoded and wrong! This will entirely fail when new data provider types are created... the sublisting must be done using more systematic categorization mechanism

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -649,7 +649,34 @@ def get_type_list #:nodoc:
...
     # where all portals and bourreaux are on the same machine.
-    (check_role(:site_manager) || check_role(:admin_user)) ? DataProvider.descendants.map(&:name) : %w{ SshDataProvider }
+    
+    # The above comment is not entirely accurate. Somethin about file systems being shared between two machines.
+    # Ask Tristan
+    
+    # to remove 'LocalDataProvider', we're adding conditional statement so if either of the two
+    # situations mentioned above arise, we can show the correct types. For now we remove 'LocalDataProvider' 
+    
+    if true
```
prioux: What is this "if true" there, and why is there a else that's never executed?!?

BrainPortal/app/models/data_provider.rb
```diff
@@ -999,6 +999,12 @@ def provider_full_path(userfile)
   def self.pretty_type #:nodoc:
     self.to_s
   end
+  
+  # this is a method that will return category type for this data_provider
```
prioux: Maybe returning nil in here, and explaining that returning nil is the convention to HIDE a data provider class from the interface, so that other programmers know this feature is available?

BrainPortal/app/models/en_cbrain_local_data_provider.rb
```diff
@@ -104,6 +104,11 @@ def impl_provider_rename(userfile,newname)  #:nodoc:
     return false unless FileUtils.move(oldpath.to_s,newpath.to_s)
     true
   end
+  
+  # this returns the category of the data provider -- used in view for admins
+  def self.pretty_category_name
+    "Enhanced CBRAIN Types"
```
prioux: Maybe all the strings returned should not contain "Types" at the end, but just the category name?

BrainPortal/app/models/local_data_provider.rb
```diff
@@ -38,6 +38,11 @@ class LocalDataProvider < DataProvider
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
 
+  # this returns the category of the data provider -- used in view for admins
```
prioux: Explain why nil in this particular file

BrainPortal/app/models/data_provider.rb
eafkhami: Added this comment

BrainPortal/app/models/en_cbrain_local_data_provider.rb
eafkhami: I think its clearer to return it with the type. It also makes for less work to add it later in the controller.

BrainPortal/app/models/local_data_provider.rb
eafkhami: Returning a nil is the convention that we'll use to HIDE a data provider class from the interface.
Thus, returning a NIL

prioux: I meant, explain it in the code, so that other programmer know about this convention! If it's not done, it should also be explained in the model in the base class, DataProvider.

eafkhami: Added the comments to the code

PR #140, opened 2015-06-17
Docker support

glatard:
To run a task in a Docker container, just add the name of the Docker image in the Tool configuration. The Docker image has to be stored on DockerHub (https://hub.docker.com). It assumes that a Docker service is running on the Bourreau. The behavior of tools with no Docker image configured remains unchanged.

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1630,13 +1630,11 @@ class ClusterTask < CbrainTask
   end
 
   def use_docker?
-     
     return (self.tool_config.docker_image and self.tool_config.docker_image != "")
   end
   
   # Returns the command line(s) associated with the task, wrapped in a Docker call if a Docker image has to be used.
   def docker_commands
-
     commands = self.cluster_commands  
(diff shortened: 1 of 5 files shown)


BrainPortal/app/models/cluster_task.rb
```diff
@@ -1470,7 +1470,23 @@ def submit_cluster_job
 
 # CbrainTask '#{self.name}' commands section
 
-#{commands.join("\n")}
+#{
```
prioux: This entire string interpolation is horrible, Tristan!

Please rewrite cleanly, or let me do it later this week.

glatard:
Better now?

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1628,7 +1629,27 @@ def update_size_of_cluster_workdir
     return nil
   end
 
-
+  def use_docker?
+    return (self.tool_config.docker_image and self.tool_config.docker_image != "")
```
prioux: OK but I'd have written more simply:

```ruby
self.tool_config.docker_image.present?
```

PR #151, opened 2015-06-18
Adding missing new lines to end of files

Jkat:
Correctly adding missing new lines to the end of files I found throughout the code base.


prioux:
Justin, why is it necessary to make a pull request for something like that?!?

Normally, such minor changes need only be integrated at the same time the files are being modified for something else...

Issue #13, opened 2015-04-09
 Organization of "Launch Task" drop-down menu.


natacha-beck:
The drop-down menu for "Launch Task" is feeling long and cluttered.

To reduce the visual clutter, "Conversion tools" could lead to a new drop down menu. I would also suggest removing from "Scientific tools" and adding to "Conversion tools":

mincaverage
mincmath
mincpik
mincresample
...

For "Scientific tools", group by tool. For example (in the below diagram, each arrow -> signifies a new drop-down menu):

"Scientific tools" -> "CIVET" -> CIVET, CIVET QC, CIVET Combiner
"Scientific tools" -> "FSL" -> Bedpostx, Fast, Feat, etc.

(Suggested by Lindsay)

natacha-beck:
Resolved by add a "Most Used Tools"

natacha-beck:
Resolve by the new UI

Issue #80, opened 2015-05-06
Replace filtration by filtering everywhere.


natacha-beck:
Like filtration is not a word it need to be replaced by filtering...

PR #162, opened 2015-07-14
Allow merge from offline Bourreau for new tc and more

### LINKED ISSUE #161: Allow merge from offline bourreau when create a tool_config
For the moment we cannot select a tool_config from a Bourreau that is offline when we want to create a new tool_config.

natacha-beck:
Closed #161


natacha-beck:
Close #13 and #80

PR #198, opened 2015-08-20
A couple of bug fixes

eafkhami:


BrainPortal/db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20150615174035) do
+ActiveRecord::Schema.define(:version => 20150730214453) do
```
prioux: why is there a schema change?!?

prioux:
Will be redone.

PR #205, opened 2015-09-09
Missing JSON descriptor loader

remibernard:
Similar to cbrain_task_class_loader.rb, this loader loads JSON
descriptors as CbrainTask classes. Missed from the last commits due to
the various .gitignores in the plugin directories.


remibernard:
Brain sticky note; wait for merge before nuking the branch.

PR #211, opened 2015-09-21
Scope API (new filtering system)

remibernard:
See commit 3d33b3d for additional details.

Note that this feature completely replaces the current filtering system and needs extensive testing (although a great deal has already been done).


remibernard:
Note that while the feature/system is more-or-less complete and ready to be merged, there is still a few quirks left to fix:

- Since the final attribute validation can only be performed on `apply`, invalid filters can possibly end up in the session and lock it up until an admin manually clears it. There are various ways to solve this.

- Some requests oddly sent pagination parameters even if they are not required. This makes `update_session_scopes` pointlessly create a new almost-empty `Scope` just for that parameter. This can be worked around by making `update_session_scopes` ignore these pagination parameters if the target `Scope` does not exist yet.

- The new persistent selection mechanism, designed to go hand-in-hand with scopes, is not implemented yet.

prioux:
Acknowledged, thanks Rémi!

prioux:
Bug report. The 'Browse DP' page has two selection box for the type of files or directories to be registered, but the content inside is in a random order; we have a helper already that creates such filetype selection boxes all indented in a tree, use this instead.

prioux:
Bug report: Also in the Browse DP page, as admin, when selection "Browse as another user" the list of files is not updated to reflect the content of the other user's directory.

prioux:
Bug: the filtering lists under the table column headers are unsorted, and also they have repeated counts (why three times instead of just the two original numbers?)

screen shot 2015-09-24 at 13 16 09

prioux:
Bug report: the DP browse page's 'Type' column only shows two types: 'Regular' and 'Directory'...

screen shot 2015-09-24 at 13 23 57

prioux:
Bug: tasks custom filters don't work at all, cause a 404 error when I select them in my menu.

screen shot 2015-09-24 at 16 24 51

remibernard:
I'll look at all of these issues tomorrow, but I can already point out a few things:
- For the browse DP selection boxes, I need to use the `master_select` helper to generate the lists.
  Unfortunately, I cannot use both helpers (`master` and the ordered one) at the same time.
- Repeated count; it was designed to look as much as possible like the old filters. I'll change it
  to the format you suggested.
- DP browse type column; filtering/sorting of the DP browse table is a brand-new feature possible due
  to the new filtering system, and the type column has to correspond to the type of directory entry
  for the filter to work properly. If you want it to work on both the type and registered Userfile class,
  it'll require a custom filter I'm afraid.

Also, one thing I noticed in your reports is that filter lists seem unsorted (in general). I'll check if
that can be fixed as well.

prioux:
Thanks, no prob. I'm trying to test the new code by using it in my day to day life on the DEV instances, but I have to go back to the old server for many things when they new pages don't work. But no hurry, as long as it gets fixed for the official merge to 4.2.0...

Issue #190, opened 2015-08-12
Document show table helper (Redmine 2654)


prioux:
(Tarek, December 2011)

Yeah, I was a little lazy about this. Will properly document it for RDoc.

PR #224, opened 2015-09-25
Scope fixes following review and testing

remibernard:
This pull request fixes the following Scope issues:
- Broken 'as user' DP browsing
- Filter count format: (34/100) -> (of 100)
- Odd filter menu order (now somewhat ordered by label)
- Broken user & group ordering on bourreaux
- Odd 'register all files' select box order
- Broken custom filters for tasks
- Spurious scopes with only pagination parameters
- Invalid scope filters/ordering rules cannot be removed and
  make the target page inaccessible.

And fixes issue #190

### DIFF
--- a/BrainPortal/app/controllers/data_providers_controller.rb
+++ b/BrainPortal/app/controllers/data_providers_controller.rb
@@ -297,13 +297,16 @@ class DataProvidersController < ApplicationController
     })
 
     # Browsing as a different user? Make sure the target user is set.
-    @as_user   = current_user
+    @as_user = current_user
       .available_users
-      .where(:id => @scope.custom['as_user_id'] ||= (
-        params['as_user_id'] || current_user.id
+      .where(:id => (
+        params['as_user_id'] ||
+        @scope.custom['as_user_id'] ||
+        current_user.id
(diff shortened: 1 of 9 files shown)


prioux:
New bug, in task index page. The count numbers in the header columns are wrong. I have a page with a custom filter applied, which returns 32 tasks. 29 on mindstorm, 1 on brainstorm, 2 on mammouth:  see screenshot 1. The custom filter does not include anything about execution server filtering. But when I summon the filter menu by execution server, instead of seing a menu with zero everywhere and the numbers 29, 1 and 2 for the three execution servers, I see the counts in screenshot 2.

screen shot 2015-09-26 at 11 27 21

screen shot 2015-09-26 at 11 27 35

remibernard:
This is caused by the filter headers getting the wrong scope (before custom filters are applied). I'll fix it first thing Monday morning.

prioux:
Another bug. In the userfiles index page, I have a custom filter called "ReadyToArchive" which, in the old system, returns 7000 files spread all over. It filters by update date and file type. In the new systems, the filter returns 0 entries.

bug1

Issue #60, opened 2015-04-20
Add tool_config attribute for qsub options


natacha-beck:
A tools config should contain an attribute to store
alternative values for the qsub command-line options.
These would replace (?) the options specified at the
Bourreau level.

PR #233, opened 2015-09-30
Add tool_config attribute for qsub options 

natacha-beck:
Should close this one #60. 

Not sure if it's the good way to do that. I would like to have this features on production as soon as possible ? It will be useful to test docker.

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -139,7 +139,7 @@ class ClusterTask < CbrainTask
   # job on the cluster. The value returned by a
   # CbrainTask should be conservative and be reasonably
   # larger than the longest run expected, without being
-  # overly excessive. The default value used by
+  # overlay excessive. The default value used by
   # the framework is 24.hours
   def job_walltime_estimate
     24.hours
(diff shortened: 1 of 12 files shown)


prioux:
Comments:

- why is the new attribute called "tc_extra_qsub_args", instead of simply "extra_qsub_args"? it's not like we call other attributes tc_description, tc_version_name, tc_etc_etc ?

- the modifications to the Scir subclasses seem to imply that when a value is present, it will OVERRIDE the bourreau's options... why not just append them?

- the mods to cluster_task completely ignore the values for this new attribute for the other two tool_configs involved in configuring a task... (the bourreau-specific tc and the tool-specific tc

I'm not willing to merge this PR at all unless these design problems are solved.

natacha-beck:
Maybe this one can stay open waiting my modification. 

For your point: 

- Why :tc_extra_qsub_args because you have a :cms_extra_qsub_args so for me its clearer to know where I took the extra_qsub_args. Maybe I'm wrong. 

- For your second point, if I just append it what happen with option that are define in both place ? I will just overwrite this ? Remy suggestion is to split both string and combine them giving priority to the option present in the tc. 

- Am I wrong, but for me it's make totally non sens to have extra-qsub-option at the level of the tool-specific tc, your tools can be install on different type of cluster ?! for the bourreau-specific tc it will be same idea to have it define directly in the Bourreau itself no ?

BrainPortal/app/models/cluster_task.rb
```diff
@@ -139,7 +139,7 @@ def save_results
   # job on the cluster. The value returned by a
   # CbrainTask should be conservative and be reasonably
   # larger than the longest run expected, without being
-  # overly excessive. The default value used by
+  # overlay excessive. The default value used by
```
driusan: This comment change looks like a typo to me. "overlay excessive" doesn't make grammatical sense.

natacha-beck:
@dave you have right..

PR #234, opened 2015-10-02
Dual-count filters, undo_where

remibernard:
Final (hopefully) implementation of scoped/dual-count filters, to
restore the missing functionality from the old filtering system. This
new implementation also fixes two bugs related to filtering and
undo_where (as it is no longer used for filters and will be refactored
out in a later commit), and allows for future expansion by supporting
both Scope-based filtering (on one or more Scopes) and classical AR
filtering.

This PR also removes undo_where and refactors the only other location
it was in use by shuffling some filtering around.


natacha-beck:
Need to be tested on dev.

Issue #209, opened 2015-09-15
Tag controller should check user_id and group_id, when creating and updating.


prioux:
When updating or creating a tag, let's make sure these two attributes have values that are allowed, given the current user.

PR #243, opened 2015-10-21
Tag controller should check user_id and group_id

eafkhami:
Tag controller should check user_id and group_id when creating and updating. issue #209


eafkhami:
Pierre please ignore this request. I'll do another pull request with the fixed changes.

Issue #241, opened 2015-10-15
The site selection box is not sorted


prioux:
We have a helper to create site selection boxes, but the content is not sorted. Example: in the user edit page, when changing the site a user belongs to.

eafkhami:
After a chat with pierre we decided to add a helper method site_select to be used for site select boxes.

PR #244, opened 2015-10-22
Updated fixes for issue #209 and #241

eafkhami:
The site selection box is not sorted #241
Tag controller should check user_id and group_id, when creating and updating. #209

### DIFF
--- a/BrainPortal/app/controllers/tags_controller.rb
+++ b/BrainPortal/app/controllers/tags_controller.rb
@@ -96,8 +96,8 @@ class TagsController < ApplicationController
     user_id = params[:tag][:user_id]
     group_id = params[:tag][:group_id]
 
-    params[:tag][:user_id]  = current_user.id           if !current_user.available_users.raw_first_column(:id).include?(user_id)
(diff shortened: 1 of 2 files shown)


BrainPortal/app/controllers/tags_controller.rb
```diff
@@ -89,4 +88,18 @@ def destroy #:nodoc:
...
   end
+
+  private
+
+  # This method validates that the current user has access to the user and group being assigned to the created or updated tag
+  def validate_params #:nodoc:
+    user_id = params[:tag][:user_id]
+    group_id = params[:tag][:group_id]
+
+    params[:tag][:user_id]  = current_user.id           if !current_user.available_users.raw_first_column(:id).include?(user_id)
```
prioux: I think there might be a bug here. Can you check? The params[:tag][:user_id] might be a string like "23" while the check in the if statement might generate a list of numbers, e.g. [5,9,23,99]. In that case, it means your code ALWAYS crush the values send in the form...

Issue #250, opened 2015-11-04
More scope filtering bugs


prioux:
Something similar to bug #249 , when trying to filter the task index table by server:

<pre>
Internal error: Exception Caught; Exception: ActiveRecord::ConfigurationError

Params: {"_pag_scope_name"=>"tasks", "_scope_mode"=>"replace", "_scopes"=>{"tasks"=>"H4sIAGYyOlYAA9PV1eVKs+LSVUi0UkjKLy0qSk0sjc9M4VJQKLNSMIKIl1QWpEIEnJOKEjPzQhKLs62s3IpznFJTCvKLSyogykqLU4vgWk3MjbgAkeuPNFwAAAA"}, "page"=>"1", "action"=>"index", "controller"=>"tasks"}

Exception: ActiveRecord::ConfigurationError: Association named 'remote_resource' was not found on CbrainTask; perhaps you misspelled it?
</pre>

Issue #249, opened 2015-11-03
Bug in scope filtering


prioux:
When viewing the Bourreau index page, if there are active tasks on the bourreau a link with the number of task is shown; that link is supposed to take us to the task index page with two filters applied: one for the bourrreau ID and one for the status 'active'. An exception is generated however:

```
Internal error: Exception Caught; Exception: Psych::DisallowedClass 
Exception: Psych::DisallowedClass: Tried to load unspecified class: Symbol
```

The params requests, when clicking on the link, looks like this:

```ruby
{"_scope_mode"=>"replace", "_scopes"=>{"tasks"=>"H4sIAFMsOVYAA9PV1eVKs-LSVUi0UkjKLy0qSk0sjc9M4VJQKLNSMIKIF5cklpQWQ4SsEpNLMstSuQCVhq5COAAAAA"}}
```
Decoding the filter I find:

```bash
echo -n 'H4sIAFMsOVYAA9PV1eVKs+LSVUi0UkjKLy0qSk0sjc9M4VJQKLNSMIKIF5cklpQWQ4SsEpNLMstSuQCVhq5COAAAAAAA' | base64 -d  | gunzip -c
```

which yields:

```
---
f:
- a: bourreau_id
  v: 2
- a: status
  v: :active
```

The problem seems related to the ':active' symbol.

remibernard:
Adding `value = value.to_s if value.is_a?(Symbol)` before [scope_helper.rb:774](https://github.com/aces/cbrain/blob/master/BrainPortal/app/helpers/scope_helper.rb#L774) should fix this. I'm not a fan of special-casing classes like this, but `Symbol` is the odd one out in YAML serialization due to security issues before Ruby 2.2.

I'll test & push this fix to `dev` ASAP.

PR #251, opened 2015-11-04
Scope filtering bugfixes (issues #249 & #250)

remibernard:
- Forced Symbol -> string conversion in index_scope_filter (issue #249)
- Fixed incorrect association specification (issue #250)
- Refactored session data controller to use the correct update method
  (apply_changes)
- Removed selection concern from Scope API (will be soon replaced by new
  persistent selection module)

### DIFF
--- a/BrainPortal/app/controllers/session_data_controller.rb
+++ b/BrainPortal/app/controllers/session_data_controller.rb
@@ -43,12 +43,9 @@ class SessionDataController < ApplicationController
 
   # POST /session_data
   def update #:nodoc:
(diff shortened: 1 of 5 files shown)


prioux:
Super cool. Will test immediately.

prioux:
I'm still getting the error of bug #250 when trying to sort the task index by Bourreaux.

prioux:
What do you think of simple defining the association in the CbrainTask model:

```ruby
class CbrainTask
   belongs_to :remote_resource, :class_name => "Bourreau", :foreign_key => :bourreau_id
end
```

PR #261, opened 2015-11-11
Tentative mechanism to allow PSOM/Niak to submit worker tasks to CBRAIN.

glatard:
This is a tentative mechanism to allow PSOM/Niak to submit worker tasks to CBRAIN. Please comment if you think it should be implemented differently.


Bourreau/app/models/bourreau_worker.rb
```diff
@@ -276,6 +279,13 @@ def process_task(task) # when entering this methods task is a partial object, wi
 
       worker_log.debug "Updated #{task.bname_tid} to state #{new_status}"
 
+      # Mechanism for tasks to submit other tasks: tasks that were not in a status 
+      # in COMPLETED_STATUS or FAILED_STATUS at the previous iteration may submit new
+      # tasks provided that they have the can_submit_new_tasks
+      # property.
+      handle_tasks_submitted_by(task) if task.class.properties[:can_submit_new_tasks] &&
+                                         !(CbrainTask::COMPLETED_STATUS.include?(task.status) || CbrainTask::FAILED_STATUS.include?(task.status))
```
prioux: Not sure about the two sets of states here... can we do a check in the opposite direction (e.g. if state is among a precise set of states?)

Bourreau/app/models/bourreau_worker.rb
```diff
@@ -536,4 +546,120 @@ def check_for_tasks_stuck_in_ruby
...
+    new_task.params = Hash.new
+    new_task_hash["parameters"].each do |param|
+      new_task.params[param[0].to_sym] = param[1]
+    end
+
+    # Sets other task attributes
+    # In the future, we could allow to register results to another data provider
+    # or to submit the task to another bourreau. This would require to carefully
+    # check permissions of current_task.user
+    new_task.description              = new_task_hash["description"]
```
prioux: What happens if some fields are not set? e.g. no description? Is NIL acceptable in the task model? just checking.

Bourreau/app/models/bourreau_worker.rb
```diff
@@ -536,4 +546,120 @@ def check_for_tasks_stuck_in_ruby
...
+  # * current_task: The current task from which the new task is submitted.
+  #                 Used to set the user, bourreau id and results data provider id
+  #                 for the new task.
+  # 
+  def submit_task_from_string json_string,current_task
+
+    # Parses JSON string and checks format
+    validate_json_string(json_string) # Raises an exception if string is not valid
+    new_task_hash = JSON.parse(json_string)
+    worker_log.info("Submitting task #{new_task_hash}")
```
prioux: you're serializing the entire new_task_hash in the worker_log message?

Bourreau/app/models/bourreau_worker.rb
```diff
@@ -536,4 +546,120 @@ def check_for_tasks_stuck_in_ruby
...
+    
+    # Creates task
+    task_class_name = new_task_hash["tool-class"]
+    new_task        = CbrainTask.const_get(task_class_name).new # Raises an exception if tool class is not found
```
prioux: Not strict enough, you can do

```ruby
CbrainTask.const_get('Kernel')
```
and get any class completely outside the CbrainTask:: namespace ...

Bourreau/app/models/bourreau_worker.rb
```diff
@@ -536,4 +546,120 @@ def check_for_tasks_stuck_in_ruby
...
+    tool_config_id          = new_task_hash["tool-config-id"]
+    if tool_config_id.blank?
+      # Sets tool config as the first one we find for class task_class_name
+      tool        = Tool.where(:cbrain_task_class => "#{task_class_name}").first
+      tool_config = accessible_tool_configs.where(:tool_id => tool.id).first
+    else
+      tool_config = accessible_tool_configs.find(tool_config_id)
+    end
+    raise "Cannot find accessible tool config for class #{task_class_name}" if tool_config.blank?
+    new_task.tool_config = tool_config
```
prioux: I'd make sure the ToolConfig object is a proper one for new_task ?

Bourreau/app/models/bourreau_worker.rb
```diff
@@ -536,4 +546,120 @@ def check_for_tasks_stuck_in_ruby
...
+
+    # Sets task parameters
+    new_task.params = Hash.new
+    new_task_hash["parameters"].each do |param|
```
prioux: Would that support arbitrary complex params (like arrays of hash of arrays of numbers etc) which are normally supported in a CbrainTask? Or are we limited to just one level of key-value ?

Bourreau/app/models/bourreau_worker.rb
glatard: I think this is the best way to check: we want to make sure that all the statuses are checked except the terminal ones. We may check one or two statuses that shouldn't be (e.g. standby) but I feel it's better than possibly forgetting statuses that should be checked.

Bourreau/app/models/bourreau_worker.rb
glatard: Added a default value for description. If tool-class or parameters are not defined, then method validate_json_string will throw an exception.

Bourreau/app/models/bourreau_worker.rb
glatard: Changed this to print only the tool class.

Bourreau/app/models/bourreau_worker.rb
glatard: Fixed.

Bourreau/app/models/bourreau_worker.rb
glatard: Fixed now.

Bourreau/app/models/bourreau_worker.rb
glatard: A value that is a JSON object will be parsed as a hash. However, I'm not sure it will correctly work because the resulting hash will look like {"foo"=>"bar", "plop"=>"coin"} while a CBRAIN task would rather use symbols instead of strings {:foo => "bar", :plop => "coin"}. At the first key-value level, strings are converted to symbols in method submit_task_from_string: this may have to be improved to cover multiple levels of key-value pairs.

PR #270, opened 2015-12-09
In ScirPbs, added username to qstat command-line

glatard:


prioux:
For other classes:

scir_sge :   "qstat -xml -u USERNAME"
scir_moab : "showq --xml -u USERNAME"
scir_unix :    remove the "a" option everywhere:   "ps x -o pid,uid,state"

BrainPortal/lib/scir_unix.rb
```diff
@@ -35,11 +35,11 @@ def update_job_info_cache #:nodoc:
...
       ps_command = case CBRAIN::System_Uname
         when /Linux/i
-          "ps ax -o pid,uid,state"
+          "ps x -o pid,uid,state"
         when /Solaris/i
-          "ps ax -o pid,uid,state"  # not tested
+          "ps x -o pid,uid,state"  # not tested
         else
-          "ps ax -o pid,uid,state"  # not tested
+          "ps x -o pid,uid,state"  # not tested
```
glatard: Pierre, the ps command line for Linux, Solaris and other systems is identical. Shall we remove the case statement and keep only 1 command line?

prioux: No. For the F.u.t.u.r.e.

BrainPortal/lib/scir_pbs.rb
```diff
@@ -31,8 +31,20 @@ class ScirPbs < Scir
...
-      out, err = bash_this_and_capture_out_err("qstat -f")
-      raise "Cannot get output of 'qstat -f' ?!?" if out.blank? && ! err.blank?
+      qstat_command = "qstat -u #{CBRAIN::Rails_UserName.to_s.bash_escape} -f"
+      retry_count = 0
+      # In some cases, the standard output of the qstat command 
+      # is blank even though jobs are still running on the cluster. In this situation,
+      # all the active jobs will be marked completed while they are not. 
+      # To address this issue, we retry the qstat command up to 3 times with a 3-second interval
+      # as long as the output of qstat is blank.
+      while(retry_count < 3) do
```
prioux: This is a bit costly... but OK. It would be great if we could mark individual bourreaux as 'unreliable' and only those would retry; we'll do it in a future version.

PR #271, opened 2015-12-10
Tool selection dialog

eafkhami:
This feature adds tagging functionality to tool model. The tags are used in a tool selection dialog to filter and search when users are picking tools.


BrainPortal/app/models/tool.rb
```diff
@@ -78,8 +82,90 @@ def global_tool_config
     @global_tool_config_cache ||= ToolConfig.where( :tool_id => self.id, :bourreau_id => nil ).first
   end
 
+
+  # Return an array of the tags associated with this tool
+  # by +user+. Actually returns a ActiveRecord::Relation.
+  def get_tags_for_user(user)
```
prioux: leftover from old tag association stuff ?!?

BrainPortal/app/views/userfiles/_tools_interface.html.erb
```diff
@@ -0,0 +1,129 @@
...
+              <% @tools_to_pick.each do |tool| %>
+
+              <tr
+              <%tool.get_all_tags.each do |tool_tag|%>
+              data-<%= tool_tag %> = 'true'
+              <%end%>
+              >
+                <td id="tool_<%= tool.id %>">
+
+                  <%=
```
prioux: Why is this HTML constructed as a string and resubstituted this way? Isn't it easier on the HTML side of the template?

BrainPortal/app/views/userfiles/_tools_interface.js.erb
```diff
@@ -0,0 +1,165 @@
+$(document).ready(function() {
```
prioux: Missing header (copyright notice etc)

BrainPortal/app/models/tool.rb
```diff
@@ -78,8 +82,82 @@ def global_tool_config
...
+  end
+
+  # returns all tags associated with a tool
+  def get_all_tags
+    application_type + application_package_name + application_tags
+  end
+
   private
 
+  def url_present?
```
prioux: Missing doc for method, or add #:nodoc: if it shouldn't be documented

BrainPortal/db/migrate/20150924184353_add_url_to_tools.rb
```diff
@@ -0,0 +1,5 @@
+class AddUrlToTools < ActiveRecord::Migration
```
prioux: Copyright notice is missing again. Check other new files too.

PR #278, opened 2016-01-05
Added support for a 'cbrain:inherit-from-class' private attribute in Boutiques' JSON schema.

glatard:
This modification allows for some customization of tasks created with Boutiques. Feedback from @remibernard would be appreciated if he has time.

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
@@ -30,13 +30,8 @@
 % # NOTE: This template's weird indentation is there to try and make the
 % # generated code as legible as possible.
 
-% if descriptor['cbrain:inherits-from-class']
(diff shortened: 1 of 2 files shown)


remibernard:
While I can somewhat see the appeal of switching the base class when generating task classes for manual edition, isn't this exactly one of the reasons the generator was made the way it is now?

If the goal is to customize automatically generated tasks, then modifying the base class seems like an odd choice; maybe a way to reopen the class would be a cleaner design?

glatard:
I have n tasks that need to share the same method, which clearly requires inheritance. I could generate the n classes and edit them manually to make them derive from the right class, but that sounds like an overkill while adding this attribute is so simple. In general, and as you noticed initially, it is much more maintainable to rely only on the JSON descriptor than to generate the classes.

remibernard:
Using an utility class/module is not possible?

glatard:
It wouldn't be easy... Besides, all these tasks are different pipelines from the same framework, so it makes sense that they have a super class in common.

Issue #131, opened 2015-06-08
 Don't like the Userfile#update_file_type method.


natacha-beck:
It's clunky, couples the Userfile and User models, and doesn't use the new type validations, meaning we have type-checking logic in two different places. It's only used in one place, UserfilesController#update_multiple, so I think we should be able to get rid of it (and possibly other methods that only it depends on).

Redmine: 3211

prioux:
Cannot find code that was bad... Remi to investigate.

Issue #320, opened 2016-01-29
Do not make @base_scope available in the views


prioux:
We have to replace all occurrence of @base_scope in our view code so that instead we get it from some sort of method call.

The reason (and it took 2 hours for Remi and I to debug) is that if a view template crashes, all instance variables are `inspected` and this can trigger Rails to instanciate an enormous amount of records, causing instant memory freeze.

prioux:
A safety suggestion:

Add a modified (chained/overloaded) version of the method .inspect() to ActiveRecord::Base or ActiveRecord::Querying (whichever is appropriate) that will check for the numebr of records that are about to be inspected and not do it if it's too many. Something like

```ruby
class ActiveRecord::Something

  alias :orig_inspect :inspect

  def inspect(*args)
    if self.count > 200
      "[ Oops way too many record to show, or maybe show the first 5 or so ]"
    else
      orig_inspect(*args)
    end
  end

end
```

Issue #318, opened 2016-01-27
Issues in launch tool interface (UI dev).


natacha-beck:
This issue is an overall of what is spread into the issue #291 (should be more readable).

- [x] Tool search could be improved. For instance, typing "melodic" doesn't return "fslmelodic" (because "fslmelodic" doesn't start with "melodic"?)
- [x] It would be nice to be able to select a tool by clicking anywhere in the tool element, not just the tool's name.
- [x] The dialog's contents sometimes shows up at the bottom of the UI while the page is loading.
- [x] Resizing the dialog is apparently possible, and results in extra whitespace.
- [x] Nitpick; once a tool is selected, there is an odd amount of extra whitespace added at the end of the tool element.
- [x] Tags ("cluster", "diagnostics" and "life cycle "") look wrong. (Tristan)
- [x] Unclear what "Type ("internal")" and "Software package ("CBRAIN")" are. Checkbox for each category should only be visible if user can access at least one tool of this category. 
- [x] Put the checkbox before tags. In this way all the checkbox will be aligned!
- [x] Remove "All tools" or "Clear selection" both do the same stuff.

Take a look to the discussion about the order (alphabetic vs most used...), both acceptable.

Issue #156, opened 2015-07-03
Processing of read-only files


glatard:
Read-only files cannot be processed by any task.

Error message: "You must select at least one file to which you have write access."

Najmahan:
I JUST had the same problem dealing with another group project. This is not a very logical restriction, so must be a bug?

prioux:
What would help is if tasks would provide a hint as to whether they modify their input files or not. Assume "not".

So maybe a new task property. If the property exist we can trust it.

Right now, the only CbrainTask what modifies its input is CbrainTask::CivetQC (in package cbrain-plugins-neuro) which modifies a CivetStudy userfile.

PR #322, opened 2016-01-29
Development on 4.3.0 #1

remibernard:
Closes issues #131, #320, #156 and #318

### DIFF
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -208,7 +208,8 @@ class TasksController < ApplicationController
 
     # Filter list of files as provided by the get request
     file_ids = params[:file_ids] || []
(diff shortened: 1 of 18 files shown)


BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -208,7 +208,8 @@ def new #:nodoc:
 
     # Filter list of files as provided by the get request
     file_ids = params[:file_ids] || []
-    @files   = Userfile.find_accessible_by_user(file_ids, current_user, :access_requested => :write) rescue []
+    access   = @task.class.properties[:readonly_input_files] ? :read : :write
```
prioux: I don't like the name of the property... :readonly_input_files is not a good descriptor of what it means for the behavior of the task.

I think it shoudl be something like    :does_not_modify_input_files

Issue #335, opened 2016-02-18
Load info no longer shown in Bourreau index page tooltip


prioux:
There used to be load information shown in a tooltip that appeared when you hovered over the numbers in the 'tasks' column of the Bourreau index page. Now the tooltip appears but the numbers are always missing. See screenshot.

![screen shot 2016-02-18 at 15 18 26](https://cloud.githubusercontent.com/assets/777588/13156943/1f4b87de-d653-11e5-8934-e36444b1bb24.png)

remibernard:
As weird as it may seem, this is a very old bug. It appears to have been introduced with commit 9eefffdf10327db0ab6da6a425b786303772d246, which makes the `load_info` partial generating the tooltip rely on instance variables (`@num_active`, `@num_queued` and `@num_processing`) supplied by the corresponding [`load_info`](https://github.com/aces/cbrain/blob/9eefffdf10327db0ab6da6a425b786303772d246/BrainPortal/app/controllers/bourreaux_controller.rb#L210) controller action. As `load_info` is not invoked when the index page is generated, the corresponding instance variables are empty and so is the tooltip.

Since I'm a bit overwhelmed with work these days, I'll have to pass this on to @natacha-beck, the author of commit 9eefffdf10327db0ab6da6a425b786303772d246.

prioux:
No prob, I suspected as much. Don't worry about issues assigned to you, in general, it's more important for you to work on the rest. And I appreciate your hints and comments, every time.

Issue #330, opened 2016-02-05
Empty session with _simple_filters


remibernard:
There is a huge caveat when using the [`_simple_filters`](https://github.com/aces/cbrain/blob/3e9fe793f5d35ab2984bdc5d07cf2bc8c3c5e2ce/BrainPortal/lib/view_scopes.rb#L1199) scope query parameter. If the scope we wish to update with `_simple_filters` does not exist yet (with an empty session, for example) the filters will be placed in the default scope (`default_scope_name`). This is very rarely what is desired since this default is rarely used.

For example, a simple filtering URL like `/userfiles?_simple_filters=1&data_provider_id=1` will create a `userfiles#index` scope instead of the expected `userfiles` one (if there is no `userfiles` scope yet) and the specified filters wont be applied, as the controller expects its scope to be `userfiles`, not `userfiles#index`.

[view_scopes.rb:1235](https://github.com/aces/cbrain/blob/3e9fe793f5d35ab2984bdc5d07cf2bc8c3c5e2ce/BrainPortal/lib/view_scopes.rb#L1235):
```ruby
if (simple = params['_simple_filters'])
      # Determine which scope to update
      known      = (current_session['scopes'] ||= {})
      controller = params[:controller].to_s.downcase
      name   = simple     if known.has_key?(simple)
      name ||= controller if known.has_key?(controller)
      name ||= default_scope_name
# ...
```

Unfortunately, there is no way to know at that point in `update_session_scopes` where to place the given filters if the session does not contain either the parameter (`simple`) or the controller name (`controller`). This issue notably only concerns the first request made; on subsequent requests, the controller will have set up its scope and `name ||= controller if known.has_key?(controller)` will correctly assign the desired scope name.

Underlying cause of https://github.com/aces/cbrain-apis/issues/5

Issue #18, opened 2015-04-10
The 'delete files' button in the DP browse page seems to block.


prioux:
I think the 'delete files' in the 'browse' page does not start in background...

Note by Natacha:

Seems that no register/unregister/delete start in background, it will be a good idea to split the code into methods.

PR #345, opened 2016-03-01
Development on 4.3.0 #2

### LINKED ISSUE #337: Upload limit on dev?
In app/view/userfiles/_dialogs.html.erb, we have a nice TODO...:
 
`data-max-upload-size="" <%# TODO %>`

I found no upload limit elsewhere for the upload-size, am I right ? So in the Dev branch the user can upload files of any size ?

remibernard:
Closes #337, #330, #335, #18 and #349


BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,329 @@ def browse
...
-        params['as_user_id'] || current_user.id
-      ))
-      .first
-    @as_user ||= current_user
+    flash[:notice] ||= ''
+    api_request = [:json, :xml].include?(request.format.to_sym)
 
-    unless @provider.is_browsable?(current_user)
-      flash[:error] = "You cannot register files from this provider."
+    ensure_browsable(@provider, 'delete', proc do |html, data, status|
```
prioux: Is it really worth it to replace a "if" statement containing a respond block (the original code) with a method which takes a three parameter callback ? Repeating the original if statement in all three actions register, unregister and delete would have taken less code in total...

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,329 @@ def browse
...
+    # (the everyone project, under which no file is ever registered, or a
+    # project the user doesn't have access to).
+    group_id = current_user.own_group.id if (
+      group_id == Group.everyone.id ||
+      ! current_user.available_groups.map(&:id).include?(group_id)
+    )
+
+    # Register the given userfiles in background.
+    userfiles = userfiles_from_basenames(@provider, @as_user, params[:basenames])
+    userfiles_count = userfiles.count # Avoids a cute race condition
```
prioux: I suppose the race condition is when files are being moved in background ? :-)

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,329 @@ def browse
...
+
+    registered, already_registered = [], []
+    succeeded, failed = [], {}
+
+    CBRAIN.spawn_with_active_records_if(
+      api_request,
```
prioux: Hum, do I understand properly, we spawn ONLY if it's an API request? No when it's a web-based, interactive session?

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,329 @@ def browse
...
+      current_user,
+      "Register files (data_provider: #{@provider.id})"
+    ) do
+      userfiles.keys.each do |basename|
+        begin
+          # Is the file already registered?
+          unless userfiles[basename].blank?
```
prioux: Change to     if userfiles[basename].present?

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: And the `userfiles` variable is getting re-used before the count is fetched for the message, yep.

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: Looking at `ensure_browsable`, it basically avoids duplicating the message. The original idea was to make all the checks use the same die proc, but it turns out there's only one check. I'll refactor this ASAP.

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: Oops, you're right, that should be `! api_request`.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
-    registered_files = registered_files.where( :user_id => @as_user.id ) if ! @provider.allow_file_owner_change?
-    registered_files = registered_files.all.index_by(&:name)
+        rescue => e
+          (failed[e.message] ||= []) << basename
+        end
+      end
 
-    basenames.shuffle.each do |basename|
+      # Remind the user if browsing as another user
+      flash[:notice] += "Important note! Since you were browsing as user '#{@as_user.login}', the files were registered as belonging to that user instead of you!\n" if
```
prioux: I don't think it makes sense to modify the flash withing a CBRAIN.spawn_with_active_records... it won't work... I think.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+        succeeded.present? && @as_user != current_user
 
-      # Unregister files
+      # If files actually got registered, clear the browsing cache
+      clear_browse_provider_local_cache_file(@as_user, @provider) if
```
prioux: So this clear operation can happen in background at any time while the user is doing some more browsing within the interface? Could that cause a problem?

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
-        end
-        next
+      # Prepare to copy/move the files to the new DP
+      generic_notice_messages('register', succeeded, failed)
+      succeeded, failed = [], {}
+
+      # Will some of the file names collide?
+      collisions = Userfile.where(
+        :name             => registered.map(&:name),
+        :user_id          => current_user.id,
```
prioux: Bug! Should be `@as_user.id` instead of `current_user.id` (and yes the bug was already there, not your fault)

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+        :data_provider_id => target_dp.id
+      ).map(&:name)
```
prioux: Change `map()` to `raw_first_column('userfiles.name')` for speed, especially if thousands of files involved.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+    # Fallback to the user's own project if the one selected above is invalid
+    # (the everyone project, under which no file is ever registered, or a
+    # project the user doesn't have access to).
+    group_id = current_user.own_group.id if (
+      group_id == Group.everyone.id ||
+      ! current_user.available_groups.map(&:id).include?(group_id)
+    )
+
+    # Register the given userfiles in background.
+    userfiles = userfiles_from_basenames(@provider, @as_user, params[:basenames])
```
prioux: I think the original name for this variable was misleading: `userfiles` seems to imply this is an array of userfiles, but it in fact it's an array of a weird FileInfo object meant to look like one of the structure in Net::SFTP::Protocol . NOT urgent, but we should think about just renaming it... `fileinfos`  maybe ?

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+    # Unless one was specified explicitly via :other_group_id
+    group_id = params[:other_group_id].to_i unless
+      params[:other_group_id].blank?
+
+    # Fallback to the user's own project if the one selected above is invalid
+    # (the everyone project, under which no file is ever registered, or a
+    # project the user doesn't have access to).
+    group_id = current_user.own_group.id if (
+      group_id == Group.everyone.id ||
+      ! current_user.available_groups.map(&:id).include?(group_id)
```
prioux: Use `raw_first_column('groups.id')` instead of `map(&:id)`

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
-      file_group_id ||= current_project.try(:id) || current_user.own_group.id
-      if file_group_id == Group.everyone.id ||          # We won't register a file in group 'everyone'
-         ! current_user.available_groups.where(:id => file_group_id).exists? # or that is not in one of our groups
-         file_group_id = current_user.own_group.id
-      end
+          when :copy
+            new = userfile.provider_copy_to_otherprovider(target_dp)
+            userfile.delete rescue true # Not destroy(), as the contents must be kept.
+            userfile.destroy_log rescue true
+            (userfile = new).set_size!
```
prioux: We have two `set_size!` (the other is three lines below)

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+    succeeded, failed = [], {}
+    erasing = params.has_key?(:delete)
+
+    CBRAIN.spawn_with_active_records_if(
+      [:html, :js].include?(request.format.to_sym),
+      current_user,
+      "Unregister files (data_provider: #{@provider.id})"
+    ) do
+      userfiles.reject { |b,u| u.blank? }.shuffle.each do |basename, userfile|
+        begin
```
prioux: Just before the begin block, insert a `$0 = ""` counter line so that the admin can monitor the progress when doing 'ps'. See above in 'register'.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+      format.json { render :xml  => api_response }
     end
+  end
 
-    if newly_registered_userfiles.size > 0
-      clear_browse_provider_local_cache_file(@as_user, @provider) unless request.format.to_sym == :xml || request.format.to_sym == :json
-      flash[:notice] += "Registered #{newly_registered_userfiles.size} files.\n"
-      if @as_user != current_user
-        flash[:notice] += "Important note! Since you were browsing as user '#{@as_user.login}', the files were registered as belonging to that user instead of you!\n"
+  # Unregister a list of files from CBRAIN
```
prioux: Document that this action can do two things:

* unregister a userfile while leaving the file on the external DP side
* unregister a userfile and delete the file on the external DP side

In both case we always act upon userfiles properly registered in CBRAIN.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+      clear_browse_provider_local_cache_file(@as_user, @provider) if
+        erasing && succeeded.present? && [:html, :js].include?(request.format.to_sym)
 
-      respond_to do |format|
-        format.html { redirect_to :action => :browse }
-        format.xml  { render      :xml    => api_response }
-        format.json { render      :json   => api_response }
-      end
-      return
+      generic_notice_messages('unregister', succeeded, failed)
```
prioux: Outside the spawn block, succeeded will always be empty whenever we DO spawn.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+      format.json { render :xml  => api_response }
     end
-    to_operate = newly_registered_userfiles - collisions
+  end
 
-    if collisions.size > 0
-      flash[:error] += "Could not #{move_or_copy.downcase} some files, as files with the same names already exist:\n" +
-                         collisions.map(&:name).sort.join(", ")
+  # Directly delete a list of files from a browsable DP
+  def delete
```
prioux: Document; see comment for 'unregister'.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+            unless userfile.has_owner_access?(current_user)
+              (failed["Insufficient permissions"] ||= []) << basename
+              next
             end
-            success_list << u
-          rescue => ex
-            (failed_list[ex.message] ||= []) << u
+
+            result = userfile.destroy
+            userfile.destroy_log rescue true
```
prioux: Here the destroy_log() is always done anyway if destroy() succeeds. No need to do it explicitely.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -736,6 +801,86 @@ def get_type_list #:nodoc:
...
+        params['as_user_id'] || current_user.id
+      ))
+      .first
+    @as_user ||= current_user
+  end
+
+  # Fetch the userfiles corresponding to the given +basenames+ for
+  # +user+ on +provider+.
+  def userfiles_from_basenames(provider, user, basenames) #:nodoc:
+    userfiles = provider.userfiles.where(:name => basenames)
```
prioux: Are we not missing `where(:data_provider_id => provider.id)` ?

prioux:
@remibernard I've finished the code review; I've some concerns about using flash within the CBRAIN.spawn() calls, and proper notifications using these ugly lists (`succeeded`, `failed` etc) especially also within those spawn() calls. Aside from that, the rest of my comments are nitpicks.

You can work on this OR if you don't have time, I will merge this and make the changes myself. Your choice.

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: It doesn't make much sense indeed, and shouldn't work. I'll move it above the spawn. The only drawback is that a few checks will be skipped, but that shouldn't be a big issue.

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: Didn't we already have this kind of situation before anyway when a user is browsing while someone else is registering/deleting/unregistering files? Assuming the browsing actions automatically regenerate the cache, the only issue I can see is if both actions run exactly at the same time, a rather unlikely (famous last words) situation.

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: Using `provider` instead of `Userfile` implies this.

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: As far as the current code is concerned, `userfiles_from_basenames` returns a hash `{ <basename> => <userfile or nil> }`. As such `userfiles` seems like an appropriate variable name for it.

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: AFAIK the main reason for such a progress mechanism is that register sometimes has to move/copy files around. I'll add it to support the optional delete, but I'm pretty sure most unregisters will be too quick for it to really be of any use...

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: Thankfully, we're still in the spawn block at that point.

BrainPortal/app/controllers/data_providers_controller.rb
remibernard: It was not documented as most actions aren't. As such, I'm not sure how I should document this beyond
```ruby
# Unregister (and optionally delete) a list of files from CBRAIN
```

BrainPortal/app/controllers/data_providers_controller.rb
prioux: Duh. Of course.

BrainPortal/app/controllers/data_providers_controller.rb
prioux: Must have gotten my indentation wrong.

BrainPortal/app/controllers/data_providers_controller.rb
prioux: Hum, right, and I even thought about it in my first read through.

BrainPortal/app/controllers/data_providers_controller.rb
prioux: It's even less likely than that: the cache file has a name with the current user's ID in to, so each a user cannot even erase the cache of another. So basically, yeah, just leave it.

BrainPortal/app/controllers/data_providers_controller.rb
prioux: Or you could move it below the spawn, alongside other notifications.

BrainPortal/app/controllers/data_providers_controller.rb
prioux: I still dislike the name but it's not that important.

BrainPortal/app/controllers/data_providers_controller.rb
prioux: Still, I'd like before each action maybe 3 lines of comment instead of just one. Just to explain quickly what the action does, since there are subtleties.

prioux:
@remibernard I'm done commenting on your comments to my comments for the review of your code. :-)

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+    userfiles = userfiles_from_basenames(@provider, @as_user, params[:basenames])
+    succeeded, failed = [], {}
+    erasing = params.has_key?(:delete)
+
+    CBRAIN.spawn_with_active_records_if(
+      [:html, :js].include?(request.format.to_sym),
+      current_user,
+      "Unregister files (data_provider: #{@provider.id})"
+    ) do
+      userfiles.reject { |b,u| u.blank? }.shuffle.each do |basename, userfile|
```
prioux: Bug! There's no 'shuffle' method for a hash.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -399,263 +399,328 @@ def browse
...
+    # Erase the given userfiles in background.
+    userfiles = userfiles_from_basenames(@provider, @as_user, params[:basenames])
+    succeeded, failed = [], {}
+
+    CBRAIN.spawn_with_active_records_if(
+      [:html, :js].include?(request.format.to_sym),
+      current_user,
+      "Delete files (data_provider: #{@provider.id})"
+    ) do
+      userfiles.shuffle.each do |basename, userfile|
```
prioux: Same bug here, no shuffle() for hash tables.

BrainPortal/db/schema.rb
```diff
@@ -366,6 +366,7 @@
   add_index "userfiles", ["data_provider_id"], :name => "index_userfiles_on_data_provider_id"
   add_index "userfiles", ["group_id"], :name => "index_userfiles_on_group_id"
   add_index "userfiles", ["hidden", "id"], :name => "index_userfiles_on_hidden_and_id"
+  add_index "userfiles", ["hidden"], :name => "index_userfiles_on_hidden"
```
prioux: Did you add a migration?

prioux:
Thanks @remibernard ; I will test this a bit further today and I expect I'll merge this afternoon.

Issue #292, opened 2016-01-13
Look for an alternate email-based authentication (Persona will be decommissioned in Nov 2016). 


glatard:
See email below from the Persona mailing list. 
---
Hi Everyone,

When the Mozilla Identity team transitioned Persona to community
ownership, we committed resources to operational and security support
throughout 2014 [1], and renewed that commitment for 2015 [2].  Due to
low, declining usage, we are reallocating the project’s dedicated,
ongoing resources and will shut down the persona.org services that we run.

Persona.org and related domains will be taken offline on November 30th,
2016.

If you run a website that relies on Persona, you will need to implement
an alternative login solution for your users.  We have assembled a wiki
page with additional information and guidelines for migration [3], but
here are the important things you need to know:

Between now and November 30th, 2016, Mozilla will continue to support
the Persona service at a maintenance level:

 * Security issues will be resolved in a timely manner and the services
   will be kept online, but we do not expect to develop or deploy any
   new features.

 * Support will continue to be available on the dev-identity mailing
   list [4] and in the #services-dev IRC channel [5].

 * All websites that rely on persona.org will need to migrate to
   another means of authentication during this period.

Beginning on November 30th, 2016, the Persona service hosted by Mozilla
will be decommissioned:

  * All services hosted on the persona.org domain will be shut down.

  * Mozilla will retain control of the persona.org domain and will
    not transfer it to a third party.

  * Since the privacy of user data is of utmost importance to Mozilla,
    we will destroy all user data stored on the persona.org servers,
    and will not transfer it to third parties.

We intentionally designed Persona to expose email addresses rather than
opaque identifiers, which should ease the transition to other systems
that provide verified email addresses.  You can find guidelines on
alternative login solutions on the wiki [3] and we will continue to
update them over the coming year.  We strongly encourage affected teams
to openly discuss and blog about their migrations on the dev-identity
mailing list  so that others can learn from their experience.

Thank you for your support and involvement with Persona. If you have any
questions, please post them to the dev-identity mailing list.

  Ryan

[1]
http://identity.mozilla.com/post/78873831485/transitioning-persona-to-community-ownership
[2] https://groups.google.com/forum/#!topic/mozilla.dev.identity/rPIm7GxOeNU
[3]
https://wiki.mozilla.org/Identity/Persona_Shutdown_Guidelines_for_Reliers
[4] https://lists.mozilla.org/listinfo/dev-identity
[5] http://irc.mozilla.org/#services-dev

[...]

PR #352, opened 2016-03-10
Misc small changes.

prioux:
* Dearchiving a task updates updated_at timestamp
* Added DB consistency rules to SyncStatus model
* Removed 'Mozilla Persona' in login page HTML

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1205,16 +1205,11 @@ class ClusterTask < CbrainTask
       File.unlink(tar_file) rescue true
     end
 
-    # Keep updated_at value in order to reset it at the end of method.
-    updated_at_value = self.updated_at
-
-    # Mark task not archived
+    # Mark task not archived; this also sets the updated_at timestamp
+    # so that a task recently unarchived is considered 'updated'
     self.workdir_archived = false
     self.save!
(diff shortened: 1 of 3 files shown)


glatard:
Just curious: why is Mozilla Persona removed from login page?

driusan:
Mozilla killed Persona. The service doesn't exist anymore.

prioux:
@glatard @driusan Yes, Persona is dying.

But I removed Persona log in for another reason too. I've had several instances of people trying to connect to CBRAIN using persona, thinking they could use the service even though they didn't have an account. They thought Persona was a way to not only log in, but also to sign in for the service.

For the moment I only removed the Persona arrow in the login page and I've left the Persona code in the session controller, but I'll clean this up shortly too.

glatard:
Persona or not persona, maybe it would be useful to have a link to sign in to CBRAIN (i.e. create an account) instead of keeping it private. Removing a login option is not going to make account creation clearer...

glatard:
And yes, Persona will be decommissioned, in November 2016 (see #292). That should have left us some time to talk about possible replacements.

prioux:
Yes, we have an issue (was it in Redmine? must be on GitHub now) about having a configuration option in CBRAIN so that the admin can provide a link to a signup page. Then the link would appear within the login page. Or we could build a signup service within CBRAIN too.

PR #355, opened 2016-04-01
New subtasking mechanism, improved.

prioux:
The subtasking mechanism has been improved to make it a little bit more abstract. This is for the moment mainly used in conjunction with fsl_sub and fsl_randomize in the cbrain-plugins-neuro package.

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1767,7 +1767,7 @@ chmod 755 ./.dockerjob.sh\n
 
     # Sets tool config among tool configs accessible by user of current task
     new_tool                = new_task.tool
-    accessible_tool_configs = ToolConfig.find_all_accessible_by_user(self.user).where(:tool_id => new_tool.id)
+    accessible_tool_configs = ToolConfig.find_all_accessible_by_user(self.user).where(:tool_id => new_tool.id, :bourreau_id => self.bourreau_id)
     new_tcid                = accessible_tool_configs.limit(1).raw_first_column(:id)[0] if new_tcid.blank?
     if new_tcid.blank? || ! accessible_tool_configs.where(:id => new_tcid).exists?
       raise "Cannot find an acceptable tool config ID for this user, tool, and bourreau.\n"
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1720,8 +1720,6 @@ chmod 755 ./.dockerjob.sh\n
       end # each new json descriptor
       self.status_transition("Subtasking", current_status)
     end
-  rescue  # TODO log error?
-    nil
   end
 
   protected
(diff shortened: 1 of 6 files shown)


BrainPortal/app/models/cluster_task.rb
```diff
@@ -1687,6 +1689,151 @@ def docker_commands
...
+    # to a CBRAIN installation and cannot be easily obtained by an
+    # external agent. In case no tool config id is provided, the first
+    # accessible tool config id will be selected.
+    schema = {
+      "type"       => "object",
+      "required"   => [ "tool-class", "parameters" ],
+      "properties" => {
+        "tool-class"     => { "type" => "string" },                                         # Class of the new task
+        "tool-config-id" => { "type" => "number" },                                         # Tool config id of the new task
+        "share_wd_tid"   => { "type" => "string" },                                         # ID of other task to share workdir
```
glatard: dashes could be used instead of underscores (-> share-wd-tid)

prioux: OK. NaN added that. I didn't know they were interchangeable, or if not then it means the JSON creation code also need to be adjusted.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1687,6 +1689,151 @@ def docker_commands
...
+        ensure
+          File.unlink(taskfile) rescue nil # to make sure we won't retry forever
+          #For debug
+          #File.rename(taskfile,"#{taskfile}.processed")
+        end
+      end # each new json descriptor
+      self.status_transition("Subtasking", current_status)
+    end
+  rescue  # TODO log error?
+    nil
```
glatard: May not be a very good idea to mask an exception...

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1687,6 +1689,151 @@ def docker_commands
...
+
+    new_task.batch_id     = self.id
+    new_task.launch_time  = Time.now
+    self.level            = 0 if self.level.blank?
+    new_task.level        = self.level + 1 # New task will be one level up its "parent" task in the task table.
+    new_task.share_workdir_with(new_share_wd_tid, "Queued") if new_share_wd_tid # can be nil
+
+    # Sets tool config among tool configs accessible by user of current task
+    new_tool                = new_task.tool
+    accessible_tool_configs = ToolConfig.find_all_accessible_by_user(self.user).where(:tool_id => new_tool.id, :bourreau_id => self.bourreau_id)
```
glatard: In the future we may want to allow sub-tasks to be run in a different Bourreau. "use-same-bourreau" could be a parameter of the sub-task submission.

prioux: Okay, when we need to submit to other bourreaux we'll easily find the code and modify it accordingly. For the moment the default provides a bit of security.

PR #356, opened 2016-04-05
Esthetics1

prioux:
Some tiny esthetic changes.

- Support link now called "Help Site" instead of "user manual"
- No longer notifies users when background task actions have been sent
  because there was already a flash message about that.
- Cleanup of CSS


BrainPortal/public/stylesheets/cbrain.css
```diff
@@ -2152,7 +2147,7 @@ img {
 
 .button.grayed-out {
   opacity: 0.6;
-  filter: alpha(opacity=60);
+  filter: alpha(opacity=60);  /* darn IE */
```
remibernard: :+1:

PR #358, opened 2016-04-13
Form for "New User" is now in a separate page.

prioux:
When creating a new user, instead of having an annoying overlay, the layout is in a proper separate page, as the Rails Gods have always intended.

This is in preparation to have new fields and options when creating a new user (named templates for sets of groups).

Also, I improved the shared partial for groups checkpboxes, and fixed a bug where all the checkboxes had the same HTML ID (!)


prioux:
Andrew, we can't merge this pull request because it would undo some changes that were already commited to the `dev` branch; your created your own branch in your local repo based on an obsolete version of `dev` it seems, and it's missing some pull requests (e.g. #393).

In your lcoal repo you can try to:

* checkout dev
* pull from aces/cbrain's dev to make sure you're up to date
* either re-do all the work done here (including my comments) or try to REBASE your branch to the head of `dev` (that might work).

Also, some more general comments about this PR: some _form_fields.html.erb were left over and also the following models coudl also benefit from this work:

* feedbacks
* groups
* messages

PR #362, opened 2016-04-21
Add desc in qc panel.

natacha-beck:


BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -791,19 +791,20 @@ def quality_control_panel #:nodoc:
...
+        when :pass
+          tag_ids |= [pass_tag.id.to_s]
+        when :fail
+          tag_ids |= [fail_tag.id.to_s]
+        when :unknown
+          tag_ids |= [unknown_tag.id.to_s]
       end
-      @current_userfile.set_tags_for_user(current_user, tag_ids)
+      @current_userfile.set_tags_for_user(current_user, tag_ids) if tag_ids
+      @current_userfile.update_attributes_with_logging(params[:userfile], current_user) if params[:userfile][:description]
```
prioux: The "if" will always be true, no? How can there be a nil in the params[:userfile][:description] ?

prioux: Doesn't that allow a user to inject and modify any other attribtues in the userfile at the same time? If I edit the HTML and add a tag such as:

```html
   <input type="hidden" name="userfiles[name]" value="abcd">
```

wouldn't that change the name? Woudlnt' the same trick allow someone to chnage critical security-related attributes, like group_id, read_only, etc etc ?

BrainPortal/app/views/userfiles/_quality_control_panel.html.erb
```diff
@@ -54,14 +54,21 @@
...
-      @userfile.get_tags_for_user(current_user).raw_first_column(:id) %>
-</select>
-
-<br>
-
-<%= submit_tag "Update Tags" %>
+<div class="display_table">
+  <div class="display_cell">
+  Tags:<br>
+    <select name="tag_ids[]" multiple="multiple" style="width:150px" >
```
prioux: Don't use sizes in pixels...

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -791,19 +791,21 @@ def quality_control_panel #:nodoc:
...
-      when :unknown
-        tag_ids |= [unknown_tag.id.to_s]
+        when :pass
+          tag_ids |= [pass_tag.id.to_s]
+        when :fail
+          tag_ids |= [fail_tag.id.to_s]
+        when :unknown
+          tag_ids |= [unknown_tag.id.to_s]
       end
       @current_userfile.set_tags_for_user(current_user, tag_ids)
```
prioux: why not logging of tag changes?

PR #363, opened 2016-04-29
Do a 'docker pull' before each 'docker run' to make sure that the image is up-to-date.

glatard:
This is required to avoid inconsistencies among cluster nodes when an image, say foo:latest, is updated and the cluster nodes use it at different times.

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1677,14 +1677,14 @@ class ClusterTask < CbrainTask
     cache_dir=RemoteResource.current_resource.dp_cache_dir;
     task_dir=self.bourreau.cms_shared_dir;
     docker_commands = "cat << \"DOCKERJOB\" > .dockerjob.sh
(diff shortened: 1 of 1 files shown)


BrainPortal/app/models/cluster_task.rb
```diff
@@ -1681,6 +1681,9 @@ def docker_commands
 #{commands_joined}\n
 DOCKERJOB\n
 chmod 755 ./.dockerjob.sh\n
+# Pull the Docker image to avoid inconsistencies coming from different image versions on worker nodes
+#{docker_executable_name} pull #{self.tool_config.docker_image}\n
```
prioux: missing .bash_escape()

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1681,6 +1681,9 @@ def docker_commands
 #{commands_joined}\n
```
prioux: Why are there explicit \n added to each line of the HERE document?

PR #364, opened 2016-04-29
An interface for tasks to create a progress bar showing their status.

glatard:


BrainPortal/app/views/tasks/show.html.erb
```diff
@@ -24,6 +24,29 @@
 
 <% title(@task.name + ' Task Information') %>
 
+<!-- PROGRESS BAR -->
+<% if @task.respond_to?("progress_info") %>
```
prioux: I think the doc should by in portal_tasks.rb, with maybe a default behavior.
The return value would be better represented as a hash:

``ruby
 {
   :percent => 73,
   :message => "Oh no your data is corrupted",
  :color => "#ff0000",   # or red, or anything that is CSS compatible
}
```

Percentage is not shown if it's `nil`.

prioux: Also, respond_to?() is usually invoked with a `:symbol`

glatard:
Thanks for the comments. See update. show_percentage has to remain a boolean because the percentage value is still needed to display the progress bar when the percentage must not be shown.

PR #375, opened 2016-05-13
Fixed #374

### LINKED ISSUE #374: Handling of lists of strings in Boutiques
Lists of input parameters of type "String" are not properly handled in Boutiques. Only the first element of the list is put on the command line while all of them should be there. For instance, launching 
[this tool](https://github.com/aces/cbrain/files/263570/testList.txt) with three elements "one", "two" and "three" in input "input_list" will create the following command line: `echo one > output.txt` while it should be `echo one two three > output.txt`.

glatard:
This fixes issue #374. I tested it on a few examples to make sure it doesn't break other Boutiques tool, but I'd be interested to have @remibernard's opinion in case he has a few minutes to look at this.

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
@@ -108,7 +108,7 @@
   <ul class="tsk-prm-list">
     <li>
       <%%
-        input.(id, name,
+        input.(id, name+"[]",
           value:       first,
           optional:    optional,
           placeholder: placeholder
@@ -454,7 +454,7 @@ $(function () {
   parameters.delegate('.tsk-prm-add', 'click', function () {
     var name = $(this)
       .siblings('.tsk-prm-in')
-      .data('name');
+      .attr('name');
 
     $(this)
       .closest('ul')


BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -108,7 +108,7 @@
   <ul class="tsk-prm-list">
     <li>
       <%%
-        input.(id, name,
+        input.(id, name+"[]",
```
prioux: Be careful. In rails parameter names determine how their VALUES will be pre-processed...

```html
<input name="abc" value="val">
```

gives

```ruby
params =  { :abc => "val" }
```

while

```html
<input name="abc[]" value="val">
```

gives

```ruby
params = { :abc => [ "val" ] }
```

glatard:
Yes I know @prioux, this is precisely where the bug was coming from (arrays have to have a name ending with []). Do you actually have a concern about the code or is it just a general note?

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -454,7 +454,7 @@ $(function () {
   parameters.delegate('.tsk-prm-add', 'click', function () {
     var name = $(this)
       .siblings('.tsk-prm-in')
-      .data('name');
+      .attr('name');
```
remibernard: Be careful here. If the parameter is optional it may not have a `name` attribute. See [task_params.html.erb.erb:385](https://github.com/glatard/cbrain/blob/702a295bf4a4c0b248223797ab5fe6537da4b8ea/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb#L385). Might be easier to do `name += '[]'` in the `<% %>` block where `values` is processed and drop this hunk.

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
glatard: Yes I know @prioux, this is precisely where the bug was coming from (arrays have to have a name ending with []). Do you actually have a concern about the code or is it just a general note?

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
glatard: The problem is that list input elements don't seem to have data associated with them, so .data('name') is unde
screenshot from 2016-05-16 09-43-40
fined. See attached screenshot. Also, this code is run only when the user clicks on the "+" button to add a list element, which is only possible when the parameter is enabled when the list is optional. Did I miss anything?

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
prioux: Just a general note. I wasn't sure if the input element was to do double duty for single values and for array values; if it's supposed to always be array values, then all's fine.

Issue #373, opened 2016-05-13
Handling of flags in Boutiques.


glatard:
Boutiques implementation in CBRAIN has the following bug with the handling of flags: for it will substitute "-a false" on the command line when flag "-a" is not active.

This bug fix is required for all the Boutiques tool that have inputs of type Flag:
* https://github.com/glatard/cbrain-plugins-psom/issues/5
* MINCBet

prioux:
Just to help, I think the bug is in

    BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb

in method **apply_template()**

glatard:
btw, @ttaa9, you can write in files the Ruby classes generated by Boutiques using the following command in the Rails console:
`(SchemaTaskGenerator.generate "./lib/cbrain_task_generators/schemas/boutiques.schema.json","cbrain_plugins/.../<tool>.json").to_directory("<some_path>")`

remibernard:
[bourreau.rb.erb:438](https://github.com/aces/cbrain/blame/master/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb#L438)
```ruby
next template.gsub(key, flag) if flag && value == true
```

An explicit is-bool check instead of `value == true` and a ternary for the substitution value should do it.

As for being opaque, not sure how I could've made it clearer without making it a lot longer. Each block should be pretty straightforward to understand...

glatard:
The code is not opaque, it's clear and well documented. Thanks for the hint @remibernard!

ttaa9:
I've done as @remibernard recommended and it seems to have worked. Thanks!

PR #376, opened 2016-05-16
Fixing flag handling template generation bug (Issue #373)

ttaa9:


BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -435,7 +435,10 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
   def apply_template(template, keys, flags: {}, strip: [])
     keys.inject(template) do |template, (key, value)|
       flag = flags[key].presence
-      next template.gsub(key, flag) if flag && value == true
+  
+      # Flag type
+      isBool = (value.is_a?(TrueClass) || value.is_a?(FalseClass))
+      next template.gsub(key, (flag && value==true)  ? flag : '') if flag && isBool
```
prioux: I still don't like that we're doing
a) a conditional test
b) a substitution as an argument to `next` and
c) the loop control statement `next` for the inject method

all on a single line... anyway. Just IMHO.

PR #379, opened 2016-05-18
Added Boutiques capability for non-space flag separators.

ttaa9:


BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -215,8 +216,25 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
+%   end
 
+%   unless seps.empty?
+    # Input command-line separators (with flags) used with keys in command-line
+    # substitution.
+    seps = {
+%     key_width = max_width.(seps, 'command-line-key')  + "''".length
+%     seps.each do |sep|
+      <%=
+        "%-#{key_width}s => %s" % [
```
prioux: Can any of the keys or values include a single quote character? Because if so this will break.

ttaa9: Theoretically they could, I suppose. I think the same thing could happen with the code just above it as well (since that is what I based that part on). In some sense I guess it requires that the keys/values used in the Boutiques json schemas for these fields do not have dangerous characters like that.

prioux: I'm talking about argument that are user-supplied  using the form... e.g. a text input box. What is someone enters, as a value for one of these options, `ahaha"; rm -rf /`   (with a single double quote in the middle to trigger the injection)

remibernard: `apply_template` invokes `bash_escape` on values. As @ttaa9 noted, there's no safety if the descriptor contains quotes though.

prioux:
@remibernard Tristan and I ahve discovered that Ruby's String#gsub() function has an unfortunate side effect: of the thing you are substituting TO contains the two character sequences `[backslash][backtick]`, `[backslash][ampersand]` or `[backslash][singlequote]` 

> ( \`, \&, \' )  

then the substitution engine replace them (respectively) by `what's before the matched string`, `the matched string`, `what's after the matched string`.

```
Bianca :086 > puts "beforeXXafter".gsub("XX", "=\\`=")
before=before=after
 => nil 
Bianca :087 > puts "beforeXXafter".gsub("XX", "=\\&=")
before=XX=after
 => nil 
Bianca :088 > puts "beforeXXafter".gsub("XX", "=\\'=")
before=after=after
```

Since we do have strings containing those characters sequence, we have to re-code the gsub() part.

prioux:
Ah! I had overlooked a simple solution, @remibernard and @tta9 : 

```
Bianca :088 > puts "beforeXXafter".gsub("XX", "=\\'=")
before=after=after
Bianca :090 > puts "beforeXXafter".gsub("XX") { "=\\'=" }
before=\'=after
```

prioux:
There's even a version of gsub that takes a hash as a second argument:

```
Bianca :093 > puts "beforeXXafter".gsub("XX", "XX" => "=\\'=" )
before=\'=after
```

ttaa9:
The block argument one seems like the best solutionn

PR #386, opened 2016-06-01
Access profiles

prioux:
This is a new model that is only useful for the administrator.
Each AccessProfile is basically

* a list of groups (a.k.a projects)
* a list of users

The purpose is to simplify user management. The admin
creates a AccessProfile, assigns it a bunch of groups,
and then when creating users or modifying them, they
can be assigned to the profile. When the profile's
list of group is changed, all users associated with the
profile will have their list of groups changed too.

Users can be assigned multiple AccessProfiles, which means
they get the UNION set of all groups from those profiles.

AccessProfiles have pretty labels the show up as colorful
boxes for the admin.

### DIFF
--- a/BrainPortal/app/controllers/access_profiles_controller.rb
+++ b/BrainPortal/app/controllers/access_profiles_controller.rb
@@ -75,7 +75,7 @@ class AccessProfilesController < ApplicationController
     old_user_ids  = @access_profile.user_ids.sort
 
     # The success variable will be false for errors on ordinary attributes;
-    # for the group_ids and user_ids list, these are always updated... :-(
+    # for the group_ids and user_ids list, these are always updated with no errors or logging... :-(
     success = @access_profile.update_attributes_with_logging(params[:access_profile], current_user)
 
     # Adjust groups and users, and log differences
@@ -89,13 +89,8 @@ class AccessProfilesController < ApplicationController
       ap_removed_gids = old_group_ids - new_group_ids # what disappeared in the current AP
       affected_users = User.where(:id => params[:affected_user_ids])
       affected_users.each do |user|
-        # Find the union of all group_ids for all the APs the user is still associated with
-        add_gids    = user.access_profiles.inject([]) { |tot,ap| tot += ap.group_ids; tot }
-        remove_gids = ap_removed_gids - add_gids # must not remove gids that are also present in other APs
-
-        # Update the list of group for the user
         orig_user_gids = user.group_ids
-        user.group_ids = user.group_ids - remove_gids + add_gids
+        user.apply_access_profiles(remove_group_ids: ap_removed_gids) # (re)add all groups from all APs of the user, minus lost groups
         user.addlog_object_list_updated("Updated Access Profile '#{@access_profile.name}', Projects",
                                         Group, orig_user_gids, user.group_ids, current_user)
       end
@@ -105,7 +100,7 @@ class AccessProfilesController < ApplicationController
     added_uids   = new_user_ids - old_user_ids
     User.find(added_uids).each do |user|
       orig_user_gids = user.group_ids
-      user.group_ids = orig_user_gids + @access_profile.group_ids
+      user.apply_access_profiles() # add all groups in all APs of the user, including the current one
       user.addlog_object_list_updated("Added Access Profile '#{@access_profile.name}', Projects",
                                       Group, orig_user_gids, user.group_ids, current_user)
     end
@@ -113,12 +108,8 @@ class AccessProfilesController < ApplicationController
     # Some users lost access to the the AP? Adjust their groups
     removed_uids = old_user_ids - new_user_ids
     User.find(removed_uids).each do |user|
-      # Find the union of all group_ids for all the APs the user is still associated with
-      ap_group_ids = user.access_profiles.inject([]) { |tot,ap| tot += ap.group_ids; tot }
-      # Find what group_ids are only in the current AP, and remove them
-      remove_gids = @access_profile.group_ids - ap_group_ids
       orig_user_gids = user.group_ids
-      user.group_ids = orig_user_gids - remove_gids
+      user.apply_access_profiles(remove_group_ids: new_group_ids) # try removing all groups in current AP
       user.addlog_object_list_updated("Removed Access Profile '#{@access_profile.name}', Projects",
                                       Group, orig_user_gids, user.group_ids, current_user)
     end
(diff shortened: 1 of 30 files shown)


BrainPortal/app/views/access_profiles/new.html.erb
```diff
@@ -0,0 +1,66 @@
...
+
+    <span title="Brief description of the profile.">
+      <%= f.label :description, "Description" %><br/>
+      <%= f.text_area :description, :rows => 6, :cols => 60 %><br/>
+      <span class="field_explanation">These are your private notes about this profile.</span>
+    </span>
+
+  </div>
+
+  <div class="generalbox">
```
natacha-beck: Could be better with some space between the 2 boxes.

BrainPortal/public/doc/access_profiles/access_profiles.html
```diff
@@ -0,0 +1,58 @@
...
+<p>
+Access Profiles and all their properties are never visible to normal users.
+
+<h2>Creating an Access Profile</h2>
+
+When creating a new Access Profile, the admin can set a color to help<br>
+differentiate it from other Access Profiles. Projects can also immediately<br>
+be assigned to it.
+
+<h2>Deleting an Access Profile</h2>
```
natacha-beck: Not clear, can you re-write the description.

natacha-beck:
In Group show page, it will be useful to have the information about "Access Report".

BrainPortal/app/controllers/access_profiles_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+    old_group_ids = @access_profile.group_ids.sort
+    old_user_ids  = @access_profile.user_ids.sort
+
+    # The success variable will be false for errors on ordinary attributes;
+    # for the group_ids and user_ids list, these are always updated with no errors or logging... :-(
+    success = @access_profile.update_attributes_with_logging(params[:access_profile], current_user)
+
+    # Adjust groups and users, and log differences
+    new_group_ids = @access_profile.group_ids.sort
+    new_user_ids  = @access_profile.user_ids.sort
```
glatard: Do you really need to sort these arrays?

BrainPortal/app/controllers/access_profiles_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+    if old_group_ids != new_group_ids
+      ap_removed_gids = old_group_ids - new_group_ids # what disappeared in the current AP
+      affected_users = User.where(:id => params[:affected_user_ids]).all
+      affected_users.each do |user|
+        orig_user_gids = user.group_ids
+        user.apply_access_profiles(remove_group_ids: ap_removed_gids) # (re)add all groups from all APs of the user, minus lost groups
+        user.addlog_object_list_updated("Updated Access Profile '#{@access_profile.name}', Projects",
+                                        Group, orig_user_gids, user.group_ids, current_user)
+      end
+    end
```
glatard: So if:
* user A belongs to group X
* user A is added to AccessProfile P that contains groups X and Y
* X is removed from P
Then user A is removed from group X. I can see issues coming with that but I don't really have a solution :)

BrainPortal/app/controllers/access_profiles_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+                                      Group, orig_user_gids, user.group_ids, current_user)
+    end
+
+    # Some users lost access to the the AP? Adjust their groups
+    removed_uids = old_user_ids - new_user_ids
+    User.find(removed_uids).each do |user|
+      orig_user_gids = user.group_ids
+      user.apply_access_profiles(remove_group_ids: new_group_ids) # try removing all groups in current AP
+      user.addlog_object_list_updated("Removed Access Profile '#{@access_profile.name}', Projects",
+                                      Group, orig_user_gids, user.group_ids, current_user)
```
glatard: A similar issue exist than in the previous note.

BrainPortal/app/controllers/access_profiles_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+  def destroy #:nodoc:
+    @access_profile = AccessProfile.find(params[:id])
+    orig_user_ids   = @access_profile.user_ids
+    orig_group_ids =  @access_profile.group_ids
+    @access_profile.destroy
+    User.find(orig_user_ids).each do |user|
+      orig_user_gids = user.group_ids
+      user.apply_access_profiles(remove_group_ids: orig_group_ids) # try removing all groups in current AP
```
glatard: same comment as before.

BrainPortal/app/models/user.rb
```diff
@@ -373,6 +374,61 @@ def destroy_user_sessions
...
+  # we can supply a list of removed group_ids in +remove_group_ids+ :
+  #
+  #   ap1.group_ids = [ 12, 13 ]  # we removed 11 and 99
+  #   user.apply_access_profiles( [ 11, 99 ] )
+  #
+  # will result in the user's list of groups to lose group 11 but NOT
+  # group 99, because it's present in ap2.
+  def apply_access_profiles(remove_group_ids: [])
+    gids = union_group_ids_from_access_profiles
+    self.group_ids = self.group_ids - remove_group_ids + gids # - and + are NOT COMMUTATIVE!
```
glatard: Well + is commutative (you could have written ```gids + self.group_ids - remove_group_ids```).

driusan: Assuming those are set operations and the comment you're replying to isn't a lie, that would be a different result if anything is in both gids and remove_group_ids.

gids + self.group_ids - remove_group_ids = the things in common are removed from the set
group_ids - remove_group_ids + gids = the things in common are included in the set

prioux: Yes, that is exactly what I want:

The list "remove_group_ids" contains a list of group_ids that were removed in editing the current AP; we want to remove them form the user's own list but only if they are not already part of the list of other APs (in gids). So adding gids at the end just ensures that.

prioux: I see Dave was commenting on Tristan's comment (I had not read it). Anyway, yes, the code as shown is what I want, changing the order of the lists + and - would have resulted in a wrong result.

BrainPortal/app/controllers/access_profiles_controller.rb
prioux: No. But I like my code to be deterministic. Helps when debugging. The penalty is just 0.000000001 millisecond.

BrainPortal/app/controllers/access_profiles_controller.rb
prioux: Yes. Unfortunately the AccessProfile system doesn't interact well with existing group assignments unless you're careful. It's a known limitation; I was expecting admins to use AccessProfiles only with brand new groups expressly created for AccessProfiles, or at least if existing groups are used in access profiles, they would be used only to be added to such profiles and rarely removed.

BrainPortal/app/controllers/access_profiles_controller.rb
prioux: Yes, and same answer.

PR #387, opened 2016-06-01
MySQL performance improvements.

prioux:
Changed a little bit how some queries are performed to build the
filter lists in the index table headers. Performance improved.

### DIFF
--- a/BrainPortal/app/helpers/scope_helper.rb
+++ b/BrainPortal/app/helpers/scope_helper.rb
@@ -480,7 +480,7 @@ module ScopeHelper
     # hash (<value> => [<filter>]) and append the generated counts to each
     # value in the base filter set
     view_filters = filter_values_for(
-      scopes.inject(view) { |v, s| s.apply(v) },
+      scopes.inject(view) { |v, s| s.apply(v, sorting: false) },
       attribute,
       association: association,
       format: lambda { |x| x }
@@ -681,13 +681,10 @@ module ScopeHelper
     # Fetch the main filter values as an array of arrays:
     # [[value, label, count], [...]]
     model
-      .where("#{attribute} IS NOT NULL")
       .order(label, attribute)
       .group(attribute, label)
       .raw_rows(attribute, "#{label} AS #{label_alias}", "COUNT(#{attribute})")
       .reject { |r| r.first.blank? }
-      .map(&:to_a)
-      .to_a
   end
 
   # Fetch the possible values (and their count) for +attribute+ within
--- a/BrainPortal/lib/view_scopes.rb
+++ b/BrainPortal/lib/view_scopes.rb
@@ -189,6 +189,10 @@ module ViewScopes
     #
     # Note that pagination is not applied by default; specify +paginate+ to
     # paginate +collection+.
+    #
+    # If +sorting+ is set to true (the default), both filtering rules
+    # and sorting will be applied; when this option is set to false
+    # only the filtering rules will be applied.
     def apply(collection, on_failure: :ignore, paginate: false, sorting: true)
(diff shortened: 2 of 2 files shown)


BrainPortal/app/helpers/scope_helper.rb
```diff
@@ -681,13 +681,10 @@ def self.model_filter_values(model, attribute, label, association: nil)
     # Fetch the main filter values as an array of arrays:
     # [[value, label, count], [...]]
     model
-      .where("#{attribute} IS NOT NULL")
```
prioux: Surprisingly, this makes a huge difference!

BrainPortal/lib/view_scopes.rb
```diff
@@ -189,7 +189,7 @@ def initialize_copy(other)
     #
     # Note that pagination is not applied by default; specify +paginate+ to
     # paginate +collection+.
-    def apply(collection, on_failure: :ignore, paginate: false)
+    def apply(collection, on_failure: :ignore, paginate: false, sorting: true)
```
remibernard: Please do add some documentation for this new parameter.

BrainPortal/app/helpers/scope_helper.rb
remibernard: It does leave the possibility of `NULL`s messing up the counts, however.
It was included in the original filtering system, and included here as a safety measure.

I would guess the production environment doesn't have `NULL`s in filterable columns anyway.

BrainPortal/app/helpers/scope_helper.rb
```diff
@@ -681,13 +681,10 @@ def self.model_filter_values(model, attribute, label, association: nil)
...
       .order(label, attribute)
       .group(attribute, label)
       .raw_rows(attribute, "#{label} AS #{label_alias}", "COUNT(#{attribute})")
       .reject { |r| r.first.blank? }
-      .map(&:to_a)
-      .to_a
```
remibernard: The interface docs for this function clearly indicates that it returns arrays of arrays. These two calls are here to ensure that.
Please do update the docs accordingly.

prioux: @remibernard I'm not sure why these lines were there... raw_rows() already returns an array of array, the reject() doesn't change that, so the last two lines don't do anything. The method still always return the same type as before, an array of arrays...

BrainPortal/app/helpers/scope_helper.rb
prioux: And if there were any null, they'd be rejected() a few lines later anyway.

BrainPortal/lib/view_scopes.rb
prioux: Done; I will wait until we've finished the other discussion about the to_a() in the other file before I commit.

BrainPortal/app/helpers/scope_helper.rb
remibernard: I'm pretty sure I didn't look up `raw_rows`' docs and just assumed Enumerable when I wrote this part. But if `raw_rows` does it, you're right that these two lines serve no purpose.

BrainPortal/app/helpers/scope_helper.rb
remibernard: They will still be included in the `COUNT`, however.

prioux: Good point but I'll toleratethat if it happens. :-)

BrainPortal/app/helpers/scope_helper.rb
prioux: Well, I'm the one who wrote raw_rows() so it yes I can confirm that's what it does. :-)

PR #393, opened 2016-06-03
Cleaned up and restructured the DP class tree.

prioux:
I've introduced three new DP classes:

* FlatDirLocalDataProvider
* FlatDirSshDataProvider
* FlatDirSmartDataProvider

They are browsable with the files all in
one level of a directory; this is similar to
the SshDataProvider but with two different
transport mechanism (local and ssh). In fact,
the FlatDirSshDataProvider is just a subclass
of SshDataProvider with nothing new in it, since
all the function was already in there.

I've improved the content of LocalDataProvider
to provide a proper base set of methods for building
local DPs. It almost (but not exactly) implements
the full set of methods for FlatDirLocalDataProvider.


BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -258,9 +258,9 @@ def display
     begin
       if @viewer
         if params[:apply_div] == "false"
-          render :file   => @viewer.partial_path.to_s, :layout => false
+          render :file   => @viewer.partial_path.to_s, :layout => params[:apply_layout].present?
```
prioux: This particular change is not related to Data Providers, but it can make the CivetQC viewers must nicer in cbrain-plugins-neuro (pull request pending).

glatard:
Could you add or reference your DP diagram somewhere? It would make the relation between DP classes easier to grasp.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -814,10 +814,13 @@ def repair
   private
 
   def get_type_list #:nodoc:
-    data_provider_list = (check_role(:site_manager) || check_role(:admin_user)) ? DataProvider.descendants.map(&:name).sort : %w{ SshDataProvider }     
+    data_provider_list = [ "FlatDirSshDataProvider" ]
+    if check_role(:site_manager) || check_role(:admin_user)
+      data_provider_list = DataProvider.descendants.map(&:name)
+    end
```
glatard: Shouldn't local data providers be accessible to regular users?

natacha-beck:
> Could you add or reference your DP diagram somewhere? It would make the relation between DP classes easier to grasp.

The wiki seems to be a good place for that.

glatard:
The code looks good to me but I'm a bit worried by the complexity of the DP types (but it was the case already before this PR). For instance, what's the point of the Local and Ssh variations in addition to Smart. Couldn't everything be Smart? Also, if IncomingVault and Vault really differ only by the fact that IncomingVault is browsable (as suggested by the documentation), then couldn't IncomingVault be replaced by a boolean attribute in Vault? Similarly, couldn't the distinction between FlatDir, En, Vault and IncomingVault be just a "File organization model" property? Maybe this could be simplified at least in the GUI, to make it easier for admins to create the right DP.

PR #394, opened 2016-06-06
Various improvements in Boutiques

ttaa9:


BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
```diff
@@ -261,7 +266,55 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
+%# Check for violations pertaining to parameter groups
+%# In particular, check that mutual exclusivity and one-is-required are satisfied 
+%#
+% unless noGroups
+%# Check if any groups require checking
+% getGroupsWith = lambda { |prop| groups.select{ |g| g[prop] }.map{ |g| ":'"+g["id"]+"'" }.join(", ") }
+% mutexGrps = getGroupsWith.("mutually-exclusive") 
+% oneReqGrps = getGroupsWith.("one-is-required")
+% hasNoMutex = (mutexGrps == "")
+% hasNoOneReq = (oneReqGrps == "")
```
glatard: style: '=' signs could be aligned in the previous 5 lines.

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
```diff
@@ -261,7 +266,55 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
+% unless (hasNoMutex && hasNoOneReq)
+   
+    # Groups with the mutually exclusive or one-is-required
+    <%= 'mutexGroups = [%s]'  % mutexGrps  unless hasNoMutex %>
+    <%= 'oneReqGroups = [%s]' % oneReqGrps unless hasNoOneReq %>
+    # Mapping from groupId to members list
+%   writeLiteralMap.("gidToMbrs", gIdToMbrs)
+    # Mapping from groupId to group name
+    grpName = {<%= groups.map{ |g| ":'" + g["id"] + "\' => \'" + g["name"] + "'" }.join(", ") %>}
+%  unless hasNoMutex
```
glatard: style: what about 'if hasMutex'? hasNoMutex and hasNoOneReq are used only with unless, double negations could be avoided.

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
```diff
@@ -261,7 +266,55 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
+    # Lambda for checking mutual exclusivity
+    isMutex = lambda { |gid| gidToMbrs[gid].select{ |m| ! isInactive.(m.to_sym) }.count <= 1 }
+%  end
+%  unless hasNoOneReq
+    # Lambda for checking one-is-required (at least one is active)
+    hasOneActive = lambda { |gid| gidToMbrs[gid].select{ |m| ! isInactive.(m.to_sym) }.count > 0 }
+%  end
+    # Check for violations of group properties
+%  unless hasNoMutex
```
glatard: group with previous 'unless hasNoMutex'?

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
```diff
@@ -261,7 +266,55 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
+    hasOneActive = lambda { |gid| gidToMbrs[gid].select{ |m| ! isInactive.(m.to_sym) }.count > 0 }
+%  end
+    # Check for violations of group properties
+%  unless hasNoMutex  
+    mutexGroups.each do |group|
+      errMsg = "violates group mutual exclusivity requirement"
+      params_errors.add(grpName[group],errMsg) unless isMutex.(group)
+    end
+%  end
+%  unless hasNoOneReq
```
glatard: group with previous 'unless hasNoOneReq'?

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -318,25 +340,88 @@
...
+    <div style="border-bottom: 2px solid #bbbbbb; padding-bottom: 2px;">
+      <h3 style="border-width: 0px; margin: 0px; padding: 1px; padding-left: 0px;">
+        <%= name %> <%% groupCheckbox.('<%= id %>',<%= defaultChecked %>) %>  
+      </h3>
+      <h5 style="font-size : 9.5pt; font-weight: normal; padding-top: 2px;" class="grp-desc"> 
+        <%= groupDescription %> 
+      </h5> 
+    </div><%%= "<br>".html_safe %>
+    <%# Write group parameters %>
+    <%- groupParams.each { |p| parameter.(p, true) } -%>
```
glatard: you could write the required parameters before the optional ones.

glatard:
The code looks good to me (up to the minor comments reported above), thanks for another nice feature! It will require quite some testing though.

prioux:
Question. For task_params.html.erb.erb, do we really want to embed so much javscript and CSS code, instead of putting this is separate files? A task has access to a private assets directory where they could be put.

BrainPortal/app/views/tools/_tools_table.html.erb
```diff
@@ -113,6 +113,9 @@
 
     t.column("Help", :help) do |tool|
       path = tool.cbrain_task_class.constantize.public_path("edit_params_help.html") rescue nil
+      if path.nil? # check if it is Boutiques tool
+        path = tool.cbrain_task_class.constantize.get_boutiques_help_filepath rescue nil
```
glatard: Would this break (method does not exist) if path is nil but the tool isn't a Boutiques tool?

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -165,6 +165,20 @@ def integrate(register: true, multi_version: false)
...
+      basePath = Rails.root.join('public/' + helpFileDir) 
+      FileUtils.mkdir_p basePath.to_s # creates directory if needed
+      helpfilePath = basePath.join(helpFileName).to_s
+      File.open( helpfilePath , "w" ){ |f|
+        f.write( generated.source[:edit_help] )
+      }
+      FileUtils.chmod 0775, helpfilePath
+
+      # Add a helper method for accessing the help file (for use on the tools page) 
+      task.define_singleton_method(:get_boutiques_help_filepath){ helpFileDir + helpFileName }
```
glatard: I would just call this method "help_filepath".

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -198,7 +212,7 @@ def register(task)
       version      = @descriptor['tool-version'] || '(unknown)'
       description  = @descriptor['description']  || ''
       docker_image = @descriptor['docker-image']
-      resource     = RemoteResource.current_resource
+      resource     = RemoteResource.current_resource
```
glatard: trailing space

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -350,16 +350,28 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
     # Save (and register) all generated files to the results data provider
     outputs.each do |param, paths|
       paths = [paths] unless paths.is_a?(Enumerable)
       paths.each do |path|
-        next unless path.present? && File.exists?(path)
+        
+        # Print a warning if the file is not present
+        # Note that, since the output may be optional, the process may still be successful
+        unless path.present? && File.exists?(path)
+          self.addlog("Unable to find optional output file: #{path}") if isOptional.(param)
```
glatard: could add the if clause to the unless one above.

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -232,6 +232,11 @@ def register(task,create_tool_config)
       # sense on the portal).
       return if Rails.root.to_s =~ /BrainPortal$/
 
+      # Create a ToolConfig iff 
+      #   (1) the Bourreau has a docker executable and 
+      #   (2) the descriptor specifies a docker image
+      return if docker_image.nil? || resource.docker_executable_name.nil?
```
glatard: Be careful here: some Bourreaux may have Docker installed but no Docker executable defined. The Docker executable defaults to "docker", it's only that some clusters have wrappers for it (e.g. 'udocker').

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1668,6 +1666,11 @@ def docker_executable_name
     return RemoteResource.current_resource.docker_executable_name.presence || "docker"
   end
 
+  # Return whether the Bourreau has docker present
+  def docker_exec_present?
```
glatard: Why not just "docker_present?"

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1677,164 +1680,18 @@ def docker_commands
...
-chmod 755 ./.dockerjob.sh
-# Pull the Docker image to avoid inconsistencies coming from different image versions on worker nodes
-#{docker_executable_name} pull #{self.tool_config.docker_image.bash_escape}
-# Run the task commands
-#{docker_executable_name} run --rm -v ${PWD}:${PWD} -v #{cache_dir}:#{cache_dir} -v #{task_dir}:#{task_dir} -w ${PWD} #{self.tool_config.docker_image.bash_escape} ${PWD}/.dockerjob.sh
+#!/bin/bash -l\n
+#{commands_joined}\n
+DOCKERJOB\n
+chmod 755 ./.dockerjob.sh\n
+#{docker_executable_name} run --rm -v ${PWD}:${PWD} -v #{cache_dir}:#{cache_dir} -v #{task_dir}:#{task_dir} -w ${PWD} #{self.tool_config.docker_image} ${PWD}/.dockerjob.sh \n
```
glatard: Why did you remove the Docker pull? This is actually useful and needed. See #363.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1677,164 +1680,18 @@ def docker_commands
...
-        "required-to-post-process" => { "type" => "boolean" }  # If true, the current task will not post-process before the new task is completed
-      }
-    }
-    JSON::Validator.validate!(schema,json_string) # raises an exception if json_string is not valid
-  end
-
-
-  ##################################################################
   # Lifecycle hooks
   ##################################################################
```
glatard: Not sure why you touched this method. I don't think you should have.

BrainPortal/app/views/bourreaux/show.html.erb
```diff
@@ -360,8 +359,9 @@
       <%= show_table(@bourreau, :header => "Docker Configuration", :edit_condition => @bourreau.has_owner_access?(current_user)) do |t| %>
         <% t.edit_cell :docker_executable_name, :header => "Docker executable" do %>
           <%= text_field_tag "bourreau[docker_executable_name]", @bourreau.docker_executable_name, :size => 60 %><br/>
-           <div class="field_explanation">Name of the Docker executable available on the machines where tasks will run. Defaults to "docker" if left empty.</div>
+           <div class="field_explanation">Name of the Docker executable available on the machines where tasks will run. Defaults to "docker" if left empty.</div>
```
glatard: Introduced trailing space.

BrainPortal/app/views/bourreaux/show.html.erb
```diff
@@ -360,8 +359,9 @@
       <%= show_table(@bourreau, :header => "Docker Configuration", :edit_condition => @bourreau.has_owner_access?(current_user)) do |t| %>
         <% t.edit_cell :docker_executable_name, :header => "Docker executable" do %>
           <%= text_field_tag "bourreau[docker_executable_name]", @bourreau.docker_executable_name, :size => 60 %><br/>
-           <div class="field_explanation">Name of the Docker executable available on the machines where tasks will run. Defaults to "docker" if left empty.</div>
+           <div class="field_explanation">Name of the Docker executable available on the machines where tasks will run. Defaults to "docker" if left empty.</div> 
         <% end %>
+        <% t.boolean_edit_cell( 'bourreau[docker_present]', !!@bourreau.docker_present, true, false, :header => "Docker Present?" ) %>
```
glatard: Why double "!"?

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -235,7 +235,7 @@ def register(task,create_tool_config)
       # Create a ToolConfig iff 
       #   (1) the Bourreau has a docker executable and 
       #   (2) the descriptor specifies a docker image
-      return if docker_image.nil? || resource.docker_executable_name.nil?
+      return if docker_image.nil? || !resource.docker_present
```
glatard: OK!

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -368,12 +368,25 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
         userfile_class = Userfile.suggested_file_type(name)
-        userfile_class ||= ( File.directory?(path) ? FileCollection : SingleFile )
+        userfile_class ||= ( File.directory?(path) ? FileCollection : SingleFile ) 
+       
+        # If the file already exists, add a tag (task id + ns time) to the name
+        if userfile_exists(userfile_class, :name => name)
+          tag = '-' + (self.id.to_s || 'unk') + '-' + Time.now.nsec.to_s
+          newName = name + tag if (userfile_class == FileCollection) && name.rindex(".").nil?
+          ext = File.extname( name ) 
+          newName ||= name.gsub( /#{ext}$/, "#{tag + ext}" )
```
glatard: What about files that are not FileCollection and don't have any extension in their name?

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -368,12 +368,25 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
+        # Get name and filetype
+        self.addlog("Attempting to save result file #{path}")
+        name = File.basename(path)         
         userfile_class = Userfile.suggested_file_type(name)
-        userfile_class ||= ( File.directory?(path) ? FileCollection : SingleFile )
+        userfile_class ||= ( File.directory?(path) ? FileCollection : SingleFile ) 
+       
+        # If the file already exists, add a tag (task id + ns time) to the name
+        if userfile_exists(userfile_class, :name => name)
+          tag = '-' + (self.id.to_s || 'unk') + '-' + Time.now.nsec.to_s
```
glatard: Tags are going to be ugly... Why not just increasing integers? i.e., in pseudo-code: "tag=1; while 'name-tag' exists, tag+=1, end".

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1668,6 +1688,11 @@ def docker_executable_name
     return RemoteResource.current_resource.docker_executable_name.presence || "docker"
   end
 
+  # Return whether the Bourreau has docker present or not
+  def docker_exec_present?
```
glatard: Why not just "docker_present?"

BrainPortal/app/views/bourreaux/show.html.erb
```diff
@@ -362,6 +362,7 @@
           <%= text_field_tag "bourreau[docker_executable_name]", @bourreau.docker_executable_name, :size => 60 %><br/>
            <div class="field_explanation">Name of the Docker executable available on the machines where tasks will run. Defaults to "docker" if left empty.</div>
         <% end %>
+        <% t.boolean_edit_cell( 'bourreau[docker_present]', !!@bourreau.docker_present, true, false, :header => "Docker Present?" ) %>
```
glatard: Why double "!"?

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -166,6 +166,20 @@ def integrate(register: true, create_tool_config: false, multi_version: false)
         })[partial]
       end
 
+      # Write out a help file for this Boutiques task
+      helpFileName = name + "_help.html" 
+      helpFileDir = "cbrain_plugins/cbrain_tasks/help_files/"
```
glatard: not portable (should use File.join).

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -166,6 +166,20 @@ def integrate(register: true, create_tool_config: false, multi_version: false)
...
+      basePath = Rails.root.join('public/' + helpFileDir)
```
glatard: Use File.join.

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -166,6 +166,20 @@ def integrate(register: true, create_tool_config: false, multi_version: false)
         })[partial]
       end
 
+      # Write out a help file for this Boutiques task
+      helpFileName = name + "_help.html" 
+      helpFileDir = "cbrain_plugins/cbrain_tasks/help_files/"
+      basePath = Rails.root.join('public/' + helpFileDir) 
+      FileUtils.mkdir_p basePath.to_s # creates directory if needed
```
glatard: There is a (tacit?) convention in CBRAIN that we use parentheses in method arguments.

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
glatard: There is a (tacit?) convention in CBRAIN that we try to align '=' signs as much as possible.

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -166,6 +166,20 @@ def integrate(register: true, create_tool_config: false, multi_version: false)
...
+      helpfilePath = basePath.join(helpFileName).to_s
+      File.open( helpfilePath , "w" ){ |f|
+        f.write( generated.source[:edit_help] )
+      }
+      FileUtils.chmod 0775, helpfilePath
```
glatard: Use parentheses.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -350,18 +350,43 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
-        name = File.basename(path)
         
-	userfile_class = Userfile.suggested_file_type(name)
-        userfile_class ||= ( File.directory?(path) ? FileCollection : SingleFile )
+        # Print a warning if the file is not present
+        # Note that, since the output may be optional, the process may still be successful
+        unless path.present? && File.exists?(path)
+          self.addlog("Unable to find optional output file: #{path}") if isOptional.(param)
+          next
+        end
```
glatard: could add the if clause to the unless one above (or better:  the other way around).

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -350,18 +350,43 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
-
-        self.addlog("Saving result file #{path}")
-        name = File.basename(path)
         
-	userfile_class = Userfile.suggested_file_type(name)
-        userfile_class ||= ( File.directory?(path) ? FileCollection : SingleFile )
+        # Print a warning if the file is not present
+        # Note that, since the output may be optional, the process may still be successful
+        unless path.present? && File.exists?(path)
+          self.addlog("Unable to find optional output file: #{path}") if isOptional.(param)
```
glatard: You could group the unless and if clauses.

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
```diff
@@ -261,7 +266,55 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
+      errMsg = "violates group mutual exclusivity requirement"
+      params_errors.add(grpName[group],errMsg) unless isMutex.(group)
+    end
+%  end
+%  if hasOneReq
+    oneReqGroups.each do |group|
+      errMsg = "violates group one-is-required specification"
+      params_errors.add(grpName[group],errMsg) unless hasOneActive.(group)
+    end
+%  end
```
glatard: I'm still unsure why the if statements can't be grouped in a single "if hasMutex" and a single "if hasOneReq".

BrainPortal/lib/view_scopes.rb
```diff
@@ -212,13 +216,15 @@ def apply(collection, on_failure: :ignore, paginate: false)
...
+          collection = @order.inject(collection) do |collection, order|
+            wrap.(collection) { order.apply(collection) }
+          end
+        else
+          collection = wrap.(collection) { @order.first.apply(collection) } unless
+            @order.empty?
         end
-      else
-        collection = wrap.(collection) { @order.first.apply(collection) } unless
-          @order.empty?
```
glatard: Not sure why this file was modified. Is it from a commit of yours?

glatard:
Good work but grouping commits from different features in the same PR makes it difficult to review. Pls split in different PRs in the future. See my comments above.

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
prioux: Often the CBRAIN codebase use

```ruby
   x =Pathname.new("/some/path")
   y = x + Pathname.new("other/path")
   z = x + "subdir"
```

BrainPortal/app/views/bourreaux/show.html.erb
ttaa9: If I recall correctly, it can be nil, but it needs to be false in that case (i.e. boolean type in this case) or it does not work (i.e. a `value==false` is done somewhere, which nil does not satisfy).

BrainPortal/app/models/cluster_task.rb
ttaa9: That method might be unnecesssary actually

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
ttaa9: I agree; that is what I originally did, but I thought it would be a bit safer against race conditions.

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
glatard: Then so be it, but let's avoid + on Strings.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
ttaa9: In that case I believe it does append it to the end, because newName will be nil so the line with gsub will run and append it to the end

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
glatard: yeah... ok :)

BrainPortal/app/views/tools/_tools_table.html.erb
ttaa9: No, because of the rescue nil, it will become nil just like it would have without my changes

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
glatard: Well, ext would be the empty string, right? 

This is what I'd think would happen:
```
2.0.0-p247 :006 > "foo".gsub(/""$/, "-a")
 => "foo" 
```

BrainPortal/app/views/bourreaux/show.html.erb
glatard: OK!

BrainPortal/app/models/cluster_task.rb
ttaa9: This seems to be before my merging (outdated diff).

BrainPortal/app/models/cluster_task.rb
ttaa9: Also not something that I actually changed (outdated diff)

BrainPortal/app/views/tools/_tools_table.html.erb
glatard: mmm, ok. ```if path.nil?``` could be replaced by ```path ||= ...```.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
ttaa9: I think it is supposed to print only if the file is optional (because it prints a warning about required files above), but  the "next" statement is supposed to run in either case

glatard: ok

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
ttaa9: You're right, checking the file type is not really necessary I think. Originally I was going to only append directly to directory objects... but then I realized many tools output directories with names like blah.anat.

Incidentally, something special in the string substitution  must be going on:

```
2.2.4 :031 > "foo".gsub(/''$/,"-a")
 => "foo" 
2.2.4 :032 > "foo".gsub(/#{ File.extname('foo') }$/,"-aa")
 => "foo-aa"
```

which is why it worked when I tested that case.

ttaa9: Ah, it is because regex substitution does this I believe:
```
2.2.4 :036 > /#{''}$/
 => /$/ 
```

glatard: oops...

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -121,7 +121,7 @@ def initialize(attributes)
     # encapsulated CbrainTask in a version switcher class to allow different
     # CbrainTask classes for each tool version.
     # Returns the newly generated CbrainTask subclass.
-    def integrate(register: true, create_tool_config: false, multi_version: false)
+    def integrate(register: true, create_tool_config: true, multi_version: false)
```
glatard: BTW, why don't we simply remove this option now that it's properly handled?

glatard:
Thanks for the fixes @ttaa9, your code now looks all good to me. Before merging, I just have a doubt that requires @prioux's opinion. @prioux:
* For some reason, @ttaa9's PR contains 2 of your latest commits: 875ccbd and 52639f4.
* These 2 commits are already merged in dev.
* For some reason, the content of these commits shows in this PR's diff. Do you have any idea why? Do you think it's a problem?

glatard:
So, following discussion with @prioux, it seems that the following happened:
* @ttaa9 created the PR from his dev to cbrain/dev.
* The two commits were merged into cbrain/dev.
* @ttaa9 pulled cbrain/dev into his dev branch.
* The 2 commits were added to the PR because it was still open. They really shouldn't have been, it looks like a bug in Github. 

@ttaa9, to make sure that everything is clean, could you close this PR and re-open a new one? @prioux's commits shouldn't be in the new one (before you do that, don't forget to look at my other comment on the create_tool_config argument above).

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
ttaa9: I am not against that, since it is indeed a bit strange. Although, we could also call it something like "attempt_tool_config_creation", so that someone could still specify not to make one.

PR #399, opened 2016-06-13
Fixed seeding DB to use only generated Bourreau instances. Added indentations.

crocodoyle:


BrainPortal/db/seeds_dev.rb
```diff
@@ -806,7 +806,7 @@ def seeds_dev_tmp_file(content)
 User.all.each do |user|
   groups = [ user.own_group ] + user.groups.where( :type => 'WorkGroup' ).all
   groups.each do |group|
-    Bourreau.all.each do |bourreau|
+    [ main_bourreau, pember_bourreau, longbourne_bourreau ].each do |bourreau|
```
prioux: We discovered this change was necessary because it made the seeding work even when a Bourreau was already present; however now that I think of it we should go over the rest of this file and also fix other places where we have `Bourreau.all` or `User.all` etc. Let's make the seeding process only affect its own objects.

BrainPortal/db/seeds_dev.rb
```diff
@@ -639,39 +639,39 @@ class CBRAIN
 
 Bourreau.all.each do |bourreau|
```
prioux: See comment below.

BrainPortal/db/seeds_dev.rb
```diff
@@ -806,7 +806,7 @@ def seeds_dev_tmp_file(content)
 User.all.each do |user|
```
prioux: See, comment below.

crocodoyle:
I changed the other steps so that none of them use .all to choose what data to seed.

PR #401, opened 2016-06-14
Renamed tool attribute cbrain_task_class

prioux:
DO NOT MERGE THIS PULL REQUEST; it's fine but its impact require careful scheduling.

The ActiveRecord attribute `cbrain_task_class` in Tools has been renamed
`cbrain_task_class_name`.

The name was misleading, the attribute contains a class name,
not the class itself. This has allowed me to add a new method
`cbrain_task_class` to tool to return the class itself.

Please review the changes carefully to make sure all the affected code seems to be equivalent to what it was.

### DIFF
--- a/BrainPortal/app/models/tool.rb
+++ b/BrainPortal/app/models/tool.rb
@@ -25,7 +25,7 @@
 #
 #=Attributes:
 #[*name*] The name of the tool.
-#[*cbrain_task_class*] CbrainTask subclass associated with this tool.
+#[*cbrain_task_class_name*] The name of the CbrainTask subclass associated with this tool.
 #[*user_id*] The owner of the tool.
 #[*group_id*] The group that the tool belongs to.
 #[*category*]  The category that the tool belongs to.
@@ -46,8 +46,8 @@ class Tool < ActiveRecord::Base
 
   before_validation :set_default_attributes
 
-  validates_uniqueness_of :name, :select_menu_text, :cbrain_task_class
-  validates_presence_of   :name, :cbrain_task_class, :user_id, :group_id, :category, :select_menu_text, :description
+  validates_uniqueness_of :name, :select_menu_text, :cbrain_task_class_name
+  validates_presence_of   :name, :cbrain_task_class_name, :user_id, :group_id, :category, :select_menu_text, :description
   validates_inclusion_of  :category, :in => Categories
   validates_format_of :url, :with => URI::regexp(%w(http https)), :if => :url_present?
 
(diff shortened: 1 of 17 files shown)


BrainPortal/app/models/tool.rb
```diff
@@ -58,30 +58,25 @@ class Tool < ActiveRecord::Base
...
-  def select_random_bourreau_for(user)
-    available_group_ids = user.group_ids
-    bourreau_list = Bourreau.where( :group_id => available_group_ids, :online => true ).select(&:is_alive?)
-    bourreau_list = bourreau_list & self.bourreaux
-    if bourreau_list.empty?
-      cb_error("Unable to find an execution server. Please notify your administrator of the problem.")
-    end
-    bourreau_list.slice(rand(bourreau_list.size)).id
-  end
-
```
glatard: Any reason why this was removed? Not used any more?

glatard:
Looks good to me. I guess the only way to be sure that no side effect was introduced by the renaming is to run the tests.

BrainPortal/app/models/tool.rb
prioux: Yes, it was a method used by the old interface, and not used at all anymore anywhere.

prioux:
Just ran the test suite... three times in randomized order. All OK.
PLEASE DO NOT MERGE RIGHT AWAY, since this changes the database schema and we won't be able to run the official CBRAIN service and the DEV service at the same time. I'll merge myself before a major update of the LIVE service.

BrainPortal/app/views/tools/_tools_table.html.erb
```diff
@@ -112,12 +112,12 @@
     end
 
     t.column("Help", :help) do |tool|
-      path = tool.cbrain_task_class.constantize.public_path("edit_params_help.html") rescue nil
+      path = tool.cbrain_task_class.public_path("edit_params_help.html") rescue nil
```
natacha-beck: Maybe I'm wrong, but should be `cbrain_task_class_name` ?

prioux: Nope, it's correct.

In the old code, there was a method `cbrain_task_class()` which returned a string. It needed to be `constantize()d` to create the class on which to invoke `public_path()`.

In the new code, `cbrain_task_class()` returns the class itself. I could have changed the code to

```
path = tool.cbrain_task_class_name.constantize.public_path("edit_params_help.html") rescue nil
```

but that would have been ugly, especially since like I said the new `cbrain_task_class()` already does that.

prioux:
This branch will be rebased just before doing a release.

prioux:
I ran all the tests, and adjusted a few things, and all looks good.

PR #402, opened 2016-06-14
Boutiques features and bug fixes

ttaa9:


BrainPortal/app/models/cluster_task.rb
```diff
@@ -329,6 +329,26 @@ def safe_userfile_find_or_new(klass,attlist)
     return klass.new(attlist)
   end
 
+  # Returns whether or not the specified userfile exists
+  # Should at least specify the :name and :data_provider_id attributes.
+  # But the latter is given a default value if not.
+  def userfile_exists(klass,attlist)
```
prioux: Someone coudl wonder why this method is part of the CbrainTask::ClusterTask class; add in the documentation that it's meant to help task developers create safe code for creating output files, by filling many of the attributes of the userfiles with defaults taken from the task itself (e.g. task's owner, task's group, task's results_data_provider_id etc). The method above, safe_userfile_find_or_new did exactly this.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1668,6 +1688,11 @@ def docker_executable_name
     return RemoteResource.current_resource.docker_executable_name.presence || "docker"
   end
 
+  # Return whether the Bourreau has docker present or not
```
prioux: the description shoudl say "the bourreau of the task".

BrainPortal/app/views/tools/_tools_table.html.erb
```diff
@@ -112,7 +112,8 @@
     end
 
     t.column("Help", :help) do |tool|
-      path = tool.cbrain_task_class.constantize.public_path("edit_params_help.html") rescue nil
+      path   = tool.cbrain_task_class.constantize.public_path("edit_params_help.html") rescue nil
+      path ||= tool.cbrain_task_class.constantize.help_filepath rescue nil # Boutiques help-page
```
prioux: Where would the help_filepath come from or point to? It can only be part of static assets for the site, right? Not sure I get this change.

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -166,6 +165,20 @@ def integrate(register: true, create_tool_config: false, multi_version: false)
         })[partial]
       end
 
+      # Write out a help file for this Boutiques task
+      helpFileName = name + "_help.html"
```
prioux: Creating files in the code bas at run time is not allowed... what if the code is deployed under a UNIX user ID different from the USER that runs the Rails app ?

glatard: This is executed by BrainPortal though (the rails app).

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -318,25 +340,89 @@
...
+%   # Write info for ungrouped parameters
+    <%- for param in required do parameter.(param) if whichGroup.(param["id"]).nil? end -%>
+    <%- for param in optional do parameter.(param) if whichGroup.(param["id"]).nil? end -%>
+%   # Write info for grouped parameters
+    <%- groups.each(&writeGroup) unless noGroups  -%>
   </ul>
   <%- end -%>
 </div>
 
+%# Style block for group toggling checkbox
```
prioux: Should style be embedded or put in a separate file in the CBRAIN framework?

e.g. `BrainPortal/public/stylesheets/boutique.css`

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
prioux: That's what I'm saying. The BrainPortal app might be running as a user that cannot write to the file system where its code is.

Normally, such 'additions' to the code base is done through rake tasks.

ttaa9: I believe we talked about this before, but is there no way use them in memory? For instance, in _tools_table.html.erb, using something other than overlay_ajax_link for Boutiques tools? 

If it was a rake task, I suppose it would be an addition to the "plugins:install" command?

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -166,6 +165,12 @@ def integrate(register: true, create_tool_config: false, multi_version: false)
         })[partial]
       end
 
+      # Add a helper method for accessing the help file (for use on the tools page)
+      # Written by the rake task cbrain:plugins:install, within the subtask public_assets
+      helpFileName = name + "_help.html" 
+      helpFileDir  = File.join( "cbrain_plugins", "cbrain_tasks", "help_files/" )
```
prioux: I don't get this... what is this path "cbrain_plugins/cbrain_tasks" ? It doesn't correspond to the plugins structure?

ttaa9: I am adding a method to the task class that points to where the help files should be written out by the rake task. Then I call that method in the Tools table (to make the help button). I thought it was a bit better than hard-coding the path, in case we want to call the method else where (e.g. tests). 

Because Boutiques tools are parts of plugins, I put the help files there, under the 'public' dir (which, annoyingly to me, is implict in rails paths like this).

PR #404, opened 2016-06-14
Add Dockerfile, that install BrainPortal and Bourreau on centos5 container.

natacha-beck:
This is a first step, to implement automatic test. Here we only create a centos5 container, with the BrainPortal and the Bourreau setting up. 

We need to add a layer with a DB installation.

### DIFF
--- a/BrainPortal/Gemfile
+++ b/BrainPortal/Gemfile
@@ -36,7 +36,6 @@ gem "rails_autolink"
 gem "pbkdf2-ruby"
 gem "json-schema"
 gem "test-unit",    "~> 3.1"
-gem 'eventmachine', '< 1.2'
 
 group :development do
   gem "wirble"
--- a/BrainPortal/Gemfile
+++ b/BrainPortal/Gemfile
@@ -36,6 +36,7 @@ gem "rails_autolink"
 gem "pbkdf2-ruby"
 gem "json-schema"
 gem "test-unit",    "~> 3.1"
+gem 'eventmachine', '< 1.2'
 
 group :development do
   gem "wirble"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+FROM centos:5
+
+RUN yum update  -y 
+RUN yum install -y wget
+RUN wget http://download.fedoraproject.org/pub/epel/5/x86_64/epel-release-5-4.noarch.rpm
+RUN rpm -ivh epel-release-5-4.noarch.rpm
+
+RUN yum update -y  && yum install -y \
+        git openssh \
+        mariadb-devel \
+        mariadb  \
+        libyaml-devel \
+        glibc-headers \
+        autoconf \
+        gcc-c++ \
+        glibc-devel \
+        patch \
+        readline-devel \
+        libffi-devel \
+        make \
+        bzip2 \
+        automake \
+        libtool \
+        bison \
+        sqlite-devel \
+        libxml2 \
+        libxml2-devel \
+        libxslt \
+        gpg \
+        which \
+       openssl-devel \
+       mysql-devel
+
+RUN useradd cbrain
+COPY . /home/cbrain/cbrain
+RUN chown cbrain:cbrain -R /home/cbrain/cbrain
+USER cbrain
+
+RUN gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
+RUN cd $HOME && \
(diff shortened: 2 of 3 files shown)


prioux:
Does this Dockerfile absolutely need to be in the root of the CBRAIN project's directory? Shouldn't we isolate all docker stuff into its own subdirectory?

natacha-beck:
If I'm not wrong it should be in the root directory if we want to trigger
automated builds. But I can take a second look and see if we have better
way to do.

driusan:
It doesn't need to be in the root to trigger automated builds, you can configure it to build from a different directory on docker hub, but `docker build` is effectively (if not in reality?) chroot'ed to the directory that it's run from, so it can't copy any files from a higher (or parallel) directory to the Dockerfile into the Docker image.

BrainPortal/Gemfile
```diff
@@ -36,6 +36,7 @@ gem "rails_autolink"
 gem "pbkdf2-ruby"
 gem "json-schema"
 gem "test-unit",    "~> 3.1"
+gem 'eventmachine', '< 1.2'
```
glatard: is this really required?

Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM centos:5
+
+RUN yum update  -y 
+RUN yum install -y wget
+RUN wget http://download.fedoraproject.org/pub/epel/5/x86_64/epel-release-5-4.noarch.rpm
+RUN rpm -ivh epel-release-5-4.noarch.rpm
+
+RUN yum update -y  && yum install -y \
```
glatard: yum update is already done at line 3. You could also add wget to this list instead of doing it at line 4.

Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM centos:5
```
glatard: I don't like that we're using centos:5 here. Is there any good reason for not using at least centos 6?

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+RUN gem install bundler
+RUN cd $HOME/cbrain/BrainPortal    && \
+    bundle install                 && \
+    cd `bundle show sys-proctable` && \
+    rake install
+RUN cd $HOME/cbrain/BrainPortal    && \
+    rake cbrain:plugins:install:all
+RUN cd $HOME/cbrain/Bourreau       && \
+    bundle install                 && \
+    rake cbrain:plugins:install:plugins
```
glatard: Could be in a different RUN command, as done for the portal.

Dockerfile
prioux: If this docker image is meant to run tests, we should use the latest stable CENTOS, in fact.

glatard:
The CentOS6 version looks good to me. @prioux, what's your motivation for *not* putting the Dockerfile at the root of the repo? It's very convenient because:
* it allows to configure automatic builds of containers for the different CBRAIN branches (in particular master, dev) on DockerHub whenever something is pushed to these branches. We may manage to do that with github hooks or any other trick, but why bother when it's so simple to just put the Dockerfile at the root?
* it allows to sync the Dockerfile with the content of a particular CBRAIN branch. For instance, if two branches require different Dockerfiles (e.g. different Ruby or Rails versions), we just have to make sure that the right Dockerfile is in the right branch, which happens naturally when the Dockerfile is stored in the code repo. Again, we may synchronize two repos manually, but why would we go into this hassle?

prioux:
What is the Dockerfile FOR ? Is it to start a docker instance to run tests? Is it to package CBRAIN has a docker application? Is it both? Shouldn't we have seperate dockerfiles for separate functions? The things with "Dockerfile" is that there can only be one such file in the root of the directory tree, even when we want the same code base to be used for different things.

natacha-beck:
This one should be the base of CBRAIN.

I will create an other one on top of this one for testing purpose.

glatard:
@prioux This Dockerfile can be useful for many different things. In the short term, @natacha-beck will use it for the tests and I will use it to install CBRAIN on other infrastructures at HBM (someone explicitly requested a Docker container for that). For these two applications we should be able to use the same Dockerfile. In the future, too keep things tractable, we should also try to keep a single Dockerfile as much as possible. 

The only reason that I see why we could need different Dockerfiles is if we need containers built for different distributions (centos, ubuntu, etc). But I'm not sure we're going there yet. If that's fine with you, let's start simple with a single Dockerfile and we'll think more when we need more. It's just a text file to add/remove anyway, and it's really not uncommon to store it at the root of the Git repo, many people do that.

prioux:
Okay.

Issue #397, opened 2016-06-06
All models to get proper separate "create" pages


prioux:
Most of our index pages have `Create New XYZ` buttons that pop up an overlay with the form.

Change this so that we get a proper, separate page for creating a new object.

I've retrofitted it this way for the Users already; the AccessProfiles were created correctly right from the start.

crocodoyle:
I started looking into this this morning before you assigned me another bug @prioux. Should I continue?

prioux:
Yes, absolutely. It's a good project.

prioux:
All done by Andrew in #411

PR #411, opened 2016-06-21
Remove overlays from new

crocodoyle:
The following new objects now use a separate page to create objects instead of an overlay:
-DataProviders
-Servers
-Tools
-Sites
-Feedbacks
-Groups
-Messages

Removed obsolete  is_bourreau/is_brainportal checks from data_providers

Fix for issue #397


BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -103,16 +101,16 @@ def create #:nodoc:
...
       respond_to do |format|
-        format.js   { redirect_to :action => :index, :format => :js  }
+        format.html { redirect_to :action => :index, :format => :html}
         format.xml  { render :xml   => @provider }
         format.json { render :json  => @provider }
       end
     else
       @typelist = get_type_list
       respond_to do |format|
-        format.js   { render :partial  => "shared/failed_create", :locals => {:model_name => "data_provider"} }
```
prioux: Can you double-check something? The partial "shared/failed_create" might not bee needed anymore at all.. if it's not referenced anywhere we can delete the file.

BrainPortal/app/views/tools/_tools_table.html.erb
```diff
@@ -24,7 +24,7 @@
 
 <div class="menu_bar">
   <% if current_user.has_role?(:admin_user) %>
-    <%= new_model_button 'Create New Tool', new_tool_path %>
+    <%= link_to 'Create New Tool', new_tool_path, :class => "button menu_button" %>
```
prioux: class to be just `button` maybe ? compare with the others?

prioux:
All good. Thank you.

Issue #395, opened 2016-06-06
Update project switching menu


prioux:
The top `switch project` menu is ugly and a bit confusing.

crocodoyle:
How's this?
![screen shot 2016-06-22 at 10 22 08 am](https://cloud.githubusercontent.com/assets/345229/16270154/4b8af7cc-3863-11e6-9a75-0e5f8fdbdeda.png)

prioux:
Fixed by combining Andrew's changes and a few mods of my own.

PR #413, opened 2016-06-22
changed projects dropdown menu to be a little nicer

crocodoyle:
It uses buttons now and has a nicer border. Fun!


prioux:
Related to #395

Issue #368, opened 2016-05-05
The DataProvider index page should not do a "is_alive?" check automatically.


prioux:
The entries in the table should have a "check" button that triggers the check for that data provider.
Maybe add a "check all" button to get the current behavior.

crocodoyle:
Can I work on this?

crocodoyle:
Is it particularly important that the data providers be checked sequentially, or can they all be checked simultaneously?

prioux:
It's better one at a time, otherwise if n DPs out of m are inaccessible, they jam Rails instances and decrease the site's availability

prioux:
Fixed, thanks.

PR #418, opened 2016-06-28
Remove auto alive-checking of DP's, replace with manual checks

crocodoyle:
There is now a "Check All" button at the top of the page which will replicate the previous behaviour, and individual Data Providers can be checked by clicking the "check" link within the table.

Fix for #368

### DIFF
--- a/BrainPortal/app/views/data_providers/_data_providers_table.html.erb
+++ b/BrainPortal/app/views/data_providers/_data_providers_table.html.erb
@@ -33,6 +33,8 @@
     <%= link_to "Create New Data Provider", new_data_provider_path, :class => "button menu_button" %>
   <% end %>
 
+  <%= link_to "Check All", nil, {:class => "button check_all_dp",} %>
+
   <%= link_to "User Access Report",
              {:controller  => :data_providers, :action  => :dp_access},
               :datatype  => 'script', :class  => "button"
--- a/BrainPortal/app/views/data_providers/_one_data_provider_table.html.erb
+++ b/BrainPortal/app/views/data_providers/_one_data_provider_table.html.erb
@@ -76,8 +76,24 @@
   %>
 
   <% t.column("Alive?", :alive) do |dp| %>
-    <% if dp.online? %>
-      <div id="<%= "alive" + h(dp.id) %>"><%= html_colorize("(...)","magenta") %></div>
+      <% if dp.online? %>
+        <% if dp.alive_cached_valid? %>
+          <%= red_if( !dp.meta[:alive_cached], "Yes", "No") %>
+        <% else %>
+          <span class="dp_alive_btn" id="<%= "alive" + h(dp.id) %>">
+          <%= on_click_ajax_replace( { :element   => "span",
+                                                     :url           => url_for(:controller => :data_providers, :action  => :is_alive, :id => dp.id),
+                                                     :replace   => "alive" + h(dp.id)
+                                                   },
+                                                   {
+                                                     :before  => "<span class='loading_message'>Loading...</span>",
+                                                     :class     => "dp_alive_btn"
+                                                   }
+                                                  ) do %>
+            <a href="#" class="action_link">Check</a>
+          <% end %>
+          </span>
+        <% end %>
     <% else %>
       No
     <% end %>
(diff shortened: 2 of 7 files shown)


BrainPortal/app/views/data_providers/_one_data_provider_table.html.erb
```diff
@@ -76,15 +76,19 @@
   %>
 
   <% t.column("Alive?", :alive) do |dp| %>
-    <% if dp.online? %>
-      <%= staggered_loading "span", url_for(:action  => :is_alive, :id  => dp.id), :error_message  => "<span class=\"warning\">???</span>" do %>
-        <span class="loading_message"><%= html_colorize("(...)","magenta") %></span>
+    <%= on_click_ajax_replace( { :element  => "div",
```
prioux: If the DP is not online (an attribtue of the model) we don't nee dto do a check, we can already fill the row's information, That's why there was a IF int he original code.

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -1024,14 +991,57 @@
...
 
     $(document).delegate(".ajax_onclick_show_element", "click", ajax_onclick_show);
-    $(document).delegate(".ajax_onclick_hide_element", "click", ajax_onclick_hide);    
-    
-    // Allows to submit an interval of two dates, uses 
-    // datepicker of jquery-ui, see:  
+    $(document).delegate(".ajax_onclick_hide_element", "click", ajax_onclick_hide);
+
+
+    $(document).delegate(".check_all_dp", "click", function (event) {
```
prioux: Add a bit of documentation above here?

BrainPortal/app/views/data_providers/_data_providers_table.html.erb
```diff
@@ -33,6 +33,8 @@
     <%= link_to "Create New Data Provider", new_data_provider_path, :class => "button menu_button" %>
   <% end %>
 
+  <%= link_to "Check All", nil, {:class => "button check_all_dp",} %>
```
prioux: Behavior: if I have already checked manually some data providers in my page using their 'check' link, and I click on 'Check All', they will be rechecked again, even though the information is already on the page.

BrainPortal/app/views/data_providers/_one_data_provider_table.html.erb
```diff
@@ -76,15 +76,19 @@
...
+                                                :url      => url_for(:controller => :data_providers, :action  => :is_alive, :id => dp.id)
+                                             },
+                                             {
+                                               :before   => "<div class='loading_message'>Loading...</div>",
+                                               :class => "dp_alive"
+                                             }
+                                            ) do %>
+      <a>check</a>
```
prioux: make the link style match the active links we have in such tables, see how the 'browse' link appear for instance.

prioux: Also, maybe the check link would be better in the "operations" column" ?

BrainPortal/app/views/data_providers/_one_data_provider_table.html.erb
crocodoyle: The is_alive action only checks if it's alive if the DP is online, so I thought it was redundant.

prioux: It's still doing a full round-trip to the server side. The original page already filled all the rows for offline data providers with `No` in the "Is Alive" column and didn't even check them when doing its staggered loading.

BrainPortal/app/views/data_providers/_one_data_provider_table.html.erb
crocodoyle: The browse links appear the same, with an href attribute, which we don't want here because it does not go to a separate page. I'm not sure what you mean.

prioux:
The code all looks good Andrew, but I'll request some new changes after seeing how it works:

1- move the "check" link into the table, under the alive column (we no longer show the "(...)" purple thingie)
2- cache the result in the controller for 1 minute; this can be done by storing two metadata values on the current DP being checked when the controller action `is_alive` is initially invoked; one with the true/false result and one with a timestamp. When the action is invoked again a check for these meta values is done and if valid (within a minute) then the is_alive values stored in the metadata is used instead of calling `@data_provider.is_alive?` (which is costly)

Issue #382, opened 2016-05-26
Create tests for Boutiques


glatard:
Create a set of tests for the Boutiques code generator.

- [ ] Write a test application (e.g. a bash script) that uses all the features in a Boutiques descriptor, prints information about its parameters and returns an error when the syntax of the command line is not correct. For instance, "test.sh -a" would print "INFO: Flag 'a' is active" and return 0, while "test.sh -b" would print "ERROR: input 'b' requires a string" and return a non-zero exit code. 
- [ ] Write a JSON descriptor for this test script. 
- [ ] Write test assertions, including but not limited to:
- The JSON descriptor validates
- CBRAIN can generate a tool from it
- The generated CBRAIN tasks succeeds or fails depending on specific test inputs. For each of the test inputs, submit a task to CBRAIN through the API with the expected final status (Failed or Completed) in the description field. The format used for the test inputs and the code to submit the tasks could be based on a more general version of https://github.com/glatard/cbrain-tests-fsl-melodic that was developed specifically for FSL Melodic. Ideally, the test inputs shouldn't contain any CBRAIN-specific information. CBRAIN-specific information (e.g. tool config to be used) should be in a configuration file or parameter of the test submission script. 

The test script, JSON descriptor and test inputs could be uploaded and documented in the Boutiques github repo. The (Ruby) script to run the tests in CBRAIN should be in CBRAIN's repo. In a longer-term, this script could be generalized to test any tool in CBRAIN (not just the Boutiques test application).

@natacha-beck may have some ideas on how to make this fit in the CBRAIN testing framework.

ttaa9:
So we decided to something like: (1) check json descriptor validates, (2) check local script behaviour, (3) test behaviour in cbrain via local qsub scripts, and (4) run some tasks via the cbrain api, should it be present.

Could you expand a little on how (3) should work (and perhaps which methods/classes it should use)?

glatard:
This is what I would do:

```
t=CbrainTask::FslBet.new # Change to the class that implements your tool.
t.params=Hash.new
t.params[:inputfile_id]=2   # Assign all the required parameters.
t.cluster_commands # write these commands in a bash script, run the script, check the exit code. Be careful, method 'cluster_commands' is only available Bourreau-side.
```

ttaa9:
@prioux has suggested we might not need to use the api; we could instead test the generated methods individually.

glatard:
[...]

PR #420, opened 2016-06-29
Boutiques test suite: initial version

ttaa9:
For #382. 
Includes:
- boutiquesTestApp.rb: a mock application with as many Boutiques features as possible
- descriptor_test.json: the Boutiques descriptor for the mock program
- boutiques_tester_spec.rb: the rspec file containing the tests. There are two: one for the Bourreau side and one for the BrainPortal side.
- test_helpers.rb: contains helper methods for the tests on both sides (including a compact form of many of the tests)

Some issues:
- The automatic transaction rollback does not seem to work consistently on the Bourreau side, so right now it is handled hackily
- I have pasted the Boutiques ruby json validator into the tests folder. That is probably not where it belongs, but since it is nowhere else in cbrain (that I know of), I have put it there for now. It does not do anything special right now, but I suspect we will be adding to it in the near future with extra checks. When that happens it should be used to validate all the json descriptors at runtime (instead of just the internal validation). It is also unfortunate that it will be in two separate repos, but then again so is the main Boutiques schema. Let me know what you think of this; I can remove it fairly easily if needed.
- I'm sure there are tests for things that I missed


Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+
+# Add helper methods for performing tests
+include TestHelpers
+
+# Testing the Boutiques framework on the Bourreau side
+describe "Bourreau Boutiques Tests" do
+
+  # Run before block to create required input files
+  before(:all) do
+    FileUtils.touch('c')  # For -C
```
prioux: Where are these files created? Why not create a temp dir in /tmp and put all those fakes files there?

Bourreau/spec/rails_helper.rb
```diff
@@ -26,7 +26,7 @@
   database      = config["database"] || "nodatabase"
   password    &&= ":#{password}"
   url           = "#{adapter}://#{username}#{password}@#{host}/#{database}"
-  ENV["DATABASE_URL"] = url
+  ENV["DATABASE_URL"] = URI.escape url
```
prioux: This change was already in my pull request of last week... I suppose git will just merge them as-is, since they're identical.

Bourreau/spec/boutiques/boutiques_tester_spec.rb
natacha-beck: You can take a look to what is done in `userfiles_controller_spec`, in this one we also create a temporary file, and removed it at the end.

Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+        # Initial file should still be there with the correct, original name
+        expect( @task.userfile_exists(@userfileClass, @reqOutfileProps) ).to be true
+        expect( Userfile.where( :name => @fname ).first.nil? ).not_to be true
+        # New file should exist and have been renamed appropriately
+        expect( Userfile.all.one? { |f| (f.name.to_s =~ renamedFileRegex) == 0 } ).to be true
+      end
+
+    end # End output file handling tests
+
+    # Test that a Tool Config is created iff both the bourreau and the descriptor specify docker
```
natacha-beck: Typo here.

BrainPortal/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+
+# Run the Boutiques test on the BrainPortal side
+describe "BrainPortal Boutiques Tests" do
+
+  # Run before block to create required input files
+  before(:all) do
+    FileUtils.touch('c')  # For -C
+    FileUtils.touch('f')  # For -d
+    FileUtils.touch('jf') # For -j
+    [1,2].each { |i| FileUtils.touch("f#{i}") } # For -f
```
natacha-beck: Take care of the location of the new file, as before.

Bourreau/spec/boutiques/boutiques_tester_spec.rb
ttaa9: :)  iff = if and only if = ssi

BrainPortal/lib/tasks/seed_test_bourreau.rake
```diff
@@ -0,0 +1,37 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+namespace :db do
+  namespace :seed do
+    namespace :test do
+      desc "Seed CBRAIN test DB for Bourreau testing"
+      task :bourreau => :environment do
+        raise "Error: this task must be run in a TEST environment!" unless
```
natacha-beck: I think you can add the common command `rake db:seed:test:bourreau RAILS_ENV=test` just to be more clear.

Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+        # Delete any generated output files
+        destroyOutputFiles
+        # Also need to get rid of the (simulated) exit file
+        fename = '.qsub.exit.BoutiquesTest.-1'
+        File.delete( fename ) if File.exists?( fename )
+        # Destroy the registered userfiles and the data_provider, so as not to affect downstream tests
+        # Necessary for tests that create and/or register userfiles in the it block
+        Userfile.all.each{ |uf| uf.destroy }
+        DataProvider.all.each { |dp| dp.destroy }
+      end
```
natacha-beck: It seems useless to destroy the Userfile and the DataProvider, FactoryGirl already take care of that.

Bourreau/spec/factories/bourreau_factories.rb
```diff
@@ -0,0 +1,270 @@
...
+  factory :brain_portal, parent: :remote_resource, class: BrainPortal do
+    sequence(:name) { |n| "bp_#{n}"}
+  end
+
+  factory :bourreau, parent: :remote_resource, class: Bourreau do
+    sequence(:name) { |n| "bourreau_#{n}"}
+    cms_class       "ScirSge"
+  end
+
+
```
natacha-beck: This file is a copy of the portal_factories without the CustomFilter block. If I well understand you only did that because Custom filters are not defined on the Bourreau side. 

I personally will create a symlink `bourreau_factories.rb` that point to `portal_factories` and add a `begin .. rescue` block where needed. Like that 

```ruby 
  begin
    factory :userfile_custom_filter, parent: :custom_filter, class: UserfileCustomFilter do
      sequence(:name) { |n| "ucf_#{n}"}
    end
  rescue
    puts "On the Bourreau side of test, we do not need to create Userfile custom_filter"
  end

  begin
    factory :task_custom_filter, parent: :custom_filter, class: TaskCustomFilter do
      sequence(:name) { |n| "tcf_#{n}"}
    end
  rescue
    puts "On the Bourreau side of test, we do not need to create Task custom_filter"
  end
```

glatard:
I don't really like the copy-paste of the validator script, it is even already outdated by https://github.com/boutiques/tools/pull/20. What about using a git submodule? https://git-scm.com/docs/git-submodule

Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+          expect( @boutiquesTask.descriptor['name'] ).to eq( 'BoutiquesTest' )
+        end
+      end
+
+      # Test some basic properties
+      it "creates a tool" do
+        expect( Tool.exists?( :cbrain_task_class => @task_const.to_s ) ).to be true
+      end
+
+      it "makes the tool accessible" do
```
glatard: I would call this "makes the tool accessible as an CbrainTask object"

Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+        end
+
+        it "handles substitutions with command-line flags" do
+          s = @task.apply_template(@template, @def_keys, flags: @flags)
+          expect( s.strip ).to eq( "cmd -1 one.txt --long-flag 2 -t t.csv" )
+        end
+
+        it "handles substitution with flag-type inputs" do
+          s = @task.apply_template(@template, @def_keys.merge({'[4]' => true}), flags: {'[4]' => '-f'})
+          expect( s.strip ).to eq( "cmd one.txt 2 t.csv -f" )
```
glatard: You could as well test what happens when the flag is false.

Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+        # Destroy any tools/toolconfigs for the tool, if any exist
+        ToolConfig.where(tool_id: @task_const_name.constantize.tool.id).destroy_all
+        Tool.where(:cbrain_task_class => @task_const_name).destroy_all
+        # Ensure the Bourreau does not have docker installed by default
+        resource = RemoteResource.current_resource
+        resource.docker_present = false
+        resource.save!
+      end
+
+      it "is not done with only docker in task" do
```
glatard: you mean "tool config is not created when bourreau does not support docker"?

Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+        @boutiquesTask.descriptor['docker-image'] = 'placeholder_string'
+        @boutiquesTask.integrate if File.exists?(@descriptor)
+        expect( ToolConfig.exists?( :tool_id => @task_const_name.constantize.tool.id ) ).to be false
+      end
+
+      it "is not done with only docker in bourreau" do
```
glatard: you mean "tool config is not created when descriptor has no docker image"?

Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+        RemoteResource.current_resource.docker_present = true
+        RemoteResource.current_resource.save!
+        @boutiquesTask.integrate if File.exists?(@descriptor)
+        expect( ToolConfig.exists?( :tool_id => @task_const_name.constantize.tool.id ) ).to be false
+      end
+
+      it "is done with docker in both task and bourreau" do
```
glatard: you mean "tool config is created when descriptor has docker image and bourreau has docker"?

Bourreau/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,389 @@
...
+        resource       = RemoteResource.current_resource
+        resource.docker_present = true
+        resource.save!
+        @boutiquesTask.integrate if File.exists?(@descriptor)
+        expect( ToolConfig.exists?( :tool_id => @task_const_name.constantize.tool.id ) ).to be true
+      end
+
+      # When neither criteria is met, check that a ToolConfig is not made
+      # Also checks that the integration changes were rolled back
+      it "is not done without docker in task and bourreau" do
```
glatard: same type of comment

BrainPortal/spec/boutiques/boutiquesTestApp.rb
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env ruby
```
glatard: Add license header.

BrainPortal/spec/boutiques/boutiquesTestApp.rb
```diff
@@ -0,0 +1,210 @@
...
+# Colours
+class String
+  def colour(code) "\e[#{code}m#{self}\e[0m" end
+  def red() colour(31) end
+  def green() colour(32) end
+  def blue() colour(34) end
+end
+
+# Print raw input string
+print("\nRunning Boutiqes simple test app\nRaw Input=#{ARGV.to_s}".blue) if ARGV.include? "--verbose"
```
glatard: typo: Boutiques

BrainPortal/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/env ruby
```
glatard: add license header.

BrainPortal/spec/boutiques/test_helpers.rb
```diff
@@ -0,0 +1,184 @@
+
```
glatard: add license header.

BrainPortal/spec/boutiques/validator.rb
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/ruby
```
glatard: See comment about validator in general conversation: shall we use a git submodule? https://git-scm.com/docs/git-submodule

Bourreau/spec/boutiques/boutiques_tester_spec.rb
ttaa9: The way rspec structures these is that the 'it' block is appended to the 'contexts' or 'descriptions'. So it is actually:
"Default ToolConfig creation is not done with _only_ docker in task", which means the same thing. But yours might be clearer.

ttaa9:
So is your suggestion to have both the boutiques/schema and the boutiques/tools be git submodules of cbrain? Seems like a good idea. Would it make it more difficult to install and update cbrain though?

glatard:
I think only the validator (boutiques/tools) should be. The schema is too critical, it would break everything if it's not properly synchronized. While if the validator is not there, only the tests will break.

ttaa9:
I was thinking that the cbrain runtime validation of Boutiques tools could also use the validator script. Right now it just validates descriptors (e.g. from cbrain plugins neuro) against the json schema. But yes, it could make things more brittle.

glatard:
OK, that makes sense. In that case, we should depend on a particular tag of the boutiques repos (schema and tools), to make sure that this won't change over time.

ttaa9:
I think (hope) that addresses the comments. I also added an issue to remind us to look at the submodule idea after the Boutiques release.

Issue #365, opened 2016-05-05
Upload file with space raise an ugly error message. 


natacha-beck:
When a user tries to upload a file that file name  contains white-space the platform raises an error, and the following flash message appears:

> 
> Error sending background request:
> HTTP 422 Unprocessable Entity
> The CBRAIN administrators have been alerted about this problem.

This should be more explicit.

crocodoyle:
Working on this bug now, fyi @prioux

prioux:
Okay thanks.

prioux:
Closed, thanks Andrew.

prioux:
Wasn't this issue closed by #421 ?

PR #421, opened 2016-06-29
Added check for spaces in filename on upload

crocodoyle:
Reorganized the upload-dialog javascript because there were many functions that set the disabled status of the "Upload" button, which produced unreliable behaviour.

Fix for #365 .

### DIFF
--- a/BrainPortal/public/javascripts/userfiles.js
+++ b/BrainPortal/public/javascripts/userfiles.js
@@ -733,12 +733,14 @@ $(function() {
             var warning_text = "";
             var max = parseInt($('#upload-dialog').data('max-upload-size'));
             var bad_file;
+            var filename = $(this).prop('files')[0].name;
             var selected = !!$(this).val();
 
             // same regex as the userfiles model validation
-            var file_pattern = /^[a-zA-Z0-9][\w\~\!\@\#\%\^\&\*\(\)\-\+\=\:\[\]\{\}\|\<\>\,\.\?]*$/;
-            var bad_chars = $(this).val().match(file_pattern);
-            var spaces_in_name = ($(this).val()).includes(" ");
+            var allowed_file_pattern = /^[a-zA-Z0-9][\w\~\!\@\#\%\^\&\*\(\)\-\+\=\:\[\]\{\}\|\<\>\,\.\?]*$/;
+            var bad_chars = !allowed_file_pattern.test(filename);
+
+            var spaces_in_name = filename.includes(" ");
 
             var file_too_big;
             if ( max > 0 ){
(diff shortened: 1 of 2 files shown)


BrainPortal/public/javascripts/userfiles.js
```diff
@@ -722,25 +722,45 @@ $(function() {
...
+            var warning_text = "";
             var max = parseInt($('#upload-dialog').data('max-upload-size'));
+            var bad_file;
+            var selected = !!$(this).val();
 
-            var visible = (
-              this.files && this.files[0] &&
-              max && this.files[0].size > max
-            );
+            var spaces_in_filename = $(this).val().includes(" ");
```
prioux: Actually, why not compare it to the regex we use in the userfile model's validations ? That way we can complain about other illegal characters.

BrainPortal/public/javascripts/userfiles.js
```diff
@@ -722,25 +722,45 @@ $(function() {
...
+            if ( max > 0 ){
+              file_too_big = this.files && this.files[0] && max && this.files[0].size > max
+            } else {
+              file_too_big = false;
+            }
+
+            bad_file = ( spaces_in_filename || file_too_big );
+
+            if ( spaces_in_filename ) {
+              warning_text += "Spaces in filename not allowed! ";
```
prioux: Then this message would change to "Illegal characters in filename; filenames can only contain so and so".

BrainPortal/public/javascripts/userfiles.js
```diff
@@ -773,15 +793,6 @@ $(function() {
             $('#up-type').val(data).trigger('change.uf');
           });
         })
```
prioux: You sure that this wasn't needed ?

prioux:
Andrew, I'm closing the pull request, for some reason it now integrates all sort of changes from other unrelated pull requests.

Can you make a new one? Make sure you start from a freshly synchronized 'dev' in your work directory.

prioux:
I think it's normal. So just carry on with the comments...

BrainPortal/public/javascripts/userfiles.js
crocodoyle: It's still there, I just moved everything that modified the "disabled" status of the upload button into one method so they don't undo the work of the other ones

BrainPortal/public/javascripts/userfiles.js
crocodoyle: I have tried to use the userfiles model check for file validity, but I don't understand this RegEx:
`  def self.is_legal_filename?(basename)
    return true if basename && basename.match(/^[a-zA-Z0-9][\w\~\!\@\#\%\^\&\*\(\)\-\+\=\:\[\]\{\}\|\<\>\,\.\?]*$/)
    false
  end`

It appears to allow spaces?

natacha-beck:
@prioux and @crocodoyle I take a look to this PR and it's seems to have some extra stuff from others PR (everything related to test in bourreau side) like Pierre said. Can you take a second look.

crocodoyle:
It's because I pulled some things that were merged to dev and then committed them to this branch instead of rebasing.

prioux:
Can you try rebasing this to skip the first three commits by me?

prioux:
Or even better, rebase this to the current HEAD of the aces/dev branch .

crocodoyle:
I think it's good now

prioux:
All good Andrew!

PR #425, opened 2016-07-05
Remove SVN from revision tracking

crocodoyle:
This replaces all references to the SVN id string that was historically used to track revisions. There is a new method in cbrain_file_revision.rb called pretty(), which takes a hash and includes fields to the pretty string as needed, ex: {:author => true, :date => true}.  The default of pretty() is "short commit author date", as it was used almost everywhere in this form before.

I used ".basename" instead of the old ".file" because that's what was present in cbrain_file_revision.rb.

I tested this by running the tests and through the interfaces.


BrainPortal/app/models/bourreau.rb
```diff
@@ -247,7 +247,7 @@ def self.remote_resource_info
     info.merge!(
       # Bourreau info
       :bourreau_cms              => myself.cms_class || "Unconfigured",
-      :bourreau_cms_rev          => (myself.scir_session.revision_info.to_s rescue Object.revision_info.to_s),
+      :bourreau_cms_rev          => (myself.scir_session.revision_info rescue Object.revision_info),
```
prioux: This is wrong, the XML field cannot store a full object, the original code has a string representation of the revision info (the SVN ID string), we need something equivalent

BrainPortal/lib/cbrain_file_revision.rb
```diff
@@ -99,14 +96,32 @@ def self.[](fullpath)
     self.new(fullpath)
   end
 
-  # Returns the revision info in a fake 'svn' format:
-  #   "$Id: en_cbrain_local_data_provider.rb 12ab34 2010-03-25 17:26:05Z prioux $"
-  # Note that this triggers a self_update() call, if necessary, the first time.
   def to_s
```
prioux: Method is undocumented

BrainPortal/lib/cbrain_file_revision.rb
```diff
@@ -99,14 +96,32 @@ def self.[](fullpath)
...
-    return @fake_svn_id_string if ! @fake_svn_id_string.nil? # Cached; no need to fetch git info every time.
-    self_update()
-    @fake_svn_id_string = "$Id: #{@basename} #{@short_commit} #{@date} #{@time} #{@author} $"
-    @fake_svn_id_string
+    self.pretty({:commit => true, :author => true, :date => true})
+  end
+
+  def pretty(rev_fields = {})
```
prioux: Method undocumented

Also: you decode some possible subsets of keys in the arguments?!? What is if subset is not in the if/else cases? (e.g. :file and :time only) ?

BrainPortal/lib/scir.rb
```diff
@@ -51,12 +52,12 @@ def self.revision_info
...
   # Returns the file name for the implementation.
   def self.drmaa_implementation
-    Scir.revision_info.svn_id_file
+    Scir.revision_info.basename
   end
 
   # Returns the revision number for the implementation.
   def self.version
-    self.revision_info.svn_id_rev
+    self.revision_info.basename
```
prioux: Why basename, it used to be the SVN ID, so it should be a commit ID ?

prioux:
Missing files:

```
Bianca remove_svn_id CBRAIN > lag svn_id
./Bourreau/app/models/bourreau_worker.rb:    worker_log.info "#{@rr.class.to_s} code rev. #{@rr.revision_info.svn_id_rev} start rev. #{@rr.info.starttime_revision}"
./Bourreau/lib/bourreau_system_checks.rb:    rev = session.revision_info.svn_id_pretty_file_rev_author_date # loads it?
./BrainPortal/cbrain_plugins/cbrain-plugins-fmri-psom/cbrain_task/niak_pipeline_fmri_preprocess/bourreau/niak_pipeline_fmri_preprocess.xml.erb:  <!-- XML generated automatically by <%= self.class.to_s %> in <%= self.revision_info.svn_id_pretty_file_rev_author_date %> -->
./BrainPortal/lib/cbrain_file_revision.rb:#   puts Abcd::Revision_info.svn_id_rev # using CBRAIN's svn_id_rev parser
./BrainPortal/lib/cbrain_file_revision.rb:#   puts Abcd.revision_info.svn_id_rev  # using CBRAIN's revision_info method
./BrainPortal/lib/cbrain_file_revision.rb:#   puts x.revision_info.svn_id_rev
./BrainPortal/lib/cbrain_file_revision.rb:#   puts Abcd::Revision_info.author  # faster than svn_id_author()
```

crocodoyle:
@prioux I don't have the NIAK plugin and it's not in this git repository, so I can't fix it as part of this PR.

prioux:
In your first comment you say the default for pretty is a string of a single space ?!? Why is that?

BrainPortal/app/views/bourreaux/_runtime_info.html.erb
```diff
@@ -58,7 +58,7 @@
       <% t.cell("Cluster Management System Type")     { red_if(@bourreau.cms_class != @info.bourreau_cms, @info.bourreau_cms) } %>
 
       <% t.cell("Workers Last Change Revision")       { @info.worker_lc_rev } %>
-      <% t.cell("Cluster Management System Revision") { @info.bourreau_cms_rev.svn_id_pretty_rev_author_date } %>
+      <% t.cell("Cluster Management System Revision") { @info.bourreau_cms_rev.pretty()} %>
```
prioux: This crashes, as bourreau_cms_rev is not an object

crocodoyle: My first comment had some < and > that were hiding my text, fixed.

prioux:
The default behavior of .to_s() for a CbrainFileRevision object is missing many of the information that was in to original system:

BEFORE:

```
Scir.revision_info.to_s
 => "$Id: scir.rb 6082312d 2015-10-29 14:36:28 -0400 Pierre Rioux $" 
```

NOW:

```
Scir.revision_info.to_s
 => "4ee976a6 Andrew Doyle 2016-07-19" 
```

The filename is missing, the timestamp with timezone is missing, and the fields are not in the same order.

crocodoyle:
It's good now

prioux:
OK, this seems almost all fine, there are just a few minor nitpicks that could be made into one more commit. I'll comment in the code.

BrainPortal/app/views/bourreaux/_runtime_info.html.erb
```diff
@@ -58,7 +58,7 @@
       <% t.cell("Cluster Management System Type")     { red_if(@bourreau.cms_class != @info.bourreau_cms, @info.bourreau_cms) } %>
 
       <% t.cell("Workers Last Change Revision")       { @info.worker_lc_rev } %>
-      <% t.cell("Cluster Management System Revision") { @info.bourreau_cms_rev.svn_id_pretty_rev_author_date } %>
+      <% t.cell("Cluster Management System Revision") { @info.bourreau_cms_rev} %>
```
prioux: To make it more like the original behavior, I suggest you modify the code in bourreau.rb so that the value of bourreau_cms_rev has just the "commit_author_date" instead of what .to_s() creates right now. This is line 250 of bourreau.rb, which is currently unchanged (and I can't comment directly on it in the pull request).

BrainPortal/config/initializers/added_core_extensions/object.rb
```diff
@@ -26,6 +26,25 @@
 class Object #:nodoc:
 
   include CBRAINExtensions::ObjectExtensions::DeepClone
-  
+
+  # This method returns the value of the class constant
```
prioux: Put these two methods in a separate module file that will be included here, just like for DeepClone above.

BrainPortal/config/initializers/added_core_extensions/object.rb
```diff
@@ -26,6 +26,25 @@
...
+  end
+
+  # This method returns the value of the object's class constant
+  # named 'Revision_info', just like the class method of the
+  # same name.
+  def revision_info
+    self.class.revision_info
+  end
+
+
```
prioux: Spurious blank line. :-)

BrainPortal/lib/cbrain_file_revision.rb
```diff
@@ -55,11 +55,9 @@
...
 #
 #   # Then later on, all these calls are similar:
-#   puts Abcd::Revision_info.to_s # in SVN-like format
-#   puts Abcd::Revision_info.svn_id_rev # using CBRAIN's svn_id_rev parser
-#   puts Abcd.revision_info.svn_id_rev  # using CBRAIN's revision_info method
-#   x = Abcd.new
-#   puts x.revision_info.svn_id_rev
+#   puts Abcd::Revision_info.to_s # <commit> <author> <date> by default
+#   puts Abcd::Revision_info.pretty({:attribute => true}) # selects which of the revision's attributes to include in string
+#   puts Abcd.revision_info.to_s  # using CBRAIN's revision_info method
```
prioux: This comment block still refers to svn_id_author a few lines below.

BrainPortal/lib/cbrain_file_revision.rb
```diff
@@ -99,14 +94,37 @@ def self.[](fullpath)
     self.new(fullpath)
   end
 
-  # Returns the revision info in a fake 'svn' format:
-  #   "$Id: en_cbrain_local_data_provider.rb 12ab34 2010-03-25 17:26:05Z prioux $"
-  # Note that this triggers a self_update() call, if necessary, the first time.
+  # Returns a formatted string representing the last change to the file in git
```
prioux: This comment is not exactly true...

BrainPortal/lib/cbrain_file_revision.rb
```diff
@@ -99,14 +94,37 @@ def self.[](fullpath)
...
+  # Format: <commit> <author> <date>
   def to_s
-    return @fake_svn_id_string if ! @fake_svn_id_string.nil? # Cached; no need to fetch git info every time.
+    self.pretty({:file => true, :commit => true, :author => true, :date => true, :time => true})
+  end
+
+  # Returns a formatted string representing the last change to the file in git
+  # Format is determined by pasing attributes that you want as a hash.
+  # Defaults to: <commit> <author> <date>
+  def pretty(rev_fields = {})
```
prioux: This is the biggest change I suggest: transform pretty() to take a format string in argument:

```ruby
def pretty(format = "%s %a %d")
  etc
```

Supported format characters would be:

* %a = author
* %b = basename
* %c = commit
* %s = short_commit
* %d = date
* %t = time

You can then simply invoke a gsub() on the format string and return the result; gsub() can take a regex as first argument and a hash of subtitution in second args, making this entire new pretty() method a one-liner.

prioux:
One more note. Maybe the method pretty() could be called format(), or strfmt(), and maybe the format characters could be made compatible with those we see in "git log -h" ... just a thought. Maybe to discuss.

BrainPortal/lib/cbrain_file_revision.rb
```diff
@@ -99,14 +94,36 @@ def self.[](fullpath)
...
+  # %d = date
+  # %t = time
+  # Defaults to: <short_commit> <author> <date>
+  def format(rev_info = "%s %a %d")
     self_update()
-    @fake_svn_id_string = "$Id: #{@basename} #{@short_commit} #{@date} #{@time} #{@author} $"
-    @fake_svn_id_string
+
+    #rev_info.gsub("/%a/", @author).gsub("/%f/", @basename).gsub("/%c/", @commit).gsub("/%s/", @short_commit).gsub("/%d/", @date).gsub("/%t/", @time)
+    rev_info.gsub(/(%[afcsdt])/, {
```
prioux: This doesn't work... gsub() doesn't mutate its target, it returns a new string.

BrainPortal/lib/act_rec_log.rb
```diff
@@ -272,7 +272,7 @@ def addlog_context(context, message=nil, caller_level=0)
     class_name     = context.class.to_s
     class_name     = context.to_s if class_name == "Class"
     rev_info       = context.revision_info
-    pretty_info    = rev_info.svn_id_rev
+    pretty_info    = rev_info.short_commit
```
prioux: All these changes end up with an empty string being used instead of the short_commit (or other attributes) because the revision_info object is initially just a place holder with no data in it until `self_update()` is invoked. An example of that occurs when uploading a file and checking the file's log, we get:

```
[2016-08-08 13:13:25 EDT] create() UserfilesController rev.  Uploaded by admin
```

(see the two spaces after "rev." where the short commit is missing)

The current code base does not behave like the original (SVN compatible) one, where inquiring about any of the attribute values (commit, author, date etc) always fetched them from a self-updated object.

BrainPortal/lib/cbrain_extensions/object_extensions/revision_tracking.rb
```diff
@@ -0,0 +1,62 @@
...
+      module ClassMethods
+
+        # This method returns the value of the class constant
+        # named 'Revision_info', if it exists; otherwise it
+        # returns a default string in the same format
+        def revision_info
+          if self.const_defined?("Revision_info")
+           self.const_get("Revision_info")
+          else
+            CbrainFileRevision.new("") # dummy info object with dummy attributes
```
prioux: This produces an object which returns empty strings for all attributes, instead of the placeholder that were returned in the original code (e.g. `???` or `UnknownAuthor` etc).

prioux: Also I'm not sure the new() method should take a string argument anymore, it used to be the SVN $ID$ string but we got rid of that.

BrainPortal/lib/act_rec_log.rb
crocodoyle: When I upload a file, the author is displayed as "(author?)" in the file log. I don't know where this comes from though, because neither the Userfile model nor the CSVFile class write this line to the log
```
[2016-08-15 14:08:10 EDT] CSVFile revision 6082312d (author?) 2015-10-29
[2016-08-15 14:08:10 EDT] create() UserfilesController rev. 918a790c Uploaded by admin
```

BrainPortal/lib/cbrain_extensions/object_extensions/revision_tracking.rb
crocodoyle: Right now the new() method takes the ruby filename as a string. A CbrainFileRevision object with an empty string, when self_update() is called on it, gets populated with this:
`@basename="" @short_commit="91ffddd0" @date="2016-08-15" @time="14:06:11 -0400" @author="Andrew Doyle">`
I don't know where this comes from, it should give this:
```
    @commit   = "UnknownId"
    @date     = "1970-01-01"
    @time     = "00:00:00"
    @author   = "UnknownAuthor"
```

prioux: Why not add a class method

````ruby
  CbrainFileRevision.dummy_rev
```

(or `missing_rev`, or `default_rev` or some better name) ?

prioux: Also, yes, the `new()` method is supposed to be called with a full pathname, always, it's meant for that, and it shouldn't be otherwise.

prioux: See also my pull request of yesterday #461

PR #426, opened 2016-07-05
More boutiques features

ttaa9:
This adds some new Boutiques features as discussed in the issue boutiques/schema#19.

It includes: enum types, constraints on numerical parameters (min, max, exclusive min, exclusive max, integer), and constraints on lists (min and max length). Violations of numerical constraints are also detected and warnings are added to the portal task launch page as the user types.

Notes
- I had to update the schema or the validation step  on the json descriptors using enum types would fail (causing the whole cbrain app to fail)


glatard:
This all looks very good to me. I trust you that the tests still pass after these additions.

ttaa9:
Thanks. The pull request for the tests does not include the tests for these features though, because they are two separate branches (to avoid an enormous pull request like the one I had last time) so I didn't test them together. I felt we were more likely to merge this one before the tests were merged, and then I would add the new tests. I can add them now if you like though.

glatard:
It's ok like this, let's merge this and add the new tests later. @prioux, this can be merged as far as I'm concerned, but you may want to have a look too.

PR #439, opened 2016-07-26
Fix bug for 'Export as cvs' on firefox

natacha-beck:


prioux:
It's CSV, not CVS. :-)

PR #445, opened 2016-08-01
Created a simple, reusable color picker.

prioux:
Added it to the new and show pages for Access Profiles.

### DIFF
--- a/BrainPortal/app/views/access_profiles/new.html.erb
+++ b/BrainPortal/app/views/access_profiles/new.html.erb
@@ -35,13 +35,14 @@
   <div class="generalbox">
 
     <%= f.label :name %><br>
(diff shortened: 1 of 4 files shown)


crocodoyle:
VERY nice! All good for me

ttaa9:
I pulled and tested it, and it seems to work quite nicely! I guess you could change the css ids to classes (and assign an id dynamically) if you wanted the the modularity to have tons of color pickers on one page (a common use case in cbrain, of course). ;)

prioux:
I've discussed this with Natacha, yes it's not modular but for the moment it's Good Enough ™ . If we ever need to put more than one color picker in a page, we'll fix it.

Issue #440, opened 2016-07-27
Add ability to clean up caches using the file type


prioux:
The cache management interface (/bourreaux/rr_disk_usage) allows cleaning files from the caches by user, server, and dates, but not by file types.

Adding file types would allow us to clean output files of tasks without affecting input files (hopefully).

natacha-beck:
Implemented by PR #448

PR #448, opened 2016-08-03
Added ability to clean caches by file type.

prioux:
The cache cleaning panel now has a selection box
to filter out the report based on file types, and
when sending the clean command to a remote resource
the types selected will also be sent along.


prioux:
This implements #440 by the way.

PR #451, opened 2016-08-09
Fix: same content registered twice on some DPs

### LINKED ISSUE #180: Access rights on SSH data provider (Redmine 7971)
(From Tristan:)

 On an SSH data provider:

* User A registers file "test.txt" on data provider X.
* User B register file "test.txt" on data provider X (in a different project).
* User A's "test.txt" is overwritten by User B who has not rights on it (not even read).

prioux:
On data providers using a flat directory layout, it was
sometimes possible for two users to register files
with the same names, and then get their contents mixed up.
This is prevented by new callbacks, and also the upload
operation has been modified to handle this a bit more
gracefully.

Fixes #180


prioux:
Important note about this pull request: I'm going to implement the changes suggested in #452 and add it to this PR !

prioux:
Will fix #452 too.

PR #457, opened 2016-08-11
Removed unused support code for Mozilla Persona

prioux:
This alternative authentication mechanism was
no longer used in CBRAIN anyway.

I did my best to find all the relevant code and files, and only once I thought I was done did I compare to the original commit where the feature was added, back in February 2014.

For comparison, it was here: https://github.com/aces/cbrain/commit/b78767f201253930e97ce07834748d3b0dea0cb7


glatard:
Fine by me!

PR #384, opened 2016-06-01
Scir Amazon

glatard:
This is the first of a series of 3 pull requests coming from branch 'virtual_machines'.

It provides:
* A 'scir_amazon' class that can control VMs on Amazon using the Amazon Ruby SDK version 2. Scir Amazon can only execute 'StartVM' tasks. StartVM tasks contain the parameters required to start a VM on Amazon, except authentication params that are defined in the Bourreau. One may wonder whether starting VMs as CBRAIN task is a good idea. I chose to do that because it enables the use of all the CBRAIN task control features (monitoring, termination, etc) for VMs. 
* The 'StartVM' task, that starts a VM and monitors it until it has booted. All the instance types (i.e. memory, CPU characteristics) and images available to the account configured in the Bourreau can be used. 

Although there are tiny updates in scir.rb, all this code should have no side effect on a CBRAIN installation that does not have any Bourreau of type 'ScirAmazon'.

FYI, the next pull requests will contain:
* a mechanism to run tasks in VMs started with 'StartVM'.
* a mechanism to automatically start VMs based on the number of tasks to execute. 

Although the present PR does not contain any code related to these 2 points, having them in mind might explain some design choices.

PR #458, opened 2016-08-11
Tasks in VMs

glatard:
This is the second of the series of 3 pull requests initiated in #384. It contains a mechanism to execute tasks in VMs started with task StartVM. The main principle is the following:
* When a tool can run in a ScirCloud Bourreau, its ToolConfig contains the parameters of the cloud image where the tool is installed.
* When a task is submitted to a ScirCloud Bourreau, the Bourreau handles it differently depending whether it's a StartVM task or if it is a task that is supposed to run in a VM.
     * In the former case, the mechanism in #384 is used.
     * In the latter case, the task is assigned to a booted VM and run in it through ssh.

The next and final pull request on this topic will contain a mechanism to automatically start VMs based on the number of tasks to execute.

I'm setting this to High Priority (sorry for that...) as this is needed to run the experiments for a paper which is due for the end of August.


BrainPortal/app/models/cluster_task.rb
```diff
@@ -1566,10 +1566,33 @@ def submit_cluster_job
...
+        self.addlog("Queued as job ID '#{jobid}'.")
+      rescue => ex
+        # When the task is executed in a VM, it may not be submitted
+        # right away when no VMs are available. In such a case, method
+        # scir_cloud.run will raise an exception. We need to mask this
+        # exception and put the task back to status "New" so that
+        # submission will be attempted again by the Bourreau later
+        # on. VMs may become available for the task when other tasks
+        # complete or new VMs are started.
+        raise ex unless ex.message.include?("Cannot match task to VM.")
```
prioux: maybe check on the class of the exception instead of the message in the exception? 

```ruby
  if ex.is_a?(ScirCloudSubmitException)
  addlog()
end
```

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1566,10 +1566,33 @@ def submit_cluster_job
...
+        # This is unfortunate but we have to move back to status 'New'
+        # to give a chance to the task to be submitted again when
+        # there are VMs available. It shouldn't be too annoying
+        # though, since caches will avoid useless file transfers. To
+        # avoid this status transition, a new status could be
+        # introduced between "Setting Up" and "Queued"
+        # (e.g. "Scheduling"). That would require some more
+        # modifications in the bourreau worker logic though, so that
+        # tasks in status "Scheduling" are also picked up. 
+        self.status_transition("Setting Up","New")
```
prioux: Are you sure that it will be still in "Setting Up" ?

BrainPortal/app/models/task_vm_allocation.rb
```diff
@@ -0,0 +1,51 @@
...
+  
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  attr_accessible :task_id, :vm_id
+
+  # Automatically register the task's version when new() is invoked.
+  def initialize(arguments = {}) #:nodoc:
+    super(arguments)
+    baserev = Revision_info
+    self.addlog("#{baserev.svn_id_file} rev. #{baserev.svn_id_rev}", :caller_level => 3)
```
prioux: This is already done automatically for all objects the first time addlog() is called... no need to add the special code.

prioux: Also, the svn_id_* methods are all deprecated, will all be replaced in a pending pull request by Andrew.

BrainPortal/app/views/tool_configs/_form_fields.html.erb
```diff
@@ -115,5 +115,53 @@
     </p>
   </span>
 
+  <% bourreau = Bourreau.find(@tool_config.bourreau_id) %>
```
prioux: this will crash completely the view when a ToolConfig doesn't have a bourreau_id, which is legal (tool configs that apply to a tool for all bourreaux)

prioux: Also, it's better to use the Rails association methods than to do a find() explicitely:

```
   bourreau = @tool_config.bourreau
```

BrainPortal/app/models/cluster_task.rb
glatard: Fixed. We can even `rescue ScirCloudSubmitException => ex` to avoid using `is_a?`.

BrainPortal/app/models/cluster_task.rb
glatard: Nope, fixed, thanks.

BrainPortal/app/models/task_vm_allocation.rb
glatard: Fixed.

BrainPortal/app/views/tool_configs/_form_fields.html.erb
glatard: Yep, fixed.

BrainPortal/lib/scir_cloud.rb
```diff
@@ -52,8 +52,328 @@ def self.get_available_disk_images(bourreau)
   #     [ ["id_rsa_cbrain_portal"], ["personal_key"] ]
   # This (weird) data structure is used to pass the result of this method in a Rails select tag.
   def self.get_available_key_pairs(bourreau)
-    raise "Not implemented"
+    raise "Needs to be implemented in a sub-class"
+  end
+  
+  # Returns true of task is a VM
+  def self.is_vm_task?(task)
+    return task.is_a?(CbrainTask::StartVM)
```
prioux: I dislike a little that this link the Scir class to the CbrainTask class, but ok. One day we'll have to make sure all these layers are not interdependent.

prioux:
I hope this works... because I can't test it.

Issue #160, opened 2015-07-14
Boutiques integration


glatard:
1. Basic integration: multiply tasks only when tool has a single input file. When tool has more than 1 input file, then only 1 task is created. 
2. File and parameter lists for multiple inputs are specified in a CSV file. Drop-down menu allows to change the input associated to the list. A checkbox allows to specify that the first line in the CSV file is the input name.
3. File list selection in task interface (with filters, sort, etc). With only 1 input list. 
4. Support for multiple file input lists. Lists are combined one-by-one (when both lists have multiple elements), or all-to-one (when one list has only 1 element). Lists with multiple elements all have to have the same number of elements. 
5. Parameter multiplication with "all-to-all" (a.k.a cross) operator by default.

remibernard:
Additional notes to support multiple tool versions:
- A new wrapper class is used (with the same name) instead of the main `CbrainTask` tool subclass, which contains the Ruby `Class` instances generated for each version of the tool (in a class variable, for example).
- This wrapper class' instances wait until a `ToolConfig` is associated with it, then delegates all operations to the subclass matching the config's version.
- `remove_const` can be used to remove a class from its parent and returns the `Class` instance.

glatard:
Following recent discussions and experience, it is still not clear how we want to handle tools that have more than a single input file (item 2. to 5. above). The basic options are:
A. Add all the files selected by the user to all the input lists, and let the user choose more than 1 file in these lists. This is easy to do but it doesn't scale well to more than a few files (users won't click on 100 files). 
B. Let the user provide a CSV file where the first line of this CSV file contains input ids, and each row corresponds to a particular task to generate. This is powerful but it's a shortcut in the CBRAIN framework.
C. We don't support tools with more than a single input file (or directory). Tools that take more than a single input have to be modified to, e.g., find the relevant files in a single directory. 

This requires discussion...

ttaa9:
Given that I was just asked how to run a boutiques tool on 600 datasets, I think B seems to be the best. C might be complicated; e.g. for ica-aroma there are multiple input files, some with the same extension and some that could feasibly have multiple allowable types, that could conceivably be called anything. Also, it might be a huge annoyance for people who just want to run the tool once. And, like you said, A may not scale too well.

I am not sure how the CSV should be formatted though, especially if one wanted to run the same task with multiple, differently named inputs many times, in such a way that the format could be used across all boutiques tools. Maybe they should provide a json descriptor with the names of the inputs or something, with one entry per task.

[...]

PR #459, opened 2016-08-11
Adding multilaunch capabilities for Boutiques tasks

ttaa9:
This adds the ability to launch multiple jobs from a single task using Cbrain File list inputs to all Boutiques tasks.

It includes:
- Placing a cbcsv file in a file-type input causes n tasks to be launched, one for each row of the cbcsv, with each task being given a different input file for that parameter. If multiple cbcsvs are input, n tasks are still launched (i.e. the cbcsvs must be the same length) and the cbcsvs are iterated over simultaneously. Rows that have a nil or zero id simply give the parameter no value for that task.
- In the special case of a Boutiques tool with only one file-type input, all files are loaded into a that single parameter and a task is launched for each one, except if a given file is a cbcsv, in which case it is expanded to instead launch a task for every row of the cbcsv.
- Help documentation on usage has been added (which show up on both the tool page and the launch page itself).
-  A "Preview" button has been added to each task launch page. It builds a table that shows what and how many tasks will be launched based on the current parameter configuration, including the parameters that will be input to each .
- Rspec tests have been added for launching multiple tasks in both the general and single-file cases.

This relates to #160.


BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -829,11 +853,264 @@ $(function () {
...
-  parameters.delegate('.tsk-prm-rm, .ui-icon-minus', 'click', function () {
-    $(this)
-      .closest('li')
-      .remove();
+  /* Preview Button */
+  $("#previewer").click(function () {
+    // Classes for the modal (popup)
+    var mdiv = 'preview-modal', mcont = 'preview-modal-content', mclose = 'preview-modal-close';
+    // Input parameters (fetch ids and data from cbcsvs and json descriptor)
+    var ids    = [<%= params.map    { |p| "\"#{p['id']}\""    }.join(", ") %>];
```
prioux: Could any of these lists end up generating bad javascript? E.g. special characters in `p['name]'` or `c.name` ?

BrainPortal/public/stylesheets/boutiques.css
```diff
@@ -30,3 +33,114 @@
...
+
+/******************************************************************/
+/* Styling for modal preview popup for listing tasks in Boutiques */
+/******************************************************************/
+
+.preview-modal {
+  display: block;
+  position: fixed;
+  z-index: 1;
+  padding-top: 100px;
```
prioux: Just a general rule about CSS in CBRAIN, we try never to use the unit "pixel", because it doesn't scale well with high-density displays etc. It's better to try to express everything linear distances in, say, 'em's.

BrainPortal/spec/boutiques/boutiques_tester_spec.rb
```diff
@@ -47,9 +47,64 @@
...
+      expect( errMsgs.any? { |e| e.include? atLeastOneErrWith } ).to be true unless atLeastOneErrWith.nil?
+      expect( errMsgs.length ).to eq( checkVal )
+    }
+    # Function to manually add a null row to a cbcsv
+    @nilRowAdder = -> (text) {
+      fs, rt, n = CbrainFileList::FIELD_SEPARATOR, CbrainFileList::RECORD_TERMINATOR, CbrainFileList::ATTRIBUTES_LIST.length
+      splitTxt  = text.split( rt )
+      nilRow    = '0' + fs*(n - 1)
+      splitTxt  << nilRow
+      splitTxt.join( rt )
```
prioux: You need a record terminator after the last record too... `join(rt)` will not do that.

You could simply do

```ruby
splitTxt << ""
````

just before the line with `join()` though.

File: BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb (hunk 1 above)
Comment (ttaa9) [inline BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb]: I tested several cases, and adding an escaped double quote to the name field of the application descriptor does indeed break the JS here. I have added some escaping to fix that case. Most other issues are caught by the Boutiques json validation beforehand it seems.

File: BrainPortal/public/stylesheets/boutiques.css (hunk 2 above)
Comment (ttaa9) [inline BrainPortal/public/stylesheets/boutiques.css]: Ok, I've tried to change distances to em or % (so it's more like cbrain.css)

File: BrainPortal/spec/boutiques/boutiques_tester_spec.rb (hunk 3 above)
Comment (ttaa9) [inline BrainPortal/spec/boutiques/boutiques_tester_spec.rb]: Thanks, I did so

**Comment by prioux** [comment]
@glatard Do you want to review this further or do you feel it's ready to merge? I've made comments about things that Tristan has fixed, so I'm ok with the PR. You?

File: BrainPortal/lib/cbrain_task_generators/templates/edit_help.html.erb [hunk 4]
```diff
@@ -91,3 +91,37 @@
...
+  If multiple cbcsv files in different file inputs are given, they must be the same length, as tasks will be generated by iterating over the rows of <em>all</em> the input cbcsv files simultaneously.
+    Should one wish to launch all combinations of input file parameters among several lists, the cbcsv files must be constructed explicitly for this.
+  It is also possible to download, manipulate, and re-upload a cbcsv file (e.g. if one has local ordered lists of files for a specific task already).
+</p>
+<p>
+  Note that rows with IDs of 0 (zero) in the first column are treated as "null" inputs, and no value will be given to the task for that row under that parameter
+    (this is useful if one wishes to have no input given to a certain parameter for some of the launched tasks).
+</p>
+<%- end -%>
+
```
glatard: Very nice piece of doc!

glatard:
The code looks good to me. I have tested it briefly and it seems to be working well. Let's merge!

BrainPortal/lib/cbrain_task_generators/templates/edit_help.html.erb
ttaa9: Thank you

Issue #346, opened 2016-03-02
CBRAIN to report version number for its plugin packages


prioux:
Now that CBRAIN has a plugin mechanism, it shoudl be abel to report somewhere about the version number for these plugins.

Maybe in the credits -> more information -> platform information page.

PR #461, opened 2016-08-14
Support for revinfo in plugins.

prioux:
Plugins can now provide a flat file of revision info
in their top directory level, which will be parsed
just like for the main platform, when the app is not
deployed with git.

Script have been added/modified to CBRAIN's "script" subdirectory
to help (re)generate all the flatfiles.

The interface show the revsision info for the plugins in
the "about us" page.

Two unrelated files get better doc (in app/models).


prioux:
This will fix #346 by the way.

prioux:
Note: this PR will probably conflict with #425

PR #469, opened 2016-08-23
No longer trying to catch Exception.

prioux:
Trying to catch Exception was too strict; now trying to catch
StandardError like it was supposed to be.

Also a few enhancements.


crocodoyle:
Looks good to me. Can you elaborate a little bit on why Exception was "too strict" though?

prioux:
Yes Andrew; I was reading this last week:

http://blog.elpassion.com/Ruby-gotchas/

it mentions something I knew but it made me double-check our code. Normally when you want to catch a generic exception you should use RuntimeError, which is the superclass of all exceptions that a program will raise at run time that are not triggered by the system.

prioux:
By the way, `rescue => ex` basically means `rescue RuntimeError => ex`

prioux:
_(Corrected my previous two comments to says StandardError instead of RuntimeError)_

Issue #462, opened 2016-08-15
Make sure dumb sessions entries don't accumulate too much


prioux:
Robots and monitoring agents add one entry to the session table every time they connect.

Let's either make sure it doesn't happen, or that the entries get cleaned regularly.

We do have calls to CbrainSession.purge_sessions and CbrainSession.clean_sessions in the portal_controllers, but they're invoked only when a user logs in as an admin. And since I recently found 36K sessions entries in the DB, I'm thinking maybe it's not enough.

prioux:
This has been fixed by the whole rewrite of the session subsystem in R5.

PR #479, opened 2016-09-05
Many fixes to session maintenance.

prioux:
* CSRF authenticity token is not stored in session
  pages that do not require authentication
* removed some historical crud
* prep for API where session ID will be passed as params
* session no longer forced to be updated at each request


prioux:
This will fix #462 by the way.

Issue #474, opened 2016-09-02
Access Profile not applied after user creation


merous:
During the manual creation of a user, select one access profile and push create button.
The access profile is not applied to the newly created user. 
To do so, workaround is to go to the Access Profile in question and add the user in its member list.

prioux:
Yep, confirmed, the checkboxes are ignored, their variable names are wrong.

PR #481, opened 2016-09-06
Access Profiles are applied on user creation

crocodoyle:
Issue #474: The checkboxes had the wrong names before, and the access profiles were not applied.


prioux:
This doesn't work Andrew:

```
ActiveRecord::AssociationTypeMismatch in UsersController#create

AccessProfile(#70198353512420) expected, got String(#70198300649320)
```

prioux:
Apologies, code mixup on MY side.

PR #482, opened 2016-09-06
Audit of regexps: ^ become \A, $ become \Z

prioux:
Not all files needed ot be changed, and maybe some of these
changes won't make a difference.


crocodoyle:
I think we need \z instead of \Z, because unfortunately \Z matches a line ending at the end of the string, which is not what we want

BrainPortal/lib/username_format_validator.rb
```diff
@@ -0,0 +1,35 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+class UsernameFormatValidator < ActiveModel::EachValidator #:nodoc:
+
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  # Iterator that validates a username, UNIX style
+  def validate_each(object, attribute, value) #:nodoc:
```
crocodoyle: You have a nodoc here but there is in fact documentation. Maybe remove it?

crocodoyle:
There is a regex on line 744 of BarinPortal/public/javascripts/userfiles.js that still has ^ and $

prioux:
That's because it's javascript code, and javascript regex don't support \A and \z

crocodoyle:
This RegEx audit should extend to the plugins as well

prioux:
I already looked at the plugins, there's not much to change there.

PR #484, opened 2016-09-07
Batch tasks index checkboxes. Fixes #473.

### LINKED ISSUE #473: Tasks index page missing column when bourreau offline, in batch mode
In the tasks index page, batch mode, when a bourreau is offline and one try to expand a small batch of tasks, the rows that appear are all badly aligned because of a missing TD element where the checkbox would appear (the checkbox is properly not shown, but an empty TD shoudl be there instead). This happens for small batches (with a + sign). See screenshot for the batch of RPM tasks:

screen shot 2016-09-02 at 15 16 12

prioux:
When the bourreau was offline, the table was missing a column
for the checkboxes. Added new option to the dynamic table
framework's handling of rows.

### DIFF
--- a/BrainPortal/app/helpers/dynamic_table_helper.rb
+++ b/BrainPortal/app/helpers/dynamic_table_helper.rb
@@ -372,6 +372,12 @@ module DynamicTableHelper
     #  Defaults to 'selection'.
     #  Marks the row as selectable if specified.
     #
+    # [select_hidden]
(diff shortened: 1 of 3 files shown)


prioux:
@crocodoyle Have you reviewed this yet? I assigned it to you so you could figure out the way it was integrated with Remi's code.

crocodoyle:
yeah @prioux, I looked at it, thanks!

Issue #471, opened 2016-08-23
Get rid of CbrainTask attribute :launch_time


prioux:
This attribute was historically used to group together a bunch a tasks in a **batch**, but now there's a proper attribute called `:batch_id` that does that. Still, there are pieces of code that still sets `:launch_time` and gets its value. Everywhere this value is needed to be shown or used, we can simply use `:created_at`.

crocodoyle:
Can be closed

PR #485, opened 2016-09-07
Remove task launch time attribute

crocodoyle:
Task attribute launch_time used to be used to group tasks into batches, but isn't any more. This PR removes it, and in the one place where it was used, uses created_at instead.

Fix for #471


prioux:
All good. I'll merge this PR once we have merged the PRs for the plugins too...

BrainPortal/db/migrate/20160907191509_remove_task_launch_time.rb
```diff
@@ -0,0 +1,13 @@
+class RemoveTaskLaunchTime < ActiveRecord::Migration
+  def self.up
+    change_table :cbrain_tasks do |t|
+      t.remove :launch_time
```
prioux: please use the add_column and remove_column methods instead.

prioux:
All good.

PR #486, opened 2016-09-08
Adding environment variable support for boutiques

ttaa9:
Small update allowing task developers to specify environment variables in Boutiques descriptors.

Works by prepending `export` commands to the front of the cluster_commands array. Also added some tests for it. 

For Boutiques issue boutiques/boutiques#56, based on additions in pull request boutiques/boutiques#57.

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
@@ -54,6 +54,7 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] || 
 % out_keys = outputs.select { |o| o['command-line-key'] }
 % flags    = (inputs + outputs).select { |p| p['command-line-flag'] }
 % seps     = inputs.select { |i| i['command-line-flag-separator'] }
+% eVars    = descriptor['environment-variables'].dup rescue []
 %
   # Task properties are special boolean properties of your task, returned as a
   # hash table. Used internally by CBRAIN to enable/disable special task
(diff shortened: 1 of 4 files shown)


glatard:
Looks good to me! @prioux may want to have a final look before merging.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -125,6 +126,19 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
   # Note that this function also generates the list of output filenames
   # in params.
   def cluster_commands #:nodoc:
+
+    # Environment variables from Boutiques descriptor
+% if eVars.empty?
```
prioux: Why the `if`? You could have simply left the code in the `else` as is right?

```
  envVars = [
%    (eVars || []).each do |v|
%# The value is escaped for bash; the whole cmd is then escaped to be written into the ruby code
      <%= ("export %s=%s" % [v['name'], v['value'].bash_escape(always_quote=true)]).inspect %>,
%    end
    ]
```

BrainPortal/spec/boutiques/descriptor_test.json
```diff
@@ -4,6 +4,14 @@
 	"description" : "Simple test task for Boutiques",
 	"command-line" : "./boutiquesTestApp.rb [A] [B] [C] [a] [b] [c] [d] [e] [f] [g] [i] [j] [k] [l] [m] [n] [p] [q] [u] [v] [w] [x] [o] [r] [y] [E] [I] [N] [L]",
 	"schema-version" : "0.2",
+  "environment-variables" : [{
+    "name" : "ev1",
+    "value" : "nice_value"
+  }, {
+    "name" : "ev2",
+    "value" : "ta- 9\"\'_%^&$@][",
```
prioux: Lovely value for checking proper escaping! Cool!

prioux:
Aside from details it all looks good. Tristan merge this if you want, or Tristan change the code if you want too.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
ttaa9: Indeed. However, the reason for it is a tiny prettification change in the output code, because if the array is empty, I'd prefer to see
```
envVars = []
```
rather than
```
envVars = [
]
```

BrainPortal/spec/boutiques/descriptor_test.json
ttaa9: Thanks!

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
prioux: Lovely reason. All good. :-)

Issue #478, opened 2016-09-05
CSRF failures on login page can still allow user to log in, but CbrainSession is not properly set up


prioux:
Use a procedure almost identical to the one described in #477, change the CSRF token in the session just before logging in. Then examine the session object (which will be a different one in the DB!), and the two attributes `:user_id` and `:active` will be both blank, even though the session is actively being used (because the authentication mechanism use the `:user_id` field INSIDE the `data` attribute).

crocodoyle:
Can be closed

PR #489, opened 2016-09-12
Fixes in behaviour when CSRF token is wrong

### LINKED ISSUE #477: CSRF check failures can redirect to a POST
If a POST is performed with a bad CSRF, the session records the original page to `:return_to` and the user is returned to the login page. When authenticating successfully, the `:return_to` is sent as a GET redirect, which fails.

To test:

* Go to userfiles index page
* In the console, find the current session object under `CbrainSession`, and change in its `data` the value of the key `"_csrf_token"`.
* Do an action that triggers a POST (such as select a file to compress or uncompress
* You'll be redirected to the login page
* Enter login password
* You'll be redirected to a wrong URL (e.g. `/userfiles/uncompress`)

crocodoyle:
Prevents a user from logging in when they had a bad CSRF token.
When a session is invalid due to a bad CSRF token, a user logs back in again, and is returned to the page that triggered the logout. This can no longer be a POST.

Fixes #477 and #478.

### DIFF
--- a/BrainPortal/app/controllers/sessions_controller.rb
+++ b/BrainPortal/app/controllers/sessions_controller.rb
@@ -53,14 +53,8 @@ class SessionsController < ApplicationController
   def create #:nodoc:
     # If the csrf token is blank, it was reset by Rails' request forgery protection
     # and we want to prevent login
-    if current_session["_csrf_token"]
-      user = User.authenticate(params[:login], params[:password]) # can be nil if it fails
-      create_from_user(user)
-    else
-      flash[:error] = "Hacking CBRAIN won't work"
-      current_session["_csrf_token"] = session["_csrf_token"] #
-      auth_failed
-    end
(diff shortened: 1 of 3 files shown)


BrainPortal/app/controllers/sessions_controller.rb
```diff
@@ -51,8 +51,16 @@ def new #:nodoc:
   end
 
   def create #:nodoc:
-    user = User.authenticate(params[:login], params[:password]) # can be nil if it fails
-    create_from_user(user)
+    # If the csrf token is blank, it was reset by Rails' request forgery protection
+    # and we want to prevent login
+    if current_session["_csrf_token"]
```
prioux: That doesn't seem right, this test only verify that the token exists... ? Let me try it first...

crocodoyle:
This broke a few tests for SessionController.

PR #493, opened 2016-09-15
Params Errors API update

### LINKED ISSUE #450: Update the params_errors() API to match current Rails errors() API
We have a wrapper class `PortalTask::ParamsErrors` which is supposed to provide an API to record errors in a task's params using the exact same methods as for standard ActiveRecord attributes (which is done with the errors() method returning an ActiveModel::Errors object). However it seems many of the methods in ActiveModel::Errors have changed and so our wrapper is no longer in sync. For instance, methods such as `on()` and `on_base()` no longer exist. Maybe new ones exist too.

We need to update it our class so that its API become identical again.

crocodoyle:
Updating the ParamsErrors class to implement all the methods in the Rails 3.2.13 ActiveModel::Errors API (http://apidock.com/rails/v3.2.13/ActiveModel/Errors/), and removed deprecated methods.

The only method not implemented is generate_messages, which relies on looking up preset messages in the model, which is not implemented.

Fixes #450.


BrainPortal/app/models/portal_task.rb
```diff
@@ -545,52 +545,63 @@ def params_path_value(paramspath) #:nodoc:
...
       @real_errors.size
     end
 
-    def clear #:nodoc:
-      @real_errors.clear
+    def delete(paramspath) #:nodoc:
+      @real_errors.delete(paramspath.to_la_id.parameterize.underscore.to_sym)
     end
 
     def each(&block) #:nodoc:
```
prioux: The methods that modify the errors, or iterate over them etc should affect or access only those for the params, not those of the real attributes of the model... maybe? To discuss.

prioux:
I'm not sure what `generate_messages` does, but we do have `PortalTask.human_attribute_name` and `PortalTask.pretty_params_names` that are supposed to allow a task developer to provide nice names for the scientific params. Maybe that's what's needed?

BrainPortal/app/models/portal_task.rb
prioux: Ah never mind, we'll just let these methods work on ALL errors... but please document this behavior in the class' documention.

crocodoyle:
The `generate_messages` method looks up default error messages for each parameter that are stored in the model class. Do we want to add that?

prioux:
Oh I see its an I18N thingie. Let's just skip it for now.

BrainPortal/app/models/portal_task.rb
```diff
@@ -605,20 +616,49 @@ def empty? #:nodoc:
...
+      @real_errors.has_key?(paramspath.to_la_id.parameterize.underscore.to_sym)
+    end
+
+    def include?(paramspath) #:nodoc:
+      @real_errors.include?(paramspath.to_la_id.parameterize.underscore.to_sym)
+    end
+
+    def keys #:nodoc:
+      @real_errors.keys
+    }
```
crocodoyle: Yes. It returns the keys in the format :cbrain_task_params_xxx. Is that what we want?

prioux: I meant to say there's a spurious closing bracket at line 641 that makes the entire file not even load, because of the syntax error...

```
Bianca params_errors models > ruby -c portal_task.rb 
portal_task.rb:641: syntax error, unexpected tSTRING_DEND, expecting keyword_end
portal_task.rb:707: syntax error, unexpected tSTRING_DEND, expecting '}'
  def self.human_attribute_name(attname,options={})
                                                  ^
portal_task.rb:709: syntax error, unexpected tSTRING_DEND, expecting '}'
portal_task.rb:736: syntax error, unexpected tSTRING_DEND, expecting '}'
...ibutes(old_params, options = {}) #:nodoc:
...                               ^
portal_task.rb:737: syntax error, unexpected tSTRING_DEND, expecting '}'
portal_task.rb:739: warning: circular argument reference - options
portal_task.rb:739: syntax error, unexpected tSTRING_DEND, expecting '}'
portal_task.rb:818: syntax error, unexpected keyword_end, expecting end-of-input
```

BrainPortal/app/models/portal_task.rb
```diff
@@ -637,7 +645,9 @@ def include?(paramspath) #:nodoc:
     end
 
     def keys #:nodoc:
-      @real_errors.keys
+      display_keys = []
+      @real_errors.each {|err| display_keys << @base.class.human_attribute_name(err) }
```
prioux: This won't work I think, the each() method of the real_errors objects expects a block with two arguments; if you're using a single argument, it will be set to a pair.

BrainPortal/app/models/portal_task.rb
```diff
@@ -545,52 +545,71 @@ def params_path_value(paramspath) #:nodoc:
     foundvalue
   end
 
-  # Wrapper class around the ActiveRecord errors() method;
-  # this class answers to the same methods as the errors
+  # Wrapper around the ActiveModel::Errors class.
```
prioux: I think we're missing some methods still... check with

http://api.rubyonrails.org/classes/ActiveModel/Errors.html

For instance, `generate_message` is gone ?

BrainPortal/app/models/portal_task.rb
prioux: yes, let's make the `each` iterate ONLY on the scientific params hash keys.

BrainPortal/app/models/portal_task.rb
```diff
@@ -545,52 +545,71 @@ def params_path_value(paramspath) #:nodoc:
...
+  #
+  # See the Rails classes ActiveModel::Validations and
+  # ActiveModel::Errors for more information.
+  #
+  # Be careful when directly manipulating the errors using
+  # 'clear', 'delete', 'set', as they will modify the
+  # ActiveModel::Errors list of errors directly
+  #
+  # The 'generate_errors' method is not implemented.
+
```
prioux: The original class is an Enumerable, we should do it too.

BrainPortal/app/models/portal_task.rb
```diff
@@ -653,14 +653,18 @@ def get(paramspath) #:nodoc:
     end
 
     def has_key?(paramspath) #:nodoc:
-      @real_errors.has_key?(paramspath.to_la_id.parameterize.underscore.to_sym)
+      include?(paramspath)
```
prioux: For this method, and for `key?`, can't we just use an alias instead?

BrainPortal/app/models/portal_task.rb
```diff
@@ -545,92 +545,196 @@ def params_path_value(paramspath) #:nodoc:
...
-    def clear #:nodoc:
-      @real_errors.clear
+    def delete(paramspath) #:nodoc:
+      @real_errors.delete(paramspath.to_la_id.parameterize.underscore.to_sym)
     end
 
+    # NOTE: returns a copy of the real errors hash
     def each(&block) #:nodoc:
-      @real_errors.each(&block)
+      only_params_errors.each(&block)
```
prioux: The behavior of the `.each()` method is wrong: each yield should be given |att,msg| where msg is not a an array, but a single message, even when multiple error messages are associated with the same atttibute.

BrainPortal/app/models/portal_task.rb
```diff
@@ -545,92 +545,196 @@ def params_path_value(paramspath) #:nodoc:
...
+
+    def to_a #:nodoc:
+      full_messages
     end
 
+    def to_hash(full_messages = false) #:nodoc:
+      return_hash = {}
+      if full_messages
+        only_params_errors.each { |err, msgs|
+          return_hash[err] = []
```
prioux: This line will be a bug when `each` is fixed.

PR #494, opened 2016-09-15
Buttons on Files and Tasks show pages

### LINKED ISSUE #460: Add buttons for some actions in the userfiles and task 'show' pages.
The `show` pages for userfiles and tasks could benefit from having a few more options to trigger actions on the currently viewed file or task.

E.g.
- userfiles: a 'compress' or 'uncompress' (or 'archive' / 'unarchive') button
- tasks: an 'unarchive' button, a 'terminate' button...

There are others that don't come to mind right now, but simply compare with the capabilities of the `index` pages.

crocodoyle:
Fixes #460.


BrainPortal/app/views/tasks/show.html.erb
```diff
@@ -57,6 +57,26 @@
       <% confirm += "\n\nIMPORTANT NOTE: This will also remove its archived work directory." if @task.workdir_archive %>
       <%= link_to 'Remove Task',  { :action => 'operation', :operation => 'delete', 'tasklist[]' => @task.id }, :confirm => confirm, :class  => "button menu_button", :method => :post %>
     <% end %>
+    <% if @task.status =~ /Failed*|Completed/ %>
```
prioux: Don't use state names explicitely; use the contants defined in cbrain_task.rb ( `COMPLETED_STATUS`, `FAILED_STATUS` etc). Yes, other buttons were already making such comparisons, let's clean them up too.

BrainPortal/app/views/tasks/show.html.erb
```diff
@@ -57,6 +57,26 @@
...
+                                                  :method  => :post,
+                                                  :datatype => :script,
+                                                  :class => "button",
+                                                  :confirm =>
+                  "Are you sure you want to remove the task's work directory?"
+      %>
+    <% end %>
+
+    <!-- ARCHIVING -->
+    <% if @task.status =~ /Completed/ && @task.full_cluster_workdir.blank? == false && @task.archived_status.nil? %>
```
prioux: Don't compare with `thing.blank? == false`, use `thing.present?`, it's much more legible.

BrainPortal/app/views/tasks/show.html.erb
```diff
@@ -57,6 +57,26 @@
...
+      <%= link_to "Archive on Cluster", url_for(:action  => :operation, :operation => 'archive', 'tasklist[]' => @task.id), :method  => :post, :datatype => :script, :class => "button" %>
+      <% if @task.results_data_provider_id %>
```
prioux: The `results_data_provider_id` has nothing to do with task archiving on file; the task's archive can be on a completely different data provider than the tasks' results. It is wrong to impose archiving to happen on the same data provider as the result, and in fact it can corrupt a lot of things (especially for DPs with dedicated data types)

BrainPortal/app/views/userfiles/show.html.erb
```diff
@@ -31,6 +31,14 @@
   <% if @userfile.can_be_accessed_by?(current_user, requested_access = :read) %>
     <%= link_to(@userfile.is_a?(FileCollection) ? "Download Collection" : "Download File", { :action => 'download', :file_ids => [@userfile.id] }, :class => "button menu_button", :method  => :get) %>
   <% end %>
+
+  <% if @userfile.can_be_accessed_by?(current_user, requested_access = :write) && @userfile.is_a?(SingleFile) %>
```
prioux: Make the compress/uncompress button work for both SingleFile and FileCollection (resp: compression vs archiving). This is what the file manager interface already does.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
```diff
@@ -36,7 +36,12 @@ def self.valid_file_classes #:nodoc:
   end
 
   def pretty_type #:nodoc:
-    self.class.pretty_type + (name =~ /(\.gz|\.Z|\.gz2)$/i ? " (compressed)" : "")
+    self.class.pretty_type + (compressed? ? " (compressed)" : "")
+  end
+
+  def compressed? #:nodoc:
```
prioux: This introduces a problem; if a file is .Z or .gz2, the interface will show a 'uncompress' button, but the controller code only supports `.gz` compression/uncompression.

BrainPortal/app/views/tasks/show.html.erb
natacha-beck: Does this:

> Don't use state names explicitely; use the contants defined in cbrain_task.rb ( COMPLETED_STATUS, FAILED_STATUS etc). Yes, other buttons were already making such comparisons, let's clean them up too.
 
Apply here too

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
natacha-beck: @prioux and @crocodoyle  Should we handle .Z and .gz2 file then ? Is it possible or not and do we want that.

crocodoyle: I say we force people to `.gz`

BrainPortal/app/views/tasks/show.html.erb
prioux: Andrew, this hasn't been addressed. I was saying you cannot have a "Archive As File" button that defaults to the results_data_provider_id for the file's destination. Either we have no button for this action, or you provide a selector for the data provider.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
prioux: Please only show the "uncompress" button when the controller has the capability to uncompress the file.

prioux:
General note: the order the buttons are showing is arbitrary and confusing. Here's a screenshot:

screen shot 2016-11-16 at 12 45 12

I think we should move all dangerous operations further right, so that they are not clicked on when casually reaching for the tabs. Also, maybe they shoudl be regrouped or separated by vertical bars, so that all the "restart" buttons are together, all the "remove" togther etc.

The "Edit Parameters" maybe should be first. Try something!

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
prioux: I don't have the "Archive" button for a standard file collection...

prioux: I don't have the "Unarchive" button for an archived file collection... which was even an explicit request of #460

crocodoyle: Is it Archive/Unarchive or Compress/Uncompress?

BrainPortal/app/models/userfile.rb
```diff
@@ -186,7 +186,11 @@ def pretty_type
     self.class.pretty_type
   end
 
-
+  # Returns whether or not the userfile is compressed using the gzip format
+  def compressed?
+    gz_extension_pos = (name =~ /(\.gz)$/i)
+    !gz_extension_pos.blank?  # if no .gz extension found, file not zipped
```
prioux: Huh.. ok, but `gz_extension_pos.present?` avoids a double negation...

prioux:
I get a spurious vertical bar at the end:

screen shot 2016-11-16 at 16 20 29

Also, I'm really not sure about these bars, they look dam ugly... isn't there a nicer separator we can find?

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
prioux: I think it's Compress/Uncompress at all times now... I never liked that but I've been overruled.

BrainPortal/public/stylesheets/cbrain.css
```diff
@@ -1025,6 +1025,12 @@ pre {
   padding: 0;
 }
 
+.vertical_bar {
+  display: inline;
+  border-left: solid black;
+  margin-right: 5px;
```
prioux: Never use pixels... they are not screen-resolution independent.

BrainPortal/app/models/userfile.rb
crocodoyle: I was trying to change the minimum amount of code to accomplish the goal.

prioux:
I'm ready to merge this, except for the conflict; you'll have to resolve this Andrew.

PR #495, opened 2016-09-16
First pass at ajusting credits pages.

prioux:
Also a few changes here and there.


prioux:
I'm going to merge this after having reviewed the appearance of the page with all of the CBRAIN team; I really am not sure about it. I'll push further mods to this branch as I get feedback.

PR #502, opened 2016-09-28
Change Creator of a Project

### LINKED ISSUE #501: Better support for creator attribute of Groups
Groups (aka projects) have a creator_id attribute that points to a user.
- add this as a column of the index page of groups (table format)
- allow owner of the group (i.e. the person in the creator_id attribute) to reassign it to someone else (with a big warning that the user can't be change it back!)

crocodoyle:
From the details page of a project, the Creator field is now editable, and includes users that the current user has access to, and all members of the project. Fixes #501

### DIFF
--- a/BrainPortal/app/controllers/groups_controller.rb
+++ b/BrainPortal/app/controllers/groups_controller.rb
@@ -130,6 +130,7 @@ class GroupsController < ApplicationController
     end
 
     original_user_ids = @group.user_ids
+    original_creator = @group.creator_id
 
     params[:group] ||= {}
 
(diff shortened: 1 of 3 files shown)


BrainPortal/app/controllers/groups_controller.rb
```diff
@@ -151,8 +151,11 @@ def update #:nodoc:
       end
     end
 
+    unless (current_user.available_users.map{ |u| u.id } | @group.user_ids).include?(params[:group][:creator_id].to_i )
```
prioux: Add logging of creator_id change in controller (this can be done simply by providing the attribute name to `update_attributes_with_logging` below

crocodoyle: Adding it to the whitelist produces:
`[2016-11-02 15:46:00 UTC] update() Updated by admin: creator_id : value(1 -> 9)`
Using `addlog_object_list_updated` is prettier so I used that:
```
[2016-11-02 15:44:19 UTC] Creator updated by admin: Removed: cabingley
[2016-11-02 15:44:19 UTC] Creator updated by admin: Added: admin```

Comment (prioux) [inline BrainPortal/app/controllers/groups_controller.rb]: OK, it's a workaround that will do for the moment; the logging helper doesn't know how to link the attribute `creator_id` to the `User` model (even though Rails know about it) because it does a simple parsing of the attribute name and tries to find a class `Creator` instead. The fix would be to make it understand some Rails associations in `BrainPortal/lib/act_rec_log.rb` at the line that says

```ruby
model = Regexp.last_match[1].classify.constantize rescue nil
```

Issue #505, opened 2016-09-30
Boutiques: spacing in task launch page


glatard:
When inputs have no comments, spacing is reduced. See spacing above "Corsica" group and input "Regress confounds thre fd" in the screenshot below.
![screenshot from 2016-09-30 15-24-36](https://cloud.githubusercontent.com/assets/5174953/19004105/30a7abf4-8722-11e6-8483-db57bbdbafc0.png)

Issue #504, opened 2016-09-30
Handling of input files in Boutiques


glatard:
Input files of Boutiques tools are handled in a weird way: while task parameters initially contain file ids, which is correct, file ids are replaced by a serialized Ruby object after the task moved from status `Setting Up` to `Queued`. This is done in method `resolve_file_parameter` of the CbrainTask (current line 449 of `lib/cbrain_task_generators/templates/bourreau.rb.erb`). This behavior is actually intentional, since the Ruby object is used later on, in method `apply_template`, where `v.name` is used which would obviously fail on a file id (current line 491 of `lib/cbrain_task_generators/templates/bourreau.rb.erb`).

This breaks command-line generation *when the task is resubmitted after its parameters were edited*. I guess the clean way to solve it would be to reconstruct the Ruby object from the file id systematically in `apply_template`.

PR #506, opened 2016-10-05
Fixing Boutiques issues

ttaa9:
This is for Boutiques-related issues #504 and #505.

The former comprises the majority of the changes, and stops the changing of file ids into userfile objects. Many of the changes are updates to the test suite so that it works with the change.


ttaa9:
Yes, they do

glatard:
Perfect. Then it's fine for merging as far as I'm concerned. Maybe @prioux wants to have a final look though.

PR #520, opened 2016-10-17
ParamsErrors handle deep params struct. Fixes #518

### LINKED ISSUE #518: ParamsErrors don't highlight input fields for deep params
If an error in a CbrainTask's params is recorded, where the param is deep inside the hash, as in the *paramspath* `myhash[mykey]` or deeper:

```ruby
task.params_errors["myhash[mykey]"] = "is bad"
```

then in the task's param's page the field will not by highlighted in red. The reason is that the ID of the input field will be provided by the `.to_la_id()` method and will be shown as

```html
<input id="cbrain_task_params_myhash_mykey">
```

whereas in the params_errors object they key is stored as the symbol `:"cbrain_task_params_myhash[mykey]"` (yes a symbol with brackets in its name).

This was convenient because it allowed the entire API of ParamsErrors to always provide the developer a consistent interface using *paramspaths*; however now Rails fail to match the ID and the fact that an error was recorded.

Not sure how to solve that yet; maybe stuff the errors object with a duplicate key `:cbrain_task_params_myhash_mykey` key ?

prioux:
These are modifications to the convention for storing errors related to scientific parameters in a task, improving on PR #493 . This fixes two problems:

* there is no longer an ambiguity between a parameter "xyz_abc" at the top level of the hash, and one that is a bit deeper such as "xyz[abc]"
* the automatic colorizing in red of params with errors is done properly no matter where the params path is (top level, or deeper).

The new key comvention, as in  (and associated getter method name) seems a bit clumsy, but as a programmer one should never need to use them directly.

Old convention prior to PR #493 : (use a wide window!)

```
# Paramspath    to_la_id()                                              Internal key in ParamsErrors
#-------------- ------------------------------------------------------- ----------------------------
:xyz            "cbrain_task_params_xyz"                                (same as to_la_id).to_sym
"abc_def"       "cbrain_task_params_abc_def"                            (same as to_la_id).to_sym
"abc[def]"      "cbrain_task_params_abc_def"                            (same as to_la_id).to_sym
```

Convention used in PR 493 : 

```
# Paramspath    to_la_id()                                              Internal key in ParamsErrors
#-------------- ------------------------------------------------------- ----------------------------
:xyz            "cbrain_task_params_xyz"                                "cbrain_task_params_xyz".to_sym
"abc_def"       "cbrain_task_params_abc_def"                            "cbrain_task_params_abc_def".to_sym
"abc[def]"      "cbrain_task_params_abc_def"                            "cbrain_task_params_abc[def]".to_sym
```

Current convention:

```
# Paramspath    to_la_id()                                              Internal key in ParamsErrors
#-------------- ------------------------------------------------------- ----------------------------
:xyz            "cbrain_task_BRA_params_KET__BRA_xyz_KET_"              (same as to_la_id).to_sym
"abc_def"       "cbrain_task_BRA_params_KET__BRA_abc_def_KET_"          (same as to_la_id).to_sym
"abc[def]"      "cbrain_task_BRA_params_KET__BRA_abc_KET__BRA_def_KET_" (same as to_la_id).to_sym
```


crocodoyle:
Nice solution, I would change the following though: `_BRAC_` and `_KET_` instead of `_BRA_` and `_KET_` so that it spells BRACKET

prioux:
I'm using the German word _BRAKET_ which means the dorsal side of an octopus' arm. Duh.

Issue #133, opened 2015-06-08
Capture ourselves the STDOUT and STDERR of ClusterTasks


prioux:
#### Main feature

On the clusters, instead of showing the STDOUT and STDERR as captured by the qsub/msub commands, which often are not sent to the files in real time, we should introduce a new layer of wrapping that captures them to file ourselves. That way, when looking at a task that is currently "On CPU", we would be able to see the outputs as they are being produced. The bash script submitted to the cluster would be simple and  look something like:

```bash
#!/bin/bash

bash original_bash_script 1>outcapture 2>errcapture
```

#### Special supplemental feature

Since often the cluster's management system also produces useful prologue and epilogue associated with the submitted wrapper, we could integrate merge the full thing with a special keyword placeholder:

```bash
#!/bin/bash

echo _SPECIAL_PLACEHOLDER_FOR_STDOUT_SUBSITUTION_
echo _SPECIAL_PLACEHOLDER_FOR_STDERR_SUBSITUTION_ 1>&2
bash original_bash_script 1>outcapture 2>errcapture
```

In that case, after the job is fully finished, we can rewrite the content of outcapture and errcapture to include the prologue and epilogue outputs.

prioux:
I really need to think about this carefully.

prioux:
Related to issue #64 : both can be done at the same time.

glatard:
I'm tempted to argue for the exit code check again here. If the application commands are wrapped in a separate script, then I really think it would make sense that the exit code of this script can be retrieved, and the task would be failed on cluster when the exit code is not 0 *and a specific task property, e.g. :check_exit_code_value, has been set by the task developer*. When this property is activated, it is then up to the task developer to write a script that will properly set the exit code when the task fails. This would prevent task developers from writing their own buggy exit code checks in an ad-hoc manner.

glatard:
This feature would be very useful for the PSOM integration as PSOM pipelines report on their progression during execution (using #364).

glatard:
This would be really nice to have for long processes like PSOM pipelines that write progress information in their stdout during execution.

poquirion:
A feature I have seen in a some cluster manager I have used il the past is the possibility to provide a path to be feed tou a `tail` like function and/our a path plus a regex to feed a number to a progress bar. In any case a feature to seen what is going with the computation is more than necessary for the cbrain psom integration. I have a test that have been running for two days, and I have no feed back except for the fact that  it in `On CPU`.

[...]

PR #524, opened 2016-10-21
Tasks are started with a wrapper bash script.

### LINKED ISSUE #64: Create a second level of script wrapping for cluster jobs
In order to improve robustness on clusters that do not
provide reliable information about active jobs, Tarek
and I discussed and bit and came up with an improvement.

The user-supplied bash script is no longer to be submitted
as thejob, instead a wrapper bash script is submitted, which
would look like this:

```
got_sigterm() {
  touch terminated; exit 20
}

trap got_sigterm SIGTERM

touch started
bash the_job_script
touch finished
```

This way we can more or less detect cases where the cluster's scheduler
is lying to CBRAIN, and properly wait some more if we think the
jobs is invisibly on CPU, or if it has never been executed at all
(two common problems).

prioux:
This allows us to view live the output of the scientific
script.

The rest of the API has been kept perfectly compatible.
One added behavior is that the framework, at task post processing,
will check for a particular sentence echoed back at the end of the wrapper
script to make sure the task ran properly to the end, and
otherwise will mark the task as Failed On Cluster immediately.

This will fix #64 and #133 .

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -70,10 +70,22 @@ class ClusterTask < CbrainTask
 
   include NumericalSubdirTree
 
-  # These basenames might get modified with suffixes appended to them.
-  QSUB_SCRIPT_BASENAME = ".qsub"      # appended: ".{name}.{run_id}.sh"
-  QSUB_STDOUT_BASENAME = ".qsub.out"  # appended: ".{name}.{run_id}"
-  QSUB_STDERR_BASENAME = ".qsub.err"  # appended: ".{name}.{run_id}"
+  # The 'science' script is where the scientific commands
+  # of the task are held. It is a bash script, where most
+  # of the earlier commands are created from the ToolConfigs,
+  # and the last commands are supplied by the cluster_commands()
+  # method.
+  #
+  # These basenames get modified with suffixes appended to them.
+  SCIENCE_SCRIPT_BASENAME = ".science"     # appended: ".{name}.{run_id}.sh"
+  SCIENCE_STDOUT_BASENAME = ".science.out" # appended: ".{name}.{run_id}"
+  SCIENCE_STDERR_BASENAME = ".science.err" # appended: ".{name}.{run_id}"
+
+  # The qsub script is a bash wrapper that executes the scientific script.
+  # These basenames get modified with suffixes appended to them.
+  QSUB_SCRIPT_BASENAME    = ".qsub"        # appended: ".{name}.{run_id}.sh"
+  QSUB_STDOUT_BASENAME    = ".qsub.out"    # appended: ".{name}.{run_id}"
+  QSUB_STDERR_BASENAME    = ".qsub.err"    # appended: ".{name}.{run_id}"
 
   before_destroy :before_destroy_terminate_and_rm_workdir
   validate       :task_is_proper_subclass
@@ -81,7 +93,7 @@ class ClusterTask < CbrainTask
 
 
   ##################################################################
-  # Core Object Methods
+  # Core Model Methods
   ##################################################################
 
   # Automatically register the task's version when new() is invoked.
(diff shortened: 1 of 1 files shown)


BrainPortal/app/models/cluster_task.rb
```diff
@@ -70,18 +70,30 @@ class ClusterTask < CbrainTask
...
+  # The 'science' script is where the scientific commands
+  # of the task are held. It is a bash script, where most
+  # of the earlier commands are created from the ToolConfigs,
+  # and the last commands are supplied by the cluster_commands()
+  # method.
+  #
+  # These basenames get modified with suffixes appended to them.
+  SCIENCE_SCRIPT_BASENAME = ".science"     # appended: ".{name}.{run_id}.sh"
+  SCIENCE_STDOUT_BASENAME = ".science.out" # appended: ".{name}.{run_id}"
+  SCIENCE_STDERR_BASENAME = ".science.err" # appended: ".{name}.{run_id}"
```
glatard: Nitpick: I would say "application" or "tool" rather than "science".

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1007,48 +1023,131 @@ def prerequisites_fulfilled?(for_state)
...
+    return science_outerr if ! File.exists?(qsub_outerr)
+    qsub_content = check_task_ending_keyword(qsub_outerr) # returns nil unless content contains "CBRAIN Task Exiting"
+    return science_outerr if qsub_content.blank?
+
+    # Create combined report
+    science_content = File.read(science_outerr) # can be big but generally not to much
+    qsub_content.sub!('__CBRAIN_CAPTURE_PLACEHOLDER__',science_content)
+    File.open("#{combined_file}-tmp#{Process.pid}","w") { |fh| fh.write(qsub_content) }
+    File.rename("#{combined_file}-tmp#{Process.pid}", combined_file) # atomicity for the win
+    combined_file
```
glatard: So the user will actually see the "science" report while the task is running, and suddenly it will change to the complete report when the task is completed. How about we keep these two reports always separate and show them in different tabs in the interface? In most cases, the qsub report won't be useful for the user, only admins may look at it. It would also avoid doing the merging trick.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1007,48 +1023,131 @@ def prerequisites_fulfilled?(for_state)
...
+    cb_error "Current directory is not the task's work directory?" unless self.we_are_in_workdir
+
+    # Report already prepared? Just return its basename
+    combined_file = "#{qsub_outerr}-combined"
+    return combined_file  if File.exists?(combined_file)
+
+    # Task outputs incomplete? Point to science output captured
+    return nil            if ! File.exists?(science_outerr)
+    return science_outerr if ! File.exists?(qsub_outerr)
+    qsub_content = check_task_ending_keyword(qsub_outerr) # returns nil unless content contains "CBRAIN Task Exiting"
```
glatard: Why not rely on the task status instead of parsing the stdout file?

BrainPortal/app/models/cluster_task.rb
prioux: Ah, anyway users don't see any of this nomenclature.

BrainPortal/app/models/cluster_task.rb
prioux: Because this check here is intended to detect tasks that have been completely killed with "kill -9" or even a system reboot. The token here is not generated by the scientific script, but by the new qsub wrapper. The scientific task can die/exit in any number of ways, but this token here WILL end up in the captured output as along as the cluster works properly.

Also, there has never been a requirement for return status of the scientific script provided by the programmer.

BrainPortal/app/models/cluster_task.rb
prioux: Yes, I considered doing that, but the difference is minimal: about 1 line at he beginning, and 5-6 lines at the end. And I think seeing the the content of the qsub output in the interface can be useful for the admin, but adding separate tabs will be darn confusing for the users...

It could be done, but I'd have to add two new fields to the Controls object that shuffles information between the Bourreau and the Portal, and that's a layer I don't like to touch.

PR #526, opened 2016-10-24
Collection downloads

### LINKED ISSUE #519: Downloading a file as a 1-step process (Lindsay)
Lindsay said in https://github.com/aces/cbrain-plugins-neuro/issues/73

```
It would be nice to be able to download a file from a collection directly, as a one-step process, rather
than having to click on "Extract Files from Collection", wait to be taken back to the list of files, then
have to scan the entire list of files for the one you just extracted, and then "Download."
```

crocodoyle:
Refactors the `data_link` method to add download links to objects within file collections, only for real files and not for symlinks. Fixes #519.

### DIFF
--- a/BrainPortal/app/helpers/userfiles_helper.rb
+++ b/BrainPortal/app/helpers/userfiles_helper.rb
@@ -80,17 +80,14 @@ module UserfilesHelper
   # of type TextFile or ImageFile
   # Generates download link for any other type of file
   def data_link(file_name, userfile)
-    dp = DataProvider.find(userfile.data_provider_id)
-    cache_path = dp.cache_full_path(userfile)
-
-    full_path_name  = Pathname.new(cache_path.dirname + file_name)
-
-    file_lstat = full_path_name.lstat  # lstat doesn't follow symlinks, so we can tell if it is one
+    full_path_name  = Pathname.new(userfile.cache_full_path.dirname + file_name)
 
     display_name  = full_path_name.basename.to_s
     return h(display_name) unless userfile.is_locally_synced?
+
+    file_lstat = full_path_name.lstat  # lstat doesn't follow symlinks, so we can tell if it is one
+
     return h(display_name) unless file_lstat.file?
-    return h(display_name) if file_lstat.symlink?
 
     matched_class = SingleFile.descendants.unshift(SingleFile).find { |c| file_name =~ c.file_name_pattern }
 
--- a/BrainPortal/app/helpers/userfiles_helper.rb
+++ b/BrainPortal/app/helpers/userfiles_helper.rb
@@ -77,10 +77,20 @@ module UserfilesHelper
   end
 
   # Generates links to pretty file content for userfiles
-  # of type TextFile or ImageFile, downloads any other type of content
+  # of type TextFile or ImageFile
+  # Generates download link for any other type of file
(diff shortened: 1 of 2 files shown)


BrainPortal/app/helpers/userfiles_helper.rb
```diff
@@ -77,16 +77,24 @@ def file_link_table(previous_userfile, next_userfile, sort_index, options = {})
   end
 
   # Generates links to pretty file content for userfiles
-  # of type TextFile or ImageFile; this method is going to
-  # be replaced by a proper generic framework in 4.3.0 !
+  # of type TextFile or ImageFile
+  # Generates download link for any other type of file
   def data_link(file_name, userfile)
-    display_name  = Pathname.new(file_name).basename.to_s
+    dp = DataProvider.find(userfile.data_provider_id)
```
prioux: It's unnecessary to find the DP and call the cache_full_path method on it, the userfile model already has shortcuts:

```ruby
full_path = userfile.cache_full_path.dirname + file_name
```

(because cache_full_path already returns a Pathname object)

BrainPortal/app/helpers/userfiles_helper.rb
```diff
@@ -77,16 +77,24 @@ def file_link_table(previous_userfile, next_userfile, sort_index, options = {})
...
+    cache_path = dp.cache_full_path(userfile)
+
+    full_path_name  = Pathname.new(cache_path.dirname + file_name)
+
+    file_lstat = full_path_name.lstat  # lstat doesn't follow symlinks, so we can tell if it is one
+
+    display_name  = full_path_name.basename.to_s
     return h(display_name) unless userfile.is_locally_synced?
+    return h(display_name) unless file_lstat.file?
```
prioux: Isn't the symlink? test redundant oif we already tested .file? before?

prioux: If possible, let's try to have the method return the default values as early as possible withotu accessing the file system if it can avoid it

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/file_collection/views/_plain_file_list_row.html.erb
```diff
@@ -41,7 +41,7 @@
     <% else %>
       <%= image_tag "/images/file_icon.png" %>
     <% end %>
-    <% if file.size > 0%>
+    <% if file.size > 0 && file.symbolic_type != :directory %>
```
prioux: Instead of comparing as `not a directory`, compare to `is a plain file`, to be safer

crocodoyle: The `data_link` already checks if it's a `.file?` on an `lstat` structure, which doesn't allow the following of symlinks, so this directory check is not even useful.

prioux: I see, but that cause another problem. The return is done too early. If we want to add data links for things that are not files, e.g. subdirectories that match a FileCollection model, we can't.

PR #528, opened 2016-10-24
Log Filters

### LINKED ISSUE #515: Adjust log viewer filtering rules
The 'view log' feature has a few checkboxes to filter out log lines, such as "Load", "CACHE", "Rendered" etc. These use to match particular log line patterns, but the format of the Rails log has changed slightly since then and they are not up to date. SQL lines are now looking like:

```
  (0.2ms)  SELECT COUNT(*) FROM `users` 
```

and I don't see any of the 'AREL' lines anymore.

Update the list of patterns and filtering code.

crocodoyle:
Changed the log filters so they will filter out the relevant log lines again. The SQL filter may pick up more than just the SQL statements, as it matches `SELECT`, but I don't think this is a problem as if you're trying to remove SQL statements you probably don't care about cache SQL statements. Fixes #515.

### DIFF
--- a/BrainPortal/app/controllers/portal_controller.rb
+++ b/BrainPortal/app/controllers/portal_controller.rb
@@ -95,10 +95,13 @@ class PortalController < ApplicationController
     remove_egrep << "^ *Redirected"   if params[:hide_redirected].presence == "1"
     remove_egrep << "^User:"          if params[:hide_user].presence       == "1"
     remove_egrep << "^Completed"      if params[:hide_completed].presence  == "1"
-    # Note that in production, 'SQL', 'CACHE', 'AREL' and 'LOAD' are never shown.
+    # Note that in production, 'SQL', 'CACHE' and 'LOAD' are never shown.
     remove_egrep << "^ *SQL "         if params[:hide_sql].presence        == "1"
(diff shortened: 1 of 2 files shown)


BrainPortal/app/controllers/portal_controller.rb
```diff
@@ -95,10 +95,13 @@ def portal_log #:nodoc:
     remove_egrep << "^ *Redirected"   if params[:hide_redirected].presence == "1"
     remove_egrep << "^User:"          if params[:hide_user].presence       == "1"
     remove_egrep << "^Completed"      if params[:hide_completed].presence  == "1"
-    # Note that in production, 'SQL', 'CACHE', 'AREL' and 'LOAD' are never shown.
+    # Note that in production, 'SQL', 'CACHE' and 'LOAD' are never shown.
     remove_egrep << "^ *SQL "         if params[:hide_sql].presence        == "1"
+    remove_egrep << "BEGIN"           if params[:hide_sql].presence        == "1"
```
prioux: Your regex are not anchored, we'll be removing WAY too many lines

prioux:
All good! Thanks!

PR #541, opened 2016-11-02
Less verbose 'rake cbrain:plugins:...'

prioux:
I made the rake tasks less verbose: basically they only show what they DO, not was they skip because it's done. There's a `verbose` toggle at the top which re-enables all the origin message but it's not yet selectable using arguments on the command-line.

### DIFF
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -786,7 +786,7 @@ class TasksController < ApplicationController
           next
         end
 
-        if task.user_id != current_user.id && current_user.type != 'AdminUser'
+        if task.user_id != current_user.id && ! current_user.has_role?(:admin_user)
           (skipped_list["you not allowed to #{operation} this task(s)"] ||= []) << task
           next
         end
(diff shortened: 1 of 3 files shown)


crocodoyle:
I would rather see what was skipped, it makes it easier to figure out what the rake task is actually doing and whether or not it did what you thought it was supposed to do. I think it should be verbose by default and have a command line argument to turn off verbosity.

crocodoyle:
We should also take this opportunity to standardize the rake task calls for Bourreau and BrainPortal. Currently it is:
`rake cbrain:plugins:install:all` on the BrainPortal side
`rake cbrain:plugins:install:plugins` on the Bourreau side

This causes confusion even amongst our most experienced developers.

natacha-beck:
@prioux Do you want to implement the verbose flag suggested by @crocodoyle before I merge ? 

And what's about the second question, the 2 tasks are different for me they are no way to standardize this one.

prioux:
Do not merge yet; yes I have ideas, just too sick to talk at length.

prioux:
The verbose flag was already there; by default I want it NON verbose because in the past, whenever we updated a plugins or installed a new one, running the rake task would report about everything and we would have difficulty making sure that we had properly installed something new. So I prefer a task that only reports what it has just done. There is no point in saying "this symlink is OK" all over.

As for the distinction between `rake cbrain:plugins:install:all` and `rake cbrain:plugins:install:plugins` , I've fixed this too with my most recent commit: you can invoked `install:all` no matter where you are and things will work properly now. I had to remove a rake dependency on the environment, but it had been introduced only because of Boutique stuff and that part shoudl evenutually be made into separate rake tasks.

PR #546, opened 2016-11-08
SshAgent fix

prioux:
Fixes a bug where the output of `ssh-add -l` has changed slightly, so the class would no longer detect that a SshAgent is properly running.


prioux:
Wrong base branch.

PR #547, opened 2016-11-08
Fixed SshAgent handling for OpenSSH 6.8+

prioux:
Fixes a bug where the output of `ssh-add -l` has changed slightly, so the class would no longer detect that a SshAgent is properly running.


crocodoyle:
Looks good to me, I like that the default return is `false` now as well.

Issue #525, opened 2016-10-21
Support latest boutiques schema


glatard:
- Should be at least 0.3, maybe 0.4 if the new release is there.
- Make sure that all the cbrain plugins have been migrated to the latest schema.
- In particular, support the new `file-template` property for configuration files. It might be easier to include `localExec.py` in CBRAIN as a way to generate Boutiques command lines.

natacha-beck:
We already support 0.5. So I close it.

PR #550, opened 2016-11-14
Boutiques filepaths

### LINKED ISSUE #527: Boutiques tools with inputs that don't have the 'optional' propery defined crash.
If the optional property is not present, it should be assumed that the input is not optional, as it is meant in JSON.

The tool currently crashes with the error message attached. 
screenshot from 2016-10-24 14-08-42

ttaa9:
- Adds support for some new Boutiques features, including path types (absolute vs relative) and the more flexible container syntax (see boutiques/boutiques#58, boutiques/boutiques#51)
- Updates the schema to a newer version (not the latest)
- Adds support for "working-directories" in containers. This includes detecting when output files are not underneath the mounted directory's subtree, and moving them before exiting the container (so that cbrain can detect and save them) via appending `mv` commands to the end of `cluster_commands`
- Adds tests for the above
- Fixes #527; related to #525.
- Note: should be merged in conjunction with aces/cbrain-plugins-neuro#83


BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -267,10 +298,42 @@ class CbrainTask::<%= name %> < <%= descriptor['cbrain:inherits-from-class'] ||
...
+        # Relative paths are assumed to be with respect to the mounted directory
+        currentPath = File.absolute_path( params[ outid ], container_wd )
+        # If the current path is not in the directory subtree of the mounted directory,
+        # issue a move command at the end of cluster_commands and alter the path to search
+        # for output files accordingly
+        unless is_underneath.( container_wd, currentPath )
+          # Move the file to the mounted dir (to be run post-execution, in the container)
+          # Warning: potential overwrite dangers if there is a same-named local file
+          newPosition = File.join( container_wd, File.basename(currentPath) )
+          outfileMoveCommands << "mv #{currentPath} #{newPosition}"
```
prioux: Will these `mv` command properly escape path with special characters in them? Filenames with spaces, hash signs, at signs, dollar signs, backticks, UTF8-encoded characters?

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1904,8 +1916,9 @@ def docker_present?
   # Returns the command line(s) associated with the task, wrapped in
   # a Docker call if a Docker image has to be used.
   def docker_commands
+    workDir  = self.container_working_directory || '${PWD}'
```
prioux: I'm not asking you to change it, but in general avoid using variables with `namesUsingInternalCapitals`, it's not the Ruby way.

BrainPortal/spec/boutiques/test_helpers.rb
```diff
@@ -148,7 +148,7 @@ def destroyOutputFiles
     ["works with optional string", baseArgs2 + "-a s -w", 0],
     ["works with optional flag", baseArgs2 + "-c -w", 0],
     ["works with optional string list", baseArgs2 + "-e s1 s2 -w", 0],
-    ["works with optional file list", baseArgs2 + "-f #{f_files} -w", 0],
+    ["works with optional file list", baseArgs2 + "-f #{F_files} -w", 0],
```
prioux: Was this a bug? Just curious.

ttaa9: No, I just thought it was cleaner as a constant

BrainPortal/app/models/cluster_task.rb
ttaa9: Sorry, too used to worse languages :(

Issue #548, opened 2016-11-10
Travis CI improvements.


prioux:
I know this feature is not yet merged, but here are some thoughts about improvements to it.

- [x] rename the environment variable that Travis use to identify the docker image; instead of `IMAGE_NAME` maybe use something like `CBRAIN_TRAVIS_IMAGE_NAME`
- [x] seems the bash `SECONDS` counter doesn't increment as expected in travis.sh
- [x] maybe perform the _dockerize_ steps for the template in the docker image instead of at run time?
- [x] document the `CBRAIN_FAILTEST` feature
- [x] use the `--depth` option with the git clone command inside the container? Value, 20 commits?
- [x] investigate if it would be possible to build the mysqldb inside the container and keep it in seeded mode right there
- [ ] document the distinction between the subdirectories `cbrain_base`, `cbrain_travis` and `cbrain_test` in the cbrain user HOME directory
- [x] customizing the Travis build from within .travis.yml (see https://docs.travis-ci.com/user/customizing-the-build/)
- [ ] make the docker image smaller by uninstalling things that are no longer needed at the end (compilers? libs? dev files?)

prioux:
More improvements:

* run the tests on the Bourreau side
* run tests for plugins! for this we'll need some new conventions:
  - have a`spec` directory in the root of a plugins package
  - have the `.travis.yml` there too
  - need new entry point in the base CBRAIN for running those tests
  - the cbrain_travis docker mountpoint will no longer exists, we need maybe to point one level up and make the script find out what plugin is being tested

prioux:
This is all moot. Travis is no longer used.

PR #551, opened 2016-11-17
Added support for Bourreau-side tests

prioux:
Hi. I added support for running the tests on the Bourreau side now in Travis.

For the moment I hardcoded that only `spec/boutiques` is supplied in argument to `rspec` because other files we have in the `spec/` subdirectory are all broken; only the boutiques files work.

The test script will run rspec on the portal side, and then on the bourreau side EVEN if there were failures on the portal side.

Everything is reported nice and clean at the end. I didn't even have to rebuild the Docker image for CBRAIN, it was usable as-is.

Eventually I'd like to have a lot more tests on the bourreau side and then we'll run them all.

### DIFF
--- a/BrainPortal/db/seeds_test_bourreau.rb
+++ b/BrainPortal/db/seeds_test_bourreau.rb
@@ -82,10 +82,7 @@ end
 # Seeding steps starts here
 #------------------------------------------------
 
-raise "The seeding process must be run by a process connected to a terminal" unless
-  STDIN.tty? && STDOUT.tty? && STDERR.tty?
-stty_save = `stty -g`.chomp
-trap('INT') { system('stty', stty_save) ; puts "\n\nInterrupt. Exiting."; exit(0) }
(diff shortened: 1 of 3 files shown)


prioux:
Also, you'll like this Natacha, the docker logs of `travis_cbrain_1` are now shown in real time in Travis.

prioux:
@ttaa9 Question for you, can you tell me why when I run your test files with rspec, I get the output format with a series of periods (`.......` etc) instead of the default one?

prioux:
This will take care of some point in #548

natacha-beck:
Everything seems ok for me, I am really happy with the CI of CBRAIN. We definitely need to fix the bourreau side test. And add some more tests every time we add new feature.

ttaa9:
@prioux I'm not sure. I believe the default rspec options are simply different for the Bourreau side for some reason. I get the same "condensed" output on my own machine as well.

PR #554, opened 2016-11-27
 Travis CI improved: DB inside container.

prioux:
This PR changes completely the manner that Travis CI is integrated within CBRAIN.

The docker container that we build separately now contains a fully seeded DB, and when running the tests we do the minimum amount of work necessary to bring the internal CBRAIN install up to par:

* We only bundle when Gemfiles have changed;
* we only migrate when the migration files have changed;
* we only install the plugins when their files have changed.

We also no longer need docker-compose, as we run a single container,
and I've removed the dependencies on the external utility "dockerize".


prioux:
Please note that on dockerhub I already uploaded the image `mcin/cbrain_travis:dev`

PR #556, opened 2016-12-01
Boutiques update

ttaa9:
Updates the Boutiques implementation in Cbrain to the latest Boutiques version. Includes:
- Replacing the old schema with the newest version
- Adding support for minor new Boutiques features (changes to enum types, output file command-line-flag-separators, substitution bug fixes)
- Fixing the Boutiques tests to work in random order

Note: the schema update is not backwards compatible. Hence, to work, the latest descriptors must be used in the plugins. In this case, this means the changes in aces/cbrain-plugins-neuro#86.

Related to: #525, boutiques/boutiques#70, boutiques/boutiques#59


ttaa9:
Note that, for the failing test (in 76ee00e), running the tests locally with the same random seed did not cause a failure. However, it seems it was caused by not saving a data provider before saving a userfile assigned to it. I'm not sure why it worked locally, but failed on Travis.

prioux:
I'd be OK to merge this, except for the fact that the descriptors must also be updated. It causes a problem of updating our dev server. Will think some more.

glatard:
Looks good to me. Let's update and release the descriptors before this gets merged though.

ttaa9:
If possible, could @prioux take a look at this before merging? @glatard and I suggest this be included in the next release (assuming it will be updated in conjunction with the boutiques json descriptors)

prioux:
OK I'll have a look closer (I typically look at all pull requests, but often the boutique stuff is just a superficial check; I'll do a more thorough check)

PR #557, opened 2016-12-02
Fixes some tests.

prioux:
The failures were related to allow() and expect() statements
that were made before the FactoryGirl construction of some
objects were performed, and these interfered with those constructions.
Assigning the constructed objects to variables BEFORE specifying
the test conditions solved that. It's something to remember, in
general.

Also, failures are limited to 50 on each Portal and Bourreau sides.

### DIFF
--- a/Bourreau/.rspec
+++ b/Bourreau/.rspec
@@ -3,3 +3,4 @@
 --require spec_helper
 --format documentation
 --order defined
+--fail-fast=50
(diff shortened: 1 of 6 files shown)


crocodoyle:
These failures only appear in Travis though, and not when running rpsec locally?

prioux:
Actually, they do appear when you run locally. That's how I was able to investigate and fix them. To trigger them the tests have to be run in a specific order, and so I had to find the seed that made them fail and prune the test suite until I could reliable trigger them. In the end I could make them occur by specifying an explicit line number, e.g something like

```bash
rspec --fail-fast  spec/models/en_cbrain_local_data_provider_spec.rb:55
```

PR #558, opened 2016-12-05
api_attr_visible inherits from superclasses.

prioux:
The set of whitelisted attributes is now always the union of those specified int he current class, and those specified in any of the superclasses. Cached internally. No mechanism yet to remove an attribute from the white list, but I don't think we need one right now.


crocodoyle:
Looks good to me

PR #566, opened 2016-12-09
Swagger API

crocodoyle:
This is a first pass at the complete CBRAIN API. I manually tested each of the functions in "Try it now" mode, except for some of the Userfiles posts.

There is at least one issue: for the POSTs, only one 'body' type parameter is allowed, so many of the post parameters are described within objects, and instead of nice text boxes with hints about what you should put in them, it requires the user to build a big JSON file. This could possibly be resolved by breaking these POST  parameters into 'formData' or 'query' type parameters, but I haven't tried it.


prioux:
Suggestion: since we often use

```ruby
   array_of_object.map(&:for_api)
```

why don't we define a new array method `.for_api` to make this cleaner?

BrainPortal/app/controllers/groups_controller.rb
```diff
@@ -106,17 +118,20 @@ def create  #:nodoc:
         @group.addlog_context(self,"Created by #{current_user.login}")
         flash[:notice] = 'Project was successfully created.'
         format.html { redirect_to :action => :index, :format => :html}
-        format.xml  { render :xml => @group, :status => :created, :location => @group }
+        format.xml  { render :xml  => @group.for_api, :status => :created }
+        format.json { render :json => @group.for_api, :status => :created }
       else
         @users = current_user.available_users.where( "users.login<>'admin'" ).order( :login )
```
prioux: This line only necessary for HTML... also the check condition `login<>'admin'` is old stuff that really shoudl be rewritten using the proper methods for identifying the core admin user.

BrainPortal/app/controllers/groups_controller.rb
```diff
@@ -194,20 +213,23 @@ def unregister
...
       end
     end
   end
 
   # DELETE /groups/1
   # DELETE /groups/1.xml
+  # DELETE /groups/1.json
   def destroy  #:nodoc:
-    @group = current_user.available_groups.find(params[:id])
+    @group = current_user.available_groups.where(params[:id]).first
```
prioux: Leave the `find`, which will raise an exception if the geroup isn't found and will trigger the appropriate code in the application controller!

BrainPortal/app/controllers/portal_controller.rb
```diff
@@ -369,7 +369,7 @@ def search
   def swagger
     # Find latest JSON swagger spec.
     # FIXME sort() will break when comparing versions...
-    @specfile   = Dir.entries(Rails.root + "public" + "swagger").grep(/\Acbrain-.*-swagger.json\z/).sort.last
+    @specfile   = Dir.entries(Rails.root + "public" + "swagger").grep(/\Acbrain-.*-swagger.yaml\z/).sort.last
```
prioux: Don't we want to provide the JSON version of the spec to developers?

BrainPortal/app/controllers/tags_controller.rb
```diff
@@ -24,66 +24,78 @@ class TagsController < ApplicationController
...
     end
   end
 
   # GET /tags/1
   # GET /tags/1.xml
+  # GET /tags/1.json
   def show #:nodoc:
     respond_to do |format|
-      format.xml { render :xml => current_user.tags.find(params[:id]) }
+      format.xml  { render :xml  => current_user.tags.find(params[:id]).for_api }
```
prioux: extract the find outside the respond block

BrainPortal/app/controllers/tools_controller.rb
```diff
@@ -120,6 +123,18 @@ def create #:nodoc:
     end
   end
 
+  # GET /tools/1
+  # GET /tools/1.xml
+  # GET /tools/1.json
+  def show
```
prioux: What happens to HTML requests here?

BrainPortal/app/controllers/users_controller.rb
```diff
@@ -75,8 +75,21 @@ def index #:nodoc:
 
   # GET /user/1
   # GET /user/1.xml
+  # GET /user/1.json
   def show #:nodoc:
-    @user = User.find(params[:id], :include => :groups)
+    begin
+      @user = User.find(params[:id], :include => :groups)
```
prioux: Let the application controller do all that!

BrainPortal/app/models/bourreau.rb
```diff
@@ -31,6 +31,8 @@ class Bourreau < RemoteResource
 
   attr_accessor :operation_messages # no need to store in DB
 
+  api_attr_visible :id, :name, :user_id, :group_id, :online, :read_only, :description
```
prioux: the attribute `:id` is ALAWAYS visible, it's configured as such automatically by api_attr_visible, so there' sno need to specify it every time.

BrainPortal/app/models/group.rb
```diff
@@ -63,6 +63,8 @@ class Group < ActiveRecord::Base
 
   attr_accessible         :name, :description, :site_id, :creator_id, :user_ids
 
+  api_attr_visible        :id, :name, :description, :type, :site_id, :invisible
```
prioux: Do not show the ;invisible attr (which will eventually be replaced by something better)

BrainPortal/db/schema.rb
```diff
@@ -41,6 +41,11 @@
 
   add_index "active_record_logs", ["ar_id", "ar_table_name"], :name => "index_active_record_logs_on_ar_id_and_ar_table_name"
 
+  create_table "bourreaux_tools", :id => false, :force => true do |t|
```
prioux: That table is spurious! Where does it even come from?!? There's no many-to many assoc between bourreaux and tools!

BrainPortal/db/schema.rb
```diff
@@ -117,6 +122,35 @@
   add_index "data_providers", ["type"], :name => "index_data_providers_on_type"
   add_index "data_providers", ["user_id"], :name => "index_data_providers_on_user_id"
 
+  create_table "demands", :force => true do |t|
```
prioux: That table is also spurious! Check you dev DB, it seems you've left it in an inconsistent state, so the schema here is reconstructed wrong.

BrainPortal/db/schema.rb
```diff
@@ -175,6 +209,17 @@
 
   add_index "help_documents", ["key"], :name => "index_help_documents_on_key", :unique => true
 
+  create_table "logged_exceptions", :force => true do |t|
```
prioux: spurious

BrainPortal/db/schema.rb
```diff
@@ -318,6 +363,15 @@
     t.datetime "updated_at", :null => false
   end
 
+  create_table "statistics", :force => true do |t|
```
prioux: spurious

BrainPortal/db/schema.rb
```diff
@@ -404,6 +458,17 @@
   add_index "tools", ["group_id"], :name => "index_tools_on_group_id"
   add_index "tools", ["user_id"], :name => "index_tools_on_user_id"
 
+  create_table "user_preferences", :force => true do |t|
```
prioux: spurious

BrainPortal/lib/cbrain_extensions/active_record_extensions/api_attr_visible.rb
```diff
@@ -62,7 +62,7 @@ def cumulative_api_attr_visible #:nodoc:
           return @cumulative_api_attr_visible if @cumulative_api_attr_visible
           local_visible_list = api_attr_visible.keys
           super_visible_list = superclass.respond_to?(:cumulative_api_attr_visible) ? superclass.cumulative_api_attr_visible : []
-          @cumulative_api_attr_visible = (local_visible_list | super_visible_list).map(&:to_s) # need strings :-(
+          @cumulative_api_attr_visible = (local_visible_list | super_visible_list).map(&:to_s)
```
prioux: My comment disappear but that's because we made the same fix separately

prioux:
Changes have been applied in my pull request on your fork: https://github.com/crocodoyle/cbrain/pull/6

PR #578, opened 2016-12-21
Fixed bug with container working directory

glatard:
Only Boutiques tasks have a `container_working_directory` method. Without this fix, non-Boutiques Docker tasks crash (try adding a Docker images to a Diagnostics ToolConfig). Should be a hotfix in my opinion.


prioux:
Thanks, well, if the only way to make it crash is when having a spurious image name in the tool config, I'll let it go until next release (I mean, not push it to master).

prioux:
Oh wait... non-boutique DOCKER tasks crash too? OK, maybe it shoudl go to master then.

prioux:
By the way, normally I'd supply a symbol to `respond_to?`

glatard:
ONLY non-Boutiques task crash.

PR #581, opened 2017-01-03
Swagger API Update

crocodoyle:
This pull request adds to the API Swagger UI page:
-breaks apart `params` parameters blocks into individually editable fields, with associated parameter descriptions
-sets many default values for parameters so they do not all appear as `undefined` by default
-specifies many parameters as required

Version updated to 4.7 because that is likely the next CBRIAN release.


prioux:
Eventually we will distribute within CBRAIN several .yaml and .json files for different versions, when they do differ.

PR #582, opened 2017-01-03
BourreauWorkers checks task output before postproc.

### LINKED ISSUE #577: Task ending keyword detection is too strict
Method `check_task_ending_keyword` in ClusterTask seems a bit too strict. In particular, when stdout doesn't exist, it may not be because of a failure of the task and we may want to try to upload the results anyway.

For instance, task[ #371245](https://portal.cbrain.mcgill.ca/tasks/371245) was set to "Failed on Cluster" with the following error message: "Could not find completion keywords at end of output. Process might have been interrupted.". However, the scientific task completed successfully and its stdout does contain the string "CBRAIN Task Exiting". 

This is really annoying since it puts successful tasks in a failed status. I suggest that the following line of `check_task_ending_keyword`:
`return nil unless File.exists?(basename)`
is replaced by:
`return "unknown" unless File.exists?(basename)`
The calling method could just print a warning when "unknown" is returned.

Thoughts?

prioux:
The ClusterTask class now has a method ready_to_post_process?
which is invoked by the BourreauWorker before trying to trigger
post processing.

This will fix #577 

There are a few misc other changes I bundled in the commit too.

### DIFF
--- a/Bourreau/app/models/bourreau_worker.rb
+++ b/Bourreau/app/models/bourreau_worker.rb
@@ -337,6 +337,9 @@ class BourreauWorker < Worker
 
       #####################################################################
       when 'Data Ready'
+        # If the status Data Ready is very recent and we don't find the stdout file from qsub, we wait.
+        return unless task.ready_to_post_process?
+
         action = task.prerequisites_fulfilled?(:for_post_processing)
         if action == :go
           # We need to raise an exception if we cannot successfully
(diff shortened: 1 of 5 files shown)


crocodoyle:
No exponential backoff 👎 
Otherwise looks great

prioux:
We can't have an exponential backoff, because the situation where the error file never shows up is possible too, and it's a genuine error condition that we want to detect and flag (as `Failed On Cluster`). If we waited forever we'd have users complaining that their tasks stays in 'Data Ready' forever.

crocodoyle:
We could wait 2^n seconds for n up to 8

crocodoyle:
OK whatever

Issue #584, opened 2017-01-05
Signups index page reports conflicting email when account has been created


prioux:
The 'conflicting email' message in the Signups page is supposed to highlight an email address that is a duplicate:

1- either within another Signup record or
2- to another user in CBRAIN

Right now when a user has been created properly, the entry in the Signup reports as conflicting because the account is there, obviously it should only be marked as conflicting if it's for a different account.

Issue #583, opened 2017-01-05
No pagination in Signups index page


prioux:
Please add pagination ASAP. We can't access the latest entries.

prioux:
Also, maybe the default ordering should be in reverse order of created_at

Issue #563, opened 2016-12-08
Signups improvements


prioux:
* Add a private comment field for the admins in the Signup model.
* Do not populate the full name of the User object with the person's title.

Issue #555, opened 2016-11-27
AccessProfiles are not applied at user creation time


prioux:
The IDs of the profiles are set, but they groups list is not adjusted.

It is fortunate that as soon as the user connects and modifies any aspect of his account (including resetting the password), the access profiles groups will be applied as a side effect. But it should really be done at account creation time.

crocodoyle:
Note to self: this, and other Access Profiles issues, can be solved by calling `apply_access_profiles` with the :before_save callback in the User model.

prioux:
It would be best to wrap the method inside a separate one that is registered as the callback:

```ruby
before_save :adjust_ap

def adjust_ap  # find better name
  apply_access_profiles
  true
end

That's because apply_access_profile was not originally designed to be a callback, so no guarantee about its return values etc.

PR #585, opened 2017-01-05
Signups Improvements & AccessProfiles bugfixes

crocodoyle:
Changes to Signups:
-First + Last name is used as Users' full name now instead of including their title and middle name
-Private administrator comment field for Signups
-Pagination and default sort order by creation date (descending)
-Added a "Created" column to table on index page

Changes to Access Profiles:
-Applying AccessProfiles as a callback on User save
-Fixed a bug where Users would be added to a group twice if two Access Profiles included that group.

Fixes: #584, #583, #563, #555

### DIFF
--- a/BrainPortal/app/models/user.rb
+++ b/BrainPortal/app/models/user.rb
@@ -84,6 +84,7 @@ class User < ActiveRecord::Base
   before_create             :add_system_groups
   before_save               :encrypt_password
   before_save               :destroy_sessions_if_locked
+  before_save               :apply_access_profiles
   after_update              :system_group_site_update
   after_destroy             :destroy_system_group
   after_destroy             :destroy_user_sessions
@@ -394,6 +395,9 @@ class User < ActiveRecord::Base
   # Scans the list of AccessProfiles associated
   # with the current user, and makes sure the user
   # is a member of all the groups in all these profiles.
+  # "before save" callback, so that if any changes are made
+  # to the list of AccessProfiles, the group membership will
+  # be properly updated.
   # If a list of +remove_group_ids+ is supplied,
   # the user will be removed from these groups as
   # long as they are not also in any of the AccessProfiles.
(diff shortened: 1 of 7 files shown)


BrainPortal/app/views/signups/_signups_table.html.erb
```diff
@@ -140,6 +149,15 @@
      %>
 
     <%
+      t.column("Created", :created_at,
+        :sortable => true,
+        :sort_order => :desc) do |d|
+
+        d.created_at.to_s
```
prioux: Please use to_localtime so that the admin's prefered timezone is used.

BrainPortal/app/views/signups/show.html.erb
```diff
@@ -164,6 +164,13 @@ Below is a summary of the information in this account request.
     <td>&nbsp;</td>
   </tr>
 
+  <% if @current_user.present? && @current_user.has_role?(:admin_user) %>
+    <tr>
+      <th>Admin Comments</th>
+      <td> <%= @signup.admin_comment %> </td>
```
prioux: Use a &lt;pre&gt; inside the TD to preserve spacing

BrainPortal/app/models/user.rb
```diff
@@ -423,7 +427,8 @@ def union_group_ids_from_access_profiles
   # group 99, because it's present in ap2.
   def apply_access_profiles(remove_group_ids: [])
     gids = union_group_ids_from_access_profiles
-    self.group_ids = self.group_ids - remove_group_ids + gids # - and + are NOT COMMUTATIVE!
+    self.group_ids = self.group_ids - remove_group_ids + gids.to_set.to_a # - and + are NOT COMMUTATIVE!
```
prioux: That's the wrong place to fix the bug, the bug is in union_group_ids_from_access_profiles, which is supposed to return a list without duplications; so the code here shoudl have stayed the same and it's in the other method that we shoudl fix it.

prioux: Also, I'm not sure how this could ever create a duplication, since Rails manage the association table... ?

prioux: Darn, I was sure Rails would remove duplicates; so the bug fix (to be clean) should go at both places.
Here,

```ruby
self.group_ids = (self.group_ids - remove_group_ids + gids).uniq
```

PR #592, opened 2017-01-17
Task buttons

### LINKED ISSUE #588: Label for task recovery button is full of junk
For a task in status `Terminated` I can see these buttons: check out the 'recovery' button.

screen shot 2017-01-07 at 12 43 56

This bug was introduced by https://github.com/aces/cbrain/commit/b39dbc2a92102a2fac98bfb2ea35e81e3c82ab9f

There should be no error recovery for Terminated tasks.

crocodoyle:
Fixes #588, where failed tasks could be restarted in show page and the label for the retry button was wrong.

### DIFF
--- a/BrainPortal/app/views/tasks/show.html.erb
+++ b/BrainPortal/app/views/tasks/show.html.erb
@@ -60,7 +60,7 @@
     <% end %>
 
     <!-- FAILURE RECOVERY -->
-    <% if CbrainTask::FAILED_STATUS.include?(@task.status) %>
+    <% if @task.status =~ /^Failed*/ %>
       <%= link_to 'Retry Failed', { :action => 'operation', :operation => 'recover', 'tasklist[]' => @task.id }, :class  => "button menu_button", :method => :post %>
       <span class="vertical_bar">&#x2503;</span>
     <% end %>
--- a/BrainPortal/app/views/tasks/show.html.erb
+++ b/BrainPortal/app/views/tasks/show.html.erb
@@ -66,7 +66,7 @@
     <% end %>
 
     <!-- RESTART -->
-    <% if CbrainTask::FAILED_STATUS.include?(@task.status) or @task.status =~ /Completed|Terminated|Duplicated/ %>
+    <% if @task.status =~ /Completed|Terminated|Duplicated/ %>
       <%= link_to 'Restart At Setup', { :action => 'operation', :operation => 'restart_setup', 'tasklist[]' => @task.id }, :class  => "button menu_button", :method => :post %>
       <% if !(@task.status =~ /Completed/) %>
         <span class="vertical_bar">&#x2503;</span>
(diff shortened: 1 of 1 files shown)


BrainPortal/app/views/tasks/show.html.erb
```diff
@@ -61,29 +61,20 @@
...
-        label = (
-          case @task.status
-          when /Setup/       then "Retry Setting Up"
-          when /Cluster/     then "Retry Processing"
-          when /PostProcess/ then "Retry Post Processing"
-          end
-        )
-      %>
-      <%= link_to label, { :action => 'operation', :operation => 'recover', 'tasklist[]' => @task.id }, :class  => "button menu_button", :method => :post %>
+      <%= link_to 'Retry Failed', { :action => 'operation', :operation => 'recover', 'tasklist[]' => @task.id }, :class  => "button menu_button", :method => :post %>
```
prioux: I don't understand why:

* you removed the nice labels
* we're still allowing this button to appear for states where recovery is not allowed (e.g. 'Terminated', which is in `CbrainTask::FAILED_STATUS`

I'm a bit grumpy, why can't we get this to work just like I originally designed in back in APRIL 2010, where the button only appeared for the proper states?!?

crocodoyle: So what states are allowed? The image is wrong, and it isn't documented anywhere. The "nice labels" don't add anything to the user experience. I'm a bit grumpy because my tasks are all very poorly specified, as are the task states

crocodoyle: You're the one who asked me to use the CbrainTask::FAILED_STATUS

prioux: Right. Let me document more, sorry. As can be seen in the original code (introduced in commit b5297ced in April 2010 in https://github.com/aces/cbrain/commit/b5297ced#diff-c7aa0f93adfa9b6ebe3258e64f513f13R52 ), the test in the view code was

```ruby
if @task.status =~ /^Failed*/
```

This was changed to use CbrainTask::FAILED_STATUS this past November.

Since that list CbrainTask::FAILED_STATUS contains too many statuses, I suggest we create CbrainTask::RECOVERABLE_STATUS with only the good six status keywords (and we can redefined FAILED_STATUS to be that list plus 'Terminated').

prioux: Correction, the original code was introduced on May 3rd, 2010, in https://github.com/aces/cbrain/commit/14101130

prioux: What about the nice labels? Instead of "Retry Failed" we had a nice message that changed depending on the failure state... ?

crocodoyle: I thought "Retry Failed" was less confusing

Issue #595, opened 2017-02-06
Signups "comment" field is string instead of text


prioux:
People leaving long comments gets it all truncated because of the wrong type for the column. Migrate.

PR #600, opened 2017-02-22
Hide old signups

### LINKED ISSUE #472: Add logo configuration options
Right now the logos and favicon stuff is all hard coded, it would be nice to provide a way for someone deploying CBRAIN to change them to anything else.

In the CBRAIN service I manually changed `app/views/layout/_section_menu.html.erb`: the initial link the start_page_path is now:

``` erb
<%= link_to_if(current_user.present?, image_tag("CBRAIN-alpha-01.png", :style => "height: 1.3em; vertical-align: -0.3em;").html_safe , start_page_path) %>
```

and I had to wrap each other component inside the H1 within `<span>` elements.

In the file `app/views/layouts/application.html.erb` I changed the shortcut icon to:

``` html
<link rel="shortcut icon" type="image/png" href="/images/cbrain.png" />
```

To do:
- make both of these configurable and move the style info to the CSS file
- add the apple touch icons too, as described in https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

### LINKED ISSUE #596: Signups improvements
We need a better way to identify signup records that have been dealt with successfully:

* be able to mark them as 'ignore' forever, so that the 'Approve' link stop nagging us
* be able to recognize that the account has been created but with a login name different from the one in the signup record (so that the 'Approve' link disappear too)

There are no official links (association) between these models and I'd like it to stay that way, so maybe we can improve the account creation mechanism to mark the signup record even when the login name is different, by carrying the signup_id in a hidden field in the account creation form.

crocodoyle:
Provides a mechanism to hide/show Signups
Comments are now type text instead of string
Association between Signup and User by storing the user_id in the Signup
Adds fields to BrainPortal show pages to edit the filename of the logo for custom logos.

Fixes #596 and #595.
Also fixes #472.

### DIFF
--- a/BrainPortal/app/controllers/signups_controller.rb
+++ b/BrainPortal/app/controllers/signups_controller.rb
@@ -175,7 +175,7 @@ class SignupsController < ApplicationController
     end
 
     if @scope.custom[:view_hidden]
-      @base_scope = Signup
+      @base_scope = Signup.where(:hidden => true)
     else
       @base_scope = Signup.where(:hidden => false)
     end
--- a/BrainPortal/app/controllers/signups_controller.rb
+++ b/BrainPortal/app/controllers/signups_controller.rb
@@ -166,7 +166,20 @@ class SignupsController < ApplicationController
 
     scope_default_order(@scope, 'created_at', :desc)
 
-    @base_scope       = Signup
+    if params[:view_hidden].present?
+      if params[:view_hidden] == "true"
+        @scope.custom[:view_hidden] = true
+      else
+        @scope.custom[:view_hidden] = false
+      end
+    end
+
+    if @scope.custom[:view_hidden]
+      @base_scope = Signup
+    else
+      @base_scope = Signup.where(:hidden => false)
+    end
+
     @view_scope       = @scope.apply(@base_scope)
 
     # Prepare the Pagination object
@@ -198,6 +211,10 @@ class SignupsController < ApplicationController
 
     if params[:commit] =~ /Delete/i
       return delete_multi
+    end
+
+    if params[:commit] =~ /Hide/i
+      return hide_multi
     end
 
     # Default: unknown multi action?
@@ -264,6 +281,22 @@ class SignupsController < ApplicationController
 
     flash[:notice] = "Sent " + view_pluralize(count, "confirmation email") + "."
     render :action => :multi_action
+  end
+
+  def hide_multi #:nodoc:
+    reqids = params[:reqids] || []
+    reqs   = Signup.find(reqids)
+
+    count = 0
+    reqs.each do |req|
+      req[:hidden] = true
+      req.save
+      count += 1
+    end
+
+    flash[:notice] = "Hid " + view_pluralize(count, "record")
+
+    redirect_to :action => :index
   end
 
   private
--- a/BrainPortal/app/controllers/signups_controller.rb
+++ b/BrainPortal/app/controllers/signups_controller.rb
@@ -217,6 +217,10 @@ class SignupsController < ApplicationController
       return hide_multi
     end
 
+    if params[:commit] =~ /Show/i
+      return show_multi
+    end
+
     # Default: unknown multi action?
     redirect_to :action => :index
   end
@@ -295,6 +299,22 @@ class SignupsController < ApplicationController
     end
 
     flash[:notice] = "Hid " + view_pluralize(count, "record")
+
+    redirect_to :action => :index
+  end
+
+  def show_multi #:nodoc:
+    reqids = params[:reqids] || []
+    reqs   = Signup.find(reqids)
+
+    count = 0
+    reqs.each do |req|
+      req[:hidden] = false
+      req.save
+      count += 1
+    end
+
+    flash[:notice] = "Unhid " + view_pluralize(count, "record")
 
     redirect_to :action => :index
   end
--- a/BrainPortal/app/controllers/signups_controller.rb
+++ b/BrainPortal/app/controllers/signups_controller.rb
@@ -166,22 +166,15 @@ class SignupsController < ApplicationController
 
     scope_default_order(@scope, 'created_at', :desc)
 
-    if params[:view_hidden].present?
-      if params[:view_hidden] == "true"
-        @scope.custom[:view_hidden] = true
-        @base_scope = Signup.where(:hidden => true)
-      end
-    else
-      @scope.custom[:view_hidden] = false
-      @base_scope = Signup.where(:hidden => false)
-    end
-
-    @view_scope       = @scope.apply(@base_scope)
+    view_hidden                 = params[:view_hidden].presence == "true" ? true : false
+    @scope.custom[:view_hidden] = view_hidden
+    @base_scope                 = Signup.where(:hidden => view_hidden)
+    @view_scope                 = @scope.apply(@base_scope)
 
     # Prepare the Pagination object
-    @scope.pagination ||= Scope::Pagination.from_hash({ :per_page => 25 })
+    @scope.pagination           ||= Scope::Pagination.from_hash({ :per_page => 25 })
 
-    @signups          = @scope.pagination.apply(@view_scope)
+    @signups                    = @scope.pagination.apply(@view_scope)
 
     scope_to_session(@scope, 'signups')
 
@@ -287,14 +280,12 @@ class SignupsController < ApplicationController
     reqids = params[:reqids] || []
     reqs   = Signup.find(reqids)
 
-    count = 0
     reqs.each do |req|
       req[:hidden] = true
       req.save
-      count += 1
     end
 
-    flash[:notice] = "Hid " + view_pluralize(count, "record")
+    flash[:notice] = "Hid " + view_pluralize(reqs.size, "record")
 
     redirect_to :action => :index
   end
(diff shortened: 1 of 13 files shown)


BrainPortal/app/controllers/signups_controller.rb
```diff
@@ -166,7 +166,16 @@ def index #:nodoc:
 
     scope_default_order(@scope, 'created_at', :desc)
 
-    @base_scope       = Signup
+    if params[:view_hidden].present?
```
prioux: This entire block of added lines is clumsy. Why not simply:

```ruby
   view_hidden = params[:view_hidden].presence == "true"
   @scope.custom[:view_hidden] = view_hidden
   @base_scope = Signup.where(:hidden => view_hidden)
```

THis doesn the same thing, in three lines instead of nine, without an `if..end` block inside a `if...else...end`

BrainPortal/app/controllers/signups_controller.rb
```diff
@@ -266,6 +283,38 @@ def resend_conf_multi #:nodoc:
     render :action => :multi_action
   end
 
+  def hide_multi #:nodoc:
+    reqids = params[:reqids] || []
+    reqs   = Signup.find(reqids)
+
+    count = 0
```
prioux: why a `count` variable that you increment once, when you can just use `reqs.size` ?!? It's only used for the `pluralize` message after all.

BrainPortal/app/controllers/users_controller.rb
```diff
@@ -141,12 +141,13 @@ def create #:nodoc:
       flash[:notice] = "User successfully created."
 
       # Find signup record matching login name, and log creation and transfer some info.
-      if signup = Signup.where(:login => @user.login, :approved_at => nil, :approved_by => nil).first
+      if signup = Signup.where(:id => params[:signup_id]).first
```
prioux: Why remove the safety checks for the record not having been approved before? The way it is set up it if there is a mixup in the signup_id, the wrong record will be adjusted.

BrainPortal/app/models/signup.rb
```diff
@@ -72,15 +72,15 @@ def user_full #:nodoc:
   alias full_name full
 
   def approved? #:nodoc:
-    self.approved_by.present? && self.approved_at.present?
+    self.user_id.present?
```
prioux: This will make all previous records already approved seem to no longer be approved! PLease return the check to what it was.

BrainPortal/db/schema.rb
```diff
@@ -117,6 +117,35 @@
   add_index "data_providers", ["type"], :name => "index_data_providers_on_type"
   add_index "data_providers", ["user_id"], :name => "index_data_providers_on_user_id"
 
+  create_table "demands", :force => true do |t|
```
prioux: This is a spurious table specification. Please make sure your dev DB is in sync with the schema.

natacha-beck:
@prioux If I am right @crocodoyle has changed his code based on your comments. Can you confirm @crocodoyle ?

prioux:
This PR crashes:

```
undefined method `small_logo' for #<BrainPortal:0x007fece179b398>
at
app/views/layouts/application.html.erb line 29
```

There seems to be an attribute `small_logo` and `large_logo` added to the schema in the Signups model (why?!?) and no migration that actually adds it to any models (whether BrainPortal or Signup).

What is that attribute anyway and why is it needed?

prioux:
Please revert all the stuff related to small_logo and large_logo ; it's not part of the feature request, and right now I don't want to pollute our RemoteResource model with one more UI-related attribute. It's already dirty enough.

BrainPortal/app/controllers/signups_controller.rb
```diff
@@ -266,6 +276,34 @@ def resend_conf_multi #:nodoc:
...
+      req[:hidden] = true
+      req.save
+    end
+
+    flash[:notice] = "Hid " + view_pluralize(reqs.size, "record")
+
+    redirect_to :action => :index
+  end
+
+  def show_multi #:nodoc:
```
prioux: Not sure: how will the admin be able to select the records to SHOW if they are already hidden?!?

BrainPortal/app/views/users/new.html.erb
```diff
@@ -76,6 +76,8 @@
     <p><label for="no_password_reset_needed">No need to reset initial password:</label>
     <%= check_box_tag :no_password_reset_needed, "1", params[:no_password_reset_needed] == "1" %>
 
+    <%= hidden_field_tag :signup_id, params[:signup_id] %>
```
prioux: What is this used for?!? There is no code in users_controller that reads that value.

On the other hand, it SHOULD be used in the `create` action, but the cold there has not been adjusted:

```ruby
# Line 144 of users_controller
if signup = Signup.where(:login => @user.login, blahbah)
```

BrainPortal/app/controllers/signups_controller.rb
```diff
@@ -166,13 +166,15 @@ def index #:nodoc:
 
     scope_default_order(@scope, 'created_at', :desc)
 
-    @base_scope       = Signup
-    @view_scope       = @scope.apply(@base_scope)
+    view_hidden                 = params[:view_hidden].presence == "true" ? true : false
```
crocodoyle: How do you want me to fix this? I tested it and it works.

BrainPortal/app/controllers/signups_controller.rb
crocodoyle: Did you try it?

BrainPortal/app/views/users/new.html.erb
crocodoyle: It is used in the `new` action so that it is preserved when the form fails to validate

BrainPortal/app/controllers/signups_controller.rb
prioux: I'm sure it works, I'm just saying why the complex code when you can simply do

```ruby
view_hidden = params[:view_hidden].presence == "true"
```

prioux: Or add a paren to make it more clear

```ruby
view_hidden = ( params[:view_hidden].presence == "true" )
```

prioux: I was saying there's never any reason to have

```ruby
    boolean_expression ? true : false
```

in any piece of code, it's redundant.

BrainPortal/app/views/users/new.html.erb
prioux: If the form fails to validate, the new() action is not invoked at all, and we don't need to signup_id anymore, it was already used to populate all the fields. It IS however needed in create() when making the association back to the signup models.

crocodoyle:
I would like to keep the UI-related logo, as it addresses issue #472 and this is the way you instructed me to do it.

BrainPortal/app/controllers/signups_controller.rb
prioux: I couldn't, this PR crashes, like I said in my frist comment.

prioux:
OK we'll leave the logos in. Let's fix the rest.

BrainPortal/db/migrate/20170220152128_signups_tweaks.rb
```diff
@@ -0,0 +1,14 @@
+class SignupsTweaks < ActiveRecord::Migration
```
prioux: This migration should fill the new user_id column for all the records already present and approved, the best it can, using the old logic of matching login names.

BrainPortal/app/controllers/signups_controller.rb
prioux: OK, so there's a button "show hidden" that shows all record, even those hidden. Now if I want to unhide a few of them, among that list, how can I tell which were hidden and which are not?

prioux: I think a (Hidden) keyword shoudl be shown in the status field, in the same color as in the rest of CBRAIN (purple)

BrainPortal/app/views/signups/_signups_table.html.erb
```diff
@@ -33,6 +33,14 @@
     <%= submit_tag "Delete",                :class => "button" %>
     <%= submit_tag "Adjust Login",          :class => "button" %>
     <%= submit_tag "Resend Confirm Email",  :class => "button" %>
+
+    <% if @scope.custom[:view_hidden] %>
+      <%= submit_tag "Unhide", :class => "button" %>
```
prioux: The unhide button doesn't work at all

BrainPortal/app/controllers/signups_controller.rb
crocodoyle: It displays either hidden or not hidden, not both. OK I will rework this I guess

BrainPortal/db/migrate/20170220152128_signups_tweaks.rb
```diff
@@ -5,6 +5,13 @@ def up
     add_column    :signups, :user_id, :integer
     add_column    :signups, :hidden, :boolean, :default => false
 
+    Signup.all.each do |d|
+      user = User.where(:login => d.login).first
+      if user.present?
+        user[:signup_id] = user.id
```
prioux: How can this even work? There is no signup_id in the user model! Have you even TESTED that code?

BrainPortal/db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20170308220959) do
+ActiveRecord::Schema.define(:version => 20170306210331) do
```
prioux: This schema contains all sorts of changes completely unrelated to this pull request. Please checkout the proper schema and ONLY integrate the changes that are relevant

PR #603, opened 2017-03-03
Network Operation Center: daily activity panel.

prioux:
This is a feature not normally enabled in the framework, but available if someone wants it.

It shows a single status page, updated every two minutes. The top two numbers are live, they represent the number of active tasks, and the total size of the files currently in transfer between the CBRAIN components. The information below that is about the cumulative activity on the system since midnight: task space on the execution servers, cache space on all remote resources, and sum of sizes of all updated files, for each data provider (which can be a bit misleading, since just renaming a file marks it as updated).

Boxes appear only when their numbers are non-zero, so they all disappear at midnight, except for remote resource that are offline, which are always shown and highlighted in red too.

It can be enabled by editing the `routes.rb` file, and the path to the page can be arbitrary (it doesn't have to match the `/noc/daily` path which corresponds by default to the controller and action; one could map it to `/summary` or `/i/am/a/big/bozo`).


prioux:
Andrew, I don't want the NOC to require logging in, for security. That's why it's not in the portal controller, and in fact I put the whole thing as a separate feature so that it can be ripped out without affecting anything else (`rm -tf app/controllers/noc_controller.rb app/views/noc public/stylesheets/noc.css`).

PR #608, opened 2017-03-13
Edit site group

### LINKED ISSUE #594: Cannot edit site group
No method `can_be_edited_by?` in SiteGroup model; should be the same as WorkGroup, possibly (without the test for invisibility)

crocodoyle:
Adds the ability to edit a Site Group. Fixes #594

### DIFF
--- a/BrainPortal/app/controllers/groups_controller.rb
+++ b/BrainPortal/app/controllers/groups_controller.rb
@@ -132,7 +132,7 @@ class GroupsController < ApplicationController
   # PUT /groups/1.xml
   # PUT /groups/1.json
   def update #:nodoc:
-    @group = current_user.available_groups.where( :type => "WorkGroup" ).find(params[:id])
+    @group = current_user.available_groups.find(params[:id])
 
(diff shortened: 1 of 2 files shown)


BrainPortal/app/controllers/groups_controller.rb
```diff
@@ -132,7 +132,7 @@ def create  #:nodoc:
   # PUT /groups/1.xml
   # PUT /groups/1.json
   def update #:nodoc:
-    @group = current_user.available_groups.where( :type => "WorkGroup" ).find(params[:id])
+    @group = current_user.available_groups.find(params[:id])
```
crocodoyle: No, this will allow the admin from modifying groups other than WorkGroups (like SiteGroups). Before, the update action could not find groups other than WorkGroups, so the admin could not update them.

prioux: You're right, I read the diff backwards! My bad.

prioux: Red is "removed", green is "added". Gotta remember that.

PR #634, opened 2017-06-05
Mostly Docker and Singularity fixes

prioux:
Code was cleaned, variables renamed, and some bug fixes and improvements.

* Launched container scripts checks for proper UID, refuse to work otherwise
* Boutique integration creates its tool configs with group_id limited to 'admin'.
* All special basenames used in docker and singularity have proper run_id in them, to prevent collisions.
* diagnostics tasks enhanced with a write test, and test input files using both the cache_full_path and the relative symlink created with make_available()

KNOWN ISSUE: For singularity, tasks that work with the cache_full_path of a userfile won't work, it will work only if the task creates a link with make_available(). That's because singularity has no mechanism for binding a arbitrary path if the path components don't already exist in the container. Darn.

### DIFF
--- a/BrainPortal/app/controllers/userfiles_controller.rb
+++ b/BrainPortal/app/controllers/userfiles_controller.rb
@@ -1139,7 +1139,7 @@ class UserfilesController < ApplicationController
     send_file tarfile, :stream  => true, :filename => tarfile_name
     CBRAIN.spawn_fully_independent("Download Clean Tmp #{current_user.login}") do
       sleep 3000
-      File.unlink(tarfile)
+      File.delete(tarfile)
     end
   end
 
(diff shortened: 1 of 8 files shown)


prioux:
Note about the KNOWN ISSUE: maybe one day singularity will work with binding arbitrary path, so we could in the meantime include

```
-B cache_root:cache_root
```

to the singularity command-line. I checked, right now Singularity just ignores it with a warning:

```
WARNING: Skipping user bind, non existant bind point (directory) in container: '/data1/user-data/prioux/Bourreau-Acelab-VH-2/DpCache'
```

natacha-beck:
Thanks, for the refactoring. Really appreciate the work and the clarity. 
Seems ok from my point of view. @glatard would you like to review this one ? Otherwise I will merged it.

PR #642, opened 2017-08-16
Singularity/Docker integration

natacha-beck:


BrainPortal/app/models/cluster_task.rb
```diff
@@ -2168,13 +2132,32 @@ def singularity_commands(command_script)
...
+# CBRAIN internal consistency test 1: must run under proper UID
 if test "$UID" -ne "#{Process.uid}" ; then
   echo "Singularity internal script running with wrong UID (expected UID=#{Process.uid})"
   echo "Runtime IDs: `id`"
   exit 2
 fi
 
+# CBRAIN internal consistency test 2: must have the task workdir mounted inside the container
+if test ! -d #{task_workdir.bash_escape} ; then
+  echo "Missing mount (#{task_workdir.bash_escape}) point inside the container"
```
prioux: This echo statement will fail if there are double quotes in the task workdir path. It's supposed to be, to be safe:

```
echo "Container missing mount point for task work directory:" #{task_workdir.bash_escape}
```

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2168,13 +2132,32 @@ def singularity_commands(command_script)
...
+  exit 2
+fi
+
+# CBRAIN internal consistency test 3: must have the cache_dir mounted inside the container
+if test ! -d #{cache_dir.bash_escape} ; then
+  echo "Missing mount (#{cache_dir.bash_escape}) inside the container"
```
prioux: See previous comment.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2168,13 +2132,32 @@ def singularity_commands(command_script)
...
+  exit 2
+fi
+
+
+# CBRAIN internal consistency test 4: must have the gridshare_dir mounted inside the container
+if test ! -d #{gridshare_dir.bash_escape} ; then
+  echo "Missing mount (#{gridshare_dir.bash_escape}) inside the container"
```
prioux: See previous comment too.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2350,6 +2330,47 @@ def validate_json_string json_string
     JSON::Validator.validate!(schema,json_string) # raises an exception if json_string is not valid
   end
 
+  ##################################################################
+  # Singularity load
+  ##################################################################
+
+  private
+
+  def container_image_name
```
prioux: Missing method description and :nodoc:

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2350,6 +2330,47 @@ def validate_json_string json_string
...
+  end
+
+  # Install the singularity image as a local file.
+  def load_singularity_image #:nodoc:
+    singularity_image_name = self.tool_config.containerhub_image_name
+
+    # Singularity PULL
+    if singularity_image_name.present?
+      begin
+        errfile = ".container_load_cmd.#{self.run_id}.err"
```
prioux: Send the error file to /tmp instead of the local directory

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2350,6 +2330,47 @@ def validate_json_string json_string
...
+      ensure
+        File.unlink(errfile)    rescue true
+      end
+    end
+
+    # Singularity registration (local image)
+    if singularity_image = self.tool_config.container_image # = not ==
+      # Create a link to our image; the image is a registered CBRAIN file
+      self.addlog("Syncing the singularity image")
+      singularity_image    = self.tool_config.container_image
```
prioux: This assignment to the variable `singularity_image` repeats exactly what is in the IF statement of the block

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -226,20 +228,22 @@ def register(task)
       # theres already one. Only applies to Bourreaux (as it would make no
       # sense on the portal).
       return if Rails.root.to_s =~ /BrainPortal\z/
-
       # Create a ToolConfig iff
       #   (1) the Bourreau has a docker executable and
       #   (2) the descriptor specifies a docker image
-      return if docker_image.nil? || !resource.docker_present
+      return if container_image.nil? || container_engine.nil?
```
prioux: Why use `.nil?` This will fail if the value is an empty string, because `"".nil?` returns false

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2168,13 +2132,32 @@ def singularity_commands(command_script)
...
 
+# CBRAIN internal consistency test 2: must have the task workdir mounted inside the container
+if test ! -d #{task_workdir.bash_escape} ; then
+  echo "Container missing mount point for task work directory:" #{task_workdir.bash_escape}
+  exit 2
+fi
+
+# CBRAIN internal consistency test 3: must have the cache_dir mounted inside the container
+if test ! -d #{cache_dir.bash_escape} ; then
+  echo "Container missing mount point for task work directory:" #{cache_dir.bash_escape}
```
prioux: The message needs to be adjusted, it talks about the task workdir but it's the cache dir instead

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2168,13 +2132,32 @@ def singularity_commands(command_script)
...
+  exit 2
+fi
+
+
+# CBRAIN internal consistency test 4: must have the gridshare_dir mounted inside the container
+if test ! -d #{gridshare_dir.bash_escape} ; then
+  echo "Container missing mount point for task work directory:" #{gridshare_dir.bash_escape}
```
prioux: Adjuste the message too

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2350,6 +2330,46 @@ def validate_json_string json_string
     JSON::Validator.validate!(schema,json_string) # raises an exception if json_string is not valid
   end
 
+  ##################################################################
+  # Singularity load
+  ##################################################################
+
+  private
+
+  def container_image_name #:nodoc:
```
prioux: Please add a description for the method even if it's :nodoc:

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2350,6 +2330,46 @@ def validate_json_string json_string
...
+    ".container-#{self.run_id}.img"
+  end
+
+  # Install the singularity image as a local file.
+  def load_singularity_image #:nodoc:
+    singularity_image_name = self.tool_config.containerhub_image_name
+
+    # Singularity PULL
```
prioux: 1) I would give priority to the singularity local image over the singularity pull
2) I would split this method into three:
   one method does the if/else
   one method for sing pull
   on method for local sing image
3) I woudl add a addlog() before doing singularity pull so that the task's log

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -226,20 +228,23 @@ def register(task)
...
       return if Rails.root.to_s =~ /BrainPortal\z/
-
       # Create a ToolConfig iff
-      #   (1) the Bourreau has a docker executable and
-      #   (2) the descriptor specifies a docker image
-      return if docker_image.nil? || !resource.docker_present
+      #   (1) the Bourreau has a container executable and
+      #   (2) the descriptor specifies a container image
+      return if container_engine.blank?  # Singularity or Docker
+      return if container_image.blank?   # image name or ID??
```
prioux: In my slack comment I didn't mean to have the "image name or ID??" be a literal comment, I meant for you to replace it with the proper one...

PR #648, opened 2017-09-12
Fixed Boutiques bug in save_results() implementation.

### LINKED ISSUE #647: Boutiques integration code mutates task params; makes task inconsistent
The Boutiques integration code generate a save_results() method which has a serious bug:

* At some point it does `params.extract!(*[ key, key, ...])` for the output files and folders, which mutates the params hash
* If anything goes wrong while creating the  output files, then the params hash stay that way, missing the descriptor information
* Such a task can never be restarted properly, resulting in check for `File.exists?(nil)` which triggers the exception shown below:

```
[2017-09-12 12:17:00 EDT] post_process() Exception raised while post processing results: TypeError: no implicit conversion of nil into String
[2017-09-12 12:17:00 EDT] /home/prioux/CBRAIN/Bourreau/app/models/cluster_task.rb:781:in `block (2 levels) in post_process'
[2017-09-12 12:17:00 EDT] /home/prioux/CBRAIN/Bourreau/app/models/cluster_task.rb:775:in `chdir'
```

The code generated must be modified so that the params hash is not modified that way, and the record of the created outputs is saved somewhere else in it.

prioux:
Will fix #647 . Also added feature that we can get the dump of the instanciated templates using environment variables. Example:

```
unix% mkdir /tmp/dump;CBRAIN_BOUTIQUES_DUMPDIR=/tmp/dump CBRAIN_BOUTIQUES_DUMPTASK=NiakFmriPreprocess rails console
```

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
+++ b/BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
@@ -181,7 +181,7 @@ module SchemaTaskGenerator
 
         # Redefine the CbrainTask or Object constant pointing to the task's
         # class to point to the switcher instead.
-        [ Object, CbrainTask ].select { |m| m.const_defined?(name) }.each do |m|
+        [ CbrainTask ].select { |m| m.const_defined?(name) }.each do |m|
           m.send(:remove_const, name)
           m.const_set(name, switcher)
         end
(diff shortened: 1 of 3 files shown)


prioux:
Seems to work, tested on Graham.

PR #649, opened 2017-09-13
Modified Boutiques hooks for 0.5

gkiar:
Fixed backward compatibility breaking changes when going from Boutiques 0.4 to 0.5. This includes:
  - replacing access of `["walltime-estimate"]` to `["suggested-resources"]["walltime-estimate"]`
  - now treating `singularity` like `docker` with image and index names
  - now checking the `index` field for docker and singularity pull locations, rather than assuming "docker.io" and "shub://", respectively

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2367,8 +2367,7 @@ chmod 755 #{singularity_wrapper_basename.bash_escape}
   # Perform the singularity pull
   def load_singularity_image_from_repo #:nodoc:
     singularity_image_name = self.tool_config.containerhub_image_name
-    singularity_index_location = self.tool_config.container_index_location
-    singularity_index_location = singularity_index_location.presence || "shub://"
+    singularity_index_location = self.tool_config.container_index_location.presence || "shub://"
 
     self.addlog("Pulling singularity image '#{singularity_image_name}'")
 
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2050,7 +2050,7 @@ chmod 755 #{docker_wrapper_basename.bash_escape}
   def load_docker_image_cmd #:nodoc:
     dockerhub_image_name = self.tool_config.containerhub_image_name
     docker_container_index = self.tool_config.container_index_location
-    docker_container_index += '/' if (!docker_container_index.end_with? '/') && (docker_container_index.length > 0) # Add trailing / if not present and field not empty
+    docker_container_index += '/' if docker_container_index.present? && !docker_contianer_index.end_with?('/') # Add trailing / if not present and field not empty
     # Docker PULL
     if dockerhub_image_name.present?
       return <<-DOCKER_PULL
@@ -2368,7 +2368,7 @@ chmod 755 #{singularity_wrapper_basename.bash_escape}
   def load_singularity_image_from_repo #:nodoc:
     singularity_image_name = self.tool_config.containerhub_image_name
     singularity_index_location = self.tool_config.container_index_location
-    singularity_index_location = "shub://" if singularity_index_location.length < 1
+    singularity_index_location = "shub://" if singularity_index_location.blank?
 
     self.addlog("Pulling singularity image '#{singularity_image_name}'")
 
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2050,7 +2050,14 @@ chmod 755 #{docker_wrapper_basename.bash_escape}
   def load_docker_image_cmd #:nodoc:
     dockerhub_image_name = self.tool_config.containerhub_image_name
     docker_container_index = self.tool_config.container_index_location
-    docker_container_index += '/' if docker_container_index.present? && !docker_container_index.end_with?('/') # Add trailing / if not present and field not empty
+    if docker_container_index.present?
+      if !docker_container_index.end_with?('/')
+        docker_container_index += '/'
+      end
+    else
+      docker_container_index = ''
+    end
+    docker_container_index = docker_container_index + '/' if docker_container_index.present? && !docker_container_index.end_with?('/') # Add trailing / if not present and field not empty
     # Docker PULL
     if dockerhub_image_name.present?
       return <<-DOCKER_PULL
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2050,7 +2050,7 @@ chmod 755 #{docker_wrapper_basename.bash_escape}
   def load_docker_image_cmd #:nodoc:
     dockerhub_image_name = self.tool_config.containerhub_image_name
     docker_container_index = self.tool_config.container_index_location
-    docker_container_index += '/' if docker_container_index.present? && !docker_contianer_index.end_with?('/') # Add trailing / if not present and field not empty
+    docker_container_index += '/' if docker_container_index.present? && !docker_container_index.end_with?('/') # Add trailing / if not present and field not empty
     # Docker PULL
     if dockerhub_image_name.present?
       return <<-DOCKER_PULL
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2049,12 +2049,13 @@ chmod 755 #{docker_wrapper_basename.bash_escape}
   # Then name of the image will be set in a bash variable 'docker_image_name'
   def load_docker_image_cmd #:nodoc:
     dockerhub_image_name = self.tool_config.containerhub_image_name
-
+    docker_container_index = self.tool_config.container_index_location
+    docker_container_index += '/' if (!docker_container_index.end_with? '/') && (docker_container_index.length > 0) # Add trailing / if not present and field not empty
     # Docker PULL
     if dockerhub_image_name.present?
       return <<-DOCKER_PULL
 # Pull image from DockerHub
-#{docker_executable_name} pull #{dockerhub_image_name.bash_escape}
+#{docker_executable_name} pull #{docker_container_index.bash_escape}#{dockerhub_image_name.bash_escape}
 docker_image_name=#{dockerhub_image_name.bash_escape}
       DOCKER_PULL
     end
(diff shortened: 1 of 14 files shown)


prioux:
Pull request will be merged when we have confirmation that ALL external plugins that use Boutiques descriptors have been updated.

natacha-beck:
The information about the `container_index_location` should be available on the show page of the tool_config to.

prioux:
Natacha's comment is important, we've no way of setting the value of the new attribute.

natacha-beck:
List of plugin that should be updated: 

- [ ] eeg ( @glatard )
- [ ] hpc ( @glatard )
- [ ] psom ( @poquirion )
- [ ] experimental ( @natacha-beck )
- [ ] neuro ( @natacha-beck )
- [x] nereudata ( @gkiar )

gkiar:
@prioux @natacha-beck I have added the `container_index_location` to the "Edit" and "Tool Versions" pages. I looked at the "Show" page, and noticed that there wasn't any container-related information contained here and based on how I interpreted the content of that page, I'm not sure where we would put it. Please advise if you do still want all container things on this "Show" page, otherwise, I believe the requested modifications are complete. Thanks @prioux for much help debugging today!

prioux:
@gkiar I think this is what I documented in #650, so for the moment we'll add the info as a separate PR.

gkiar:
@prioux I agree- I suppose we're just waiting on the remaining descriptors, and then a review of my latest modifications and any suggestions that arise there. Hopefully @natacha-beck and I will finish testing on her local install this week!

BrainPortal/app/models/tool_config.rb
```diff
@@ -56,14 +56,19 @@ class ToolConfig < ActiveRecord::Base
                                    :message => "must be unique per pair [tool, server]" },
                   :if         => :applies_to_bourreau_and_tool?
 
+  validates                 :container_index_location,
+                            :format => { :with => /^[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,6}$/i },
```
natacha-beck: Failed to validate "shub://" as an index. Should allow it, I had this format of regex: 

`:format => { :with => /^[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,6}$|^shub\:\/\/$/}`

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2049,12 +2049,13 @@ def docker_commands(command_script)
   # Then name of the image will be set in a bash variable 'docker_image_name'
   def load_docker_image_cmd #:nodoc:
     dockerhub_image_name = self.tool_config.containerhub_image_name
-
+    docker_container_index = self.tool_config.container_index_location
+    docker_container_index += '/' if docker_container_index.present? && !docker_contianer_index.end_with?('/') # Add trailing / if not present and field not empty
```
natacha-beck: Typo --> `docker_contianer_index` should be `docker_container_index`

BrainPortal/app/models/tool_config.rb
```diff
@@ -337,7 +339,18 @@ def container_rules #:nodoc:
     if self.container_engine.present? && ( self.containerhub_image_name.blank? && self.container_image_userfile_id.blank? )
       errors[:container_engine] = "a container hub image name or a container image userfile ID should be set when the container engine is set"
     end
+
+    if self.container_engine.present? && self.container_engine == "Singularity" 
+      if self.container_index_location.present? && ! /^[a-z0-9]+\:\/\/$/.match(self.container_index_location)
+        errors[:container_index_location] = "Invalid index location type for container engine Singularity. Should end in '://'."
```
prioux: The error messages are generated automatically using the attribute name as a subject for the sentence; so right now the error message will look weird; test it with the interface and read it carefully.

BrainPortal/app/views/tool_configs/_form_fields.html.erb
```diff
@@ -112,6 +112,16 @@
     </p>
   </span>
 
+  <span title="Index of the container image">
+    <p><%= f.label :container_index_location, "Index of the container image:" %><br/>
+      <%= f.text_field :container_index_location %><br>
+      <div class="field_explanation">
+        The index/url of the Container image in which the docker/singularity container is
```
prioux: `Container` is capitalized for no reason; add examples for singularity too (and show default for singularity)

BrainPortal/app/views/tool_configs/_tool_configs_table.html.erb
```diff
@@ -104,6 +104,13 @@
         end
       }
 
+    t.column("Container Index", :container_index_location,
+    ) { |tc|
+        if tc.container_index_location.present?
```
prioux: the whole if block can be replaced by

```ruby
tc.container_index_location
```

there is no need to check for presence/blankness etc.

BrainPortal/db/migrate/20170913185428_add_container_index_to_tool_configs.rb
```diff
@@ -0,0 +1,5 @@
+class AddContainerIndexToToolConfigs < ActiveRecord::Migration
```
prioux: Add data migration to populate existing TCs with the new attribute filled by parsing the `containerhub_image_name` content; you will have to rewrite the migration to contain an explicit `up` and `down` method

gkiar: Addressed in https://github.com/aces/cbrain/pull/649/commits/f1313b2fad911425b7dee7f0686510892d46deae

BrainPortal/app/models/tool_config.rb
```diff
@@ -56,14 +56,16 @@ class ToolConfig < ActiveRecord::Base
                                    :message => "must be unique per pair [tool, server]" },
                   :if         => :applies_to_bourreau_and_tool?
 
+
```
prioux: spurious blank line

BrainPortal/app/models/tool_config.rb
```diff
@@ -337,7 +339,18 @@ def container_rules #:nodoc:
...
+      if self.container_index_location.present? && ! /^[a-z0-9]+\:\/\/$/.match(self.container_index_location)
+        errors[:container_index_location] = "is invalid for container engine Singularity. Should end in '://'."
+      end
+    elsif self.container_engine.present? && self.container_engine == "Docker"
+      if self.container_index_location.present? && ! /^[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,6}$/.match(self.container_index_location)
+        errors[:container_index_location] = "is invalid for container engine Docker. Should be a valid hostname."
+      end
+    end
     return errors.empty?
+
```
prioux: spurious blank line too

BrainPortal/app/views/tool_configs/_form_fields.html.erb
```diff
@@ -104,19 +104,31 @@
...
-        The name and tag of the Container image in which the tool is installed,
+        The name and tag of the container image in which the tool is installed,
         for instance "centos:latest". This name refers to the Docker/Singulariry index
         accessed by the Bourreau, which is configured manually in the Bourreau
         for now.
       </div>
     </p>
   </span>
 
+  <span title="Index of the container image">
```
prioux: The order of the columns should be more intuitieve, the index shoudl be to the left of the image name, since URLS are built like that. This is what it looks like in my browser:

screen shot 2017-09-19 at 16 26 42

gkiar:
Addressed all changes- @prioux

BrainPortal/config/console_rc/lib/pretty_view.rb
```diff
@@ -342,6 +342,7 @@ def pretview
       bourreau_id.presence || 0, bourreau.try(:name),
       ncpus.presence || 0,
       container_engine.presence || "(none)",
+      container_index_location || "",
```
prioux: You cannot just add an argument to the sprintf without adjusting the control string accordingly. This will result in out of sync args . Please fix this so that the index is shown as a prefix to the image name on the same line. And please test it before you commit.

prioux: Right now `pv` doesn't work at all for any ToolConfigs because of the mismatch.

gkiar: Made the change locally - I haven't tested this beyond running the tests (which it passes) because I don't know how to find where this view is exposed in the web app, and cannot find documentation that describes this....

natacha-beck:
@prioux  and @gkiar all the plugin is updated now.

BrainPortal/config/console_rc/lib/pretty_view.rb
natacha-beck: I just test it in the console seems to be ok.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2049,12 +2049,13 @@ def docker_commands(command_script)
   # Then name of the image will be set in a bash variable 'docker_image_name'
   def load_docker_image_cmd #:nodoc:
     dockerhub_image_name = self.tool_config.containerhub_image_name
-
+    docker_container_index = self.tool_config.container_index_location
```
prioux: This will crash if `container_index_location` is `nil` when the value is substituted in the `pull` command below.

BrainPortal/app/models/tool_config.rb
```diff
@@ -337,8 +338,17 @@ def container_rules #:nodoc:
     if self.container_engine.present? && ( self.containerhub_image_name.blank? && self.container_image_userfile_id.blank? )
       errors[:container_engine] = "a container hub image name or a container image userfile ID should be set when the container engine is set"
     end
+
+    if self.container_engine.present? && self.container_engine == "Singularity" 
+      if self.container_index_location.present? && ! /^[a-z0-9]+\:\/\/$/.match(self.container_index_location)
```
prioux: The regex should be specified with the `/i` modifier. Also, it is better (IMHO) to use the syntax

```ruby
   $var !~ /regex/
```

instead of

```ruby
! /regex/.match(var)
```

prioux: There are two places where this comment applies (the other one is about 10 lines below).

BrainPortal/app/models/cluster_task.rb
gkiar: Addressed in https://github.com/aces/cbrain/pull/649/commits/9668f65b4e7cad30cdd00d548f108edaf0f76ffb

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2050,7 +2050,14 @@ def docker_commands(command_script)
   def load_docker_image_cmd #:nodoc:
     dockerhub_image_name = self.tool_config.containerhub_image_name
     docker_container_index = self.tool_config.container_index_location
-    docker_container_index += '/' if docker_container_index.present? && !docker_container_index.end_with?('/') # Add trailing / if not present and field not empty
+    if docker_container_index.present?
```
prioux: Well, that will work, but it would have been simpler to change line 2052 to

```ruby
docker_container_index = self.tool_config.container_index_location.presence || ""
```

shots47s:
Should this Issue be closed, what is still active on this?

PR #665, opened 2017-11-17
fix content loader to render html files but send other file types

ArminTaheri:
We need some way to render html files through content_loader. I don't believe this solution is the best solution so I will wait until we discuss this in the next cbrain meeting.


ArminTaheri:
The main issue here is that a file collection may change file types when `content_loader` is fetching a particular file in this file collection. Right now Niak Results are always using `:send_file` as the `content_loader.type` which currently forces the client to download the file no matter what. But for HTML files in a file collection it would be handy to render them.

My instinct tells me it should be up to the plugin to decide what exact code should run when `userfile#content` is invoked and so perhaps we could have an `:proc` `content_loader.type` so the plugin can simply return a proc for the `userfile#content` to invoke on behalf of the plugin.

This way `niak_result.rb` in https://github.com/SIMEXP/cbrain-plugins-psom can decide to render html files and send other files as downloads.

prioux:
Alternative solution being developed with Armin.

PR #666, opened 2017-11-22
new userfiles#file_collection_content action

ArminTaheri:
I have added a new simple `userfiles_controller` action, `file_collection_content`, to query files in a file collection using the rest of the route as a file path.

### DIFF
--- a/BrainPortal/app/controllers/userfiles_controller.rb
+++ b/BrainPortal/app/controllers/userfiles_controller.rb
@@ -188,6 +188,36 @@ class UserfilesController < ApplicationController
     redirect_to :action => :index
   end
 
+  # Use the rest of the route as a path into a file collection.
+  # [:file_path] a path relative to the file collection's route directory.
+  #
+  # GET /userfiles/1/file_collection_content/path/to/file/in/file_collection
+  def file_collection_content
+    @userfile = Userfile.find_accessible_by_user(params[:id], current_user, :access_requested => :read)
+    if @userfile.nil?
+      raise ActiveRecord::RecordNotFound("Could not retrieve a userfile with ID: #{params[:id]}")
+    end
+    path = @userfile.cache_full_path.to_s + '/' + params[:file_path]
+    if params[:format]
+      path += '.' + params[:format]
+    end
+    if not File.file?(path)
+      head :not_found
+      return
+    end
+    if File.symlink?(path)
+      sympath = File.expand_path(File.readlink(path))
+      if not sympath.to_s.start_with? @userfile.cache_full_path.to_s
+        head :unauthorized
+        return
+      end
+    end
+    if params[:format] == 'html'
+      render file: path, layout: false
+    else
+      send_file path
+    end
+  end
   # Transfer contents of a file.
   # If no relevant parameters are given, the controller
   # will simply attempt to send the entire file.
--- a/BrainPortal/config/routes.rb
+++ b/BrainPortal/config/routes.rb
@@ -110,6 +110,7 @@ Rails.application.routes.draw do
   resources :userfiles,       :except => [ :edit, :destroy ] do
     member do
       get  'content'
+      get  'file_collection_content/*file_path' => 'userfiles#file_collection_content'
       get  'display'
       post 'extract_from_collection'
     end
@@ -206,4 +207,3 @@ Rails.application.routes.draw do
   #get   '/service/provenance',     :controller => :service, :action => :provenance
 
 end
-


BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -188,6 +188,19 @@ def create_parent_child #:nodoc:
     redirect_to :action => :index
   end
 
+  # Use the rest of the route as a path into a file collection.
+  # [:file_path] a path relative to the file collection's route directory.
+  #
+  # GET /userfiles/1/file_collection_content/path/to/file/in/file_collection
+  def file_collection_content
+    @userfile = Userfile.find_accessible_by_user(params[:id], current_user, :access_requested => :read)
+    path = @userfile.cache_full_path.to_s + '/' + params[:file_path] + '.' + params[:format]
```
prioux: I think the extension should already be part of params[:file_path], so no need to append it from the request's format

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -188,6 +188,19 @@ def create_parent_child #:nodoc:
     redirect_to :action => :index
   end
 
+  # Use the rest of the route as a path into a file collection.
+  # [:file_path] a path relative to the file collection's route directory.
+  #
+  # GET /userfiles/1/file_collection_content/path/to/file/in/file_collection
+  def file_collection_content
+    @userfile = Userfile.find_accessible_by_user(params[:id], current_user, :access_requested => :read)
```
prioux: the `@userfile =` statement shoudl end with a `.first`

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -188,6 +188,19 @@ def create_parent_child #:nodoc:
...
+    path = @userfile.cache_full_path.to_s + '/' + params[:file_path] + '.' + params[:format]
+    if params[:format] == 'html'
```
prioux: I'd like more error checking.

* If the @userfile is nil, raise a ActiveRecord::RecordNotFound exception (the framework will handle the rest)
* If the file doesn't exist, properly return a 404 code (with no body)

BrainPortal/app/controllers/userfiles_controller.rb
ArminTaheri: I can do a check for
```
if @userfile.respond_to?(:first)
      @userfile = @userfile.first
end
```
but I should note I was going by this line: 
https://github.com/aces/cbrain/blob/master/BrainPortal/app/controllers/userfiles_controller.rb?utf8=%E2%9C%93#L203

`.first` does not seem to always be a method of the result of `Userfile.find_accessible_by_user`

ArminTaheri: I think you are perhaps mixing up `find_accessible_by_user` with `find_all_accessible_by_user`?

prioux: Yes you were right

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -188,6 +188,29 @@ def create_parent_child #:nodoc:
...
   end
 
+  # Use the rest of the route as a path into a file collection.
+  # [:file_path] a path relative to the file collection's route directory.
+  #
+  # GET /userfiles/1/file_collection_content/path/to/file/in/file_collection
+  def file_collection_content
+    @userfile = Userfile.find_accessible_by_user(params[:id], current_user, :access_requested => :read)
+    if @userfile.nil?
+      raise ActiveRecord::RecordNotFound
```
prioux: Add a nice message to the exception? `ActiveRecord::RecordNotFound.new('Message')`

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -188,6 +188,29 @@ def create_parent_child #:nodoc:
...
+    end
+    path = @userfile.cache_full_path.to_s + '/' + params[:file_path]
+    if params[:format]
+      path += '.' + params[:format]
+    end
+    if not File.file?(path)
```
prioux: I think this will fail (maybe) if the file is a symbolic link within the collection (check the doc for `File.file?()`

prioux: Also, make a check that the file path doesn't go outside the collection (there are methods in Pathname to help make sure the final destination of a path or symlink ends up within a certain directory tree)

ArminTaheri: it should be fine: https://ruby-doc.org/core-2.2.0/File.html#method-c-file-3F

BrainPortal/app/controllers/userfiles_controller.rb
ArminTaheri: oops!

BrainPortal/app/controllers/userfiles_controller.rb
ArminTaheri: ah good point!

prioux: In case of violation, maybe `head :unauthorized`

ArminTaheri: I have made the changes! I expand the path from read_link and check if it starts with the absolute path of the userfile.

ArminTaheri:
Sorry for the delay! I've rebased to the latest master with the new R5 version of cbrain

PR #676, opened 2017-12-11
remove code repetition (DRY)

ArminTaheri:
Just wondering why code was being repeated. Please close if this is not the correct logic after all.


ArminTaheri:
helped by @natacha-beck

prioux:
We have to be very carefule there, sometimes the change is subtle and it's not a repetition; but then I also agree it's entirely possible it is.

ArminTaheri:
Agreed but it's the same exact code with the difference only being `.find(id)`

prioux:
I'll make the modification in the R5 codebase instead. Thanks for reporting it. You'll be credited in the commit message.

PR #677, opened 2017-12-14
Misc small fixes post 5.0.0 release.

prioux:
* Bourreau boot process adjustments
* Boutiques tasks can work on files with only read access
* Fixed `file_collection_content` action in userfiles_controller

### DIFF
--- a/Bourreau/script/cbrain_remote_ctl
+++ b/Bourreau/script/cbrain_remote_ctl
@@ -112,7 +112,8 @@ def tunnel_to_secondary_host(host, http_port, db_port, args)
   end
   myscript = __FILE__
   myscript = (Pathname.getwd + myscript).to_s if myscript !~ /^\//
-  proxymaster.remote_shell_command_writer("ruby #{myscript.bash_escape} #{args.map(&:bash_escape).join(" ")} 2>&1" ) do |fh|
+  mydir = Pathname.getwd
+  proxymaster.remote_shell_command_writer("cd #{mydir.to_s.bash_escape}; bundle exec ruby #{myscript.bash_escape} #{args.map(&:bash_escape).join(" ")} 2>&1" ) do |fh|
     fh.write(STDIN.read) unless args.size == 1 && args[0] == 'stop'
   end
   if args.size == 1 && args[0] == 'stop'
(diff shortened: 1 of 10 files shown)


Bourreau/script/cbrain_remote_ctl
```diff
@@ -234,20 +235,25 @@ File.rename(environment_log_file, environment_log_file.sub(/\.log$/,"_#{date_ext
...
+          #"puma",
           "-d", # the -d will start it in background
           "-p", "#{port}",
           "-e", "#{environment}",
           "-b", "127.0.0.1",
-          "-P", "#{pidfile}"
+          "-P", "#{pidfile}",
+          #"--pidfile", "#{pidfile}",
+          #"--redirect-stdout", "log/stdout.tmp",
+          #"--redirect-stderr", "log/stderr.tmp",
```
natacha-beck: Why did you keep comments?

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
```diff
@@ -491,18 +495,26 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
...
-         .map    { |f| [f, Userfile.find_accessible_by_user(self.params[f], self.user)] }
+         .map    { |f| [f, Userfile.find_accessible_by_user(self.params[f], self.user, :access_requested => access)] }
          .select { |f| f[1].name.end_with?('.cbcsv') || f[1].suggested_file_type == CbrainFileList }
   end
 
   # Final set of tasks to be launched based on this task's parameters. Only
   # useful if the parameters set for this task represent a set of tasks
   # instead of just one.
   def final_task_list #:nodoc:
+    # This determines if the program expect to only read its input files,
```
natacha-beck: Maybe better to avoid the repetition. In case we change the logic in the future.

shots47s:
I have no additional comments.

shots47s:
I have no additional comments.

Bourreau/script/cbrain_remote_ctl
prioux: There are two ways to launch the server: `rails server puma [options]` and `puma [options]`. Normally the Bourreau is launched with the first way. While I was debugging the GPC problem I tried the second way too, which has slightly different options too. Since in the future maybe we'll want to launch puma directly, I left it there as a form of documentation.

prioux: Note: the `options` sets are incompatible,

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
prioux: Yes, I hate the fact that I had to repeat this several times. I can refactor into a method though, and the documentation would go int here. In fact, I'll do that.

PR #687, opened 2018-01-16
Gitbook documentation is now part of the cbrain repository

ArminTaheri:
we can now deploy documentation to `aces.github.io/cbrain`.

I believe the workflow is:
1. Developer updates docs and makes a PR
1. A judgement is made
1. An authority with repository access will then run
```
npm run build:docs
npm run deploy:docs
```

Note I've only tested this on my [fork](https://armintaheri.github.io/cbrain/).
This should not change anything though.

This is exactly what @natacha-beck had as a gitbook but moved into the cbrain repo.


docs/gitbook/0-overview/README.md
```diff
@@ -0,0 +1,60 @@
...
+The [**CBRAIN service**](https://portal.cbrain.mcgill.ca) was originally conceived
+through a [CANARIE](http://www.canarie.ca)
+grant awarded to [Professor Alan C. Evans](http://mcin-cnim.ca/people/alans-cv/), at
+[McGill University](http://www.mcgill.ca). The resulting internal
+code platform was meant
+to provide easy access to complex neuroimaging computational tools, for
+clinicians or neuroscience researchers with limited IT resources. Despite
+the name, the framework is designed to be general, and can accommodate any
+data and task for any application. This GIT repository contains the
+generic core of the framework, _which is not specific to any field of science_.
```
driusan: Skimming through this as markdown is kind of awkward because of inconsistent line length formatting. You can get more consistent formatting if you pipe the paragraphs through `fmt` (but fmt tends to eat up markdown lists in my experience, so I would only do it on the paragraphs..)

natacha-beck:
It is in now in a separate repository. We want to have it separate from the code base.

PR #690, opened 2018-01-18
Fix js implicite conversion of data value

natacha-beck:


prioux:
What about all the other JS files ?

prioux:
OK thin it covers all

natacha-beck:
I have `grep` on the whole code. It should be ok.

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -303,7 +303,7 @@
           if (!error_message) {
             error_message = "<span class='loading_message'>Error loading element</span>";
           }
-          if (replace === "true") {
+          if (replace === true) {
```
prioux: should we change this to

```
if (replace) {
```

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -324,7 +324,7 @@
       var scroll_bottom   = current_element.data("scroll-bottom");
 
       if (data) data = jQuery.parseJSON(data);
-      if (scroll_bottom === "false") scroll_bottom = false;
+      if (scroll_bottom === false) scroll_bottom = false;
```
prioux: What's this junk?

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -772,7 +772,7 @@
 
       data[submit_name] = submit_value;
 
-      if (ajax_submit !== "false") {
+      if (ajax_submit !== false) {
```
prioux: The same here,

```
if (ajax_submit) {
```

prioux:
Nat, you've tried it right, and the `replace` and `invert` tests work properly ? If so I will merge.

prioux:
I attempted a squash merge with Natacha's permission, just to see what would happen.

PR #695, opened 2018-01-25
First pass at making the API single token based.

prioux:
* No longer need to GET /session/new
* a cbrain_api_token is returned by POST /session
* this token need to be sent as a param for all other requests
* no cookie jar needs to be maintained by the client!
* that means no persistent filters or session prefs either

Swagger spec will be updated shortly.

Some unrelated changes:

* If a subtask asks to share the workdir with task #0, it means it wants to share with the current task
* Credits updated

### DIFF
--- a/BrainPortal/app/controllers/sessions_controller.rb
+++ b/BrainPortal/app/controllers/sessions_controller.rb
@@ -34,7 +34,8 @@ class SessionsController < ApplicationController
 
   api_available :only => [ :new, :show, :create, :destroy ]
 
-  before_action :user_already_logged_in, :only => [:new, :create]
+  before_action      :user_already_logged_in,    :only => [:new, :create]
+  skip_before_action :verify_authenticity_token, :only => [ :create ] # we invoke it ourselves in create()
 
   def new #:nodoc:
     reqenv           = request.env
@@ -45,16 +46,15 @@ class SessionsController < ApplicationController
 
     respond_to do |format|
       format.html
-      format.xml  { render :xml  => { :authenticity_token => form_authenticity_token } }
-      format.json { render :json => { :authenticity_token => form_authenticity_token } }
-      format.txt
+      format.any { head :unauthorized }
     end
   end
 
   def create #:nodoc:
-    # If the csrf token is blank, it was reset by Rails' request forgery protection
-    # and we want to prevent login
-    user = session["_csrf_token"].presence && User.authenticate(params[:login], params[:password]) # can be nil if it fails
+    if request.format.to_sym !~ /^(json|xml)$/i # JSON is used for API calls; XML not yet fully supported
+      verify_authenticity_token  # from Rails; will raise exception if not present.
+    end
+    user = User.authenticate(params[:login], params[:password]) # can be nil if it fails
     create_from_user(user)
   end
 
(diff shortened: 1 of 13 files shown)


prioux:
Note: I suspect we will still have problems with the CSRF token in POSTs, unless Rails is intelligent enough not to check for JSON requests.

BrainPortal/lib/authenticated_system.rb
```diff
@@ -126,11 +140,25 @@ def redirect_back_or_default(default)
...
+    end
+
+    # For API calls. A +cbrain_api_token+ is expected in the params.
+    # If the token is valid, the cbrain_session object will contain
+    # a proper LargeSessionInfo object from which the token's user
+    # can be found.
+    def user_from_api_token
+      return nil unless @cbrain_api_token
+      user = cbrain_session.user_for_cbrain_api(@cbrain_api_token)
+      #@cbrain_api_token = nil unless user # invalid, just reset it
```
natacha-beck: Why did you keep this comment?

prioux:
@shots47s It's even better than that, I've started reviewing the swagger specs, and decided to go with version3, which is OpenAPI. I have the YAMl document opened in an editor but I've only managed to document properly the sessions actions at this point.

BrainPortal/lib/authenticated_system.rb
```diff
@@ -4,22 +4,36 @@ module AuthenticatedSystem #:nodoc:
...
       !!current_user
     end
 
-    # Accesses the current user from the session.
-    # Future calls avoid the database because nil is not equal to false.
+    # Returns the currently logged-in user, if any.
     def current_user
-      @current_user ||= login_from_session unless @current_user == false
+      return @current_user if @current_user.present?
+      return nil if @current_user == false # false means we previously explicitely set current_user=(nil)
```
natacha-beck: I did not understand the comment. Can you explain?

BrainPortal/app/controllers/sessions_controller.rb
```diff
@@ -169,7 +173,7 @@ def create_from_user(user) #:nodoc:
   # when a user has not properly authenticated
   def auth_failed
     respond_to do |format|
-      format.html { render :action => 'new' }
+      format.html { render :action => 'new', :status => :ok } # should it be :unauthorized ?
```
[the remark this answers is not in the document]
prioux: I'm not sure. We're talking about a browser here: if the user enters the wrong password we still return a proper HTML form to try again, which is 200. Havign a form content return withe 40x int he header might not be supported with all browsers...

BrainPortal/lib/authenticated_system.rb
prioux: This method, surprisingly, is the main entry point to test whether or not a user is logged in. Tarek had created it such that when first invoked, it will try to identify the user ID from the session information. Initially the internal variable `@current_user` is set to nil. If it could find user from the session, it stored and cached it in`@current_user` to make further lookups fast. However, if it couldn't then it sets the internal variable to `false` as a way to both indicate there is no user, but also further calls to the method will not try again to resolve the user.

I kept the same semantics, except now we have two ways to find out which user is logged in: by the API token, or by the `session[:user_id]` (as it was).

prioux: It's annoying that this system is full of side-effects (the false is assigned in `current_user=()` when passed nil in argument.

BrainPortal/lib/authenticated_system.rb
prioux: Because I am not yet sure if, for the general behavior of CBRAIN, I want to zap the token if it's invalid, or keep it around.

natacha-beck:
@shots47s  and @gkiar for me everything is ok, I have tested this one. Can one of you try it before the merge?

shots47s:
I am dusting off what I did and will test shortly.

PR #696, opened 2018-01-30
Switching Singularity 'run' to 'exec'

shots47s:
Currently, CBRAIN executes 'singularity run' for performing a pipeline on a singularity image, which invokes the containers 'runscript'.  We really want 'exec' which allows us to execute commands directly in the container.  This change fixes.  It also becomes necessary from the singularity commands perspective to specify what is to be executed with the path, so I placed './' before the sh script to ensure that the command can find it.

Changed singularity command to:
1) Use exec instead of run
2) Specify that the .singularity.sh script is specified as with './' in front of it so that singularity command can find and execute.


shots47s:
Had a slack conversation with @prioux, and am going to merge.

Issue #697, opened 2018-01-31
Limit queries for API calls


prioux:
API calls to index actions are not limited by default, Introduce a default limit of 1000 entries.

Tasks:

1) implement .for_api() on AR::Relation, which delegates to the array method
2) this .for_api() wil first inspect the Relation to detect a limit(); if none, it will intoduce one (make it configurable on a model by model basis)
3) all controller code that has something like `relation.to_a.for_api` must be changed to `relation.for_api`

prioux:
Fixed by #700

PR #700, opened 2018-02-01
API limits

prioux:
API queries are now limited by default to 1000 results.

Each model can specify a limit with the class method `default_api_limit()` ; the method is defined in **ApplicationRecord**, which is a superclass of all **ActiveRecords**. Right now it's the one returning 1000.

The **ActiveRecord::Relation** class has been extended to gain its own `for_api()` which will check for a limit, fetch that value of not, apply it, and delegate to the `for_api()` method of **Array**.

While all this was being coded, it made sens to also tackle a refactoring issue: putting all the common code added by CBRAIN to **ActiveRecord::Base** into the new Rails5 **ApplicationRecord** instead. So this PR includes this too.

KNOWN LIMITATIONS:

None of the controllers right now implement the pagination conventions for API requests (these are normally supplied as the two parameters _per_page_ and _page_ ). This means basically that an API user cannot fetch let than the first 1000 records, and cannot get any records beyond these 1000. Oops. Will create an issue for that.

### DIFF
--- a/BrainPortal/app/controllers/sessions_controller.rb
+++ b/BrainPortal/app/controllers/sessions_controller.rb
@@ -51,7 +51,7 @@ class SessionsController < ApplicationController
   end
 
   def create #:nodoc:
-    if request.format.to_sym !~ /^(json|xml)$/i # JSON is used for API calls; XML not yet fully supported
+    if ! api_request? # JSON is used for API calls; XML not yet fully supported
       verify_authenticity_token  # from Rails; will raise exception if not present.
     end
     user = User.authenticate(params[:login], params[:password]) # can be nil if it fails
@@ -147,7 +147,7 @@ class SessionsController < ApplicationController
     end
 
     self.current_user = user
-    session[:user_id] = user.id if request.format.to_sym == :html
+    session[:user_id] = user.id  if request.format.to_sym == :html
     portal = BrainPortal.current_resource
 
     # Check if the user or the portal is locked
@@ -227,6 +227,13 @@ class SessionsController < ApplicationController
     # The authentication_mechanism is a string which describes
     # the mechanism that was used by the user to log in.
     authentication_mechanism = "password" # in future this could change
+
+    # In case of logins though the API, record that in the session too.
+    if api_request?
+      cbrain_session[:api] = 'yes'
+      authentication_mechanism = 'password/api'
+    end
+
     user.addlog("Logged in with #{authentication_mechanism} from #{pretty_host} using #{pretty_brow}")
     portal.addlog("User #{user.login} logged in with #{authentication_mechanism} from #{pretty_host} using #{pretty_brow}")
     user.update_attribute(:last_connected_at, Time.now)
--- a/BrainPortal/spec/models/data_provider_spec.rb
+++ b/BrainPortal/spec/models/data_provider_spec.rb
@@ -145,9 +145,6 @@ describe DataProvider do
       provider.not_syncable = true
       expect{provider.sync_to_cache(userfile)}.to raise_error(CbrainError, "Error: provider #{provider.name} is not syncable.")
     end
-    it "should raise an exception if sync_to_cache is called" do
-      expect{provider.sync_to_cache(userfile)}.to raise_error("Error: method not yet implemented in subclass.")
-    end
   end
 
   describe "#sync_to_provider" do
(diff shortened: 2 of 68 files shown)


prioux:
This will fix #697

glatard:
I feel like pagination might not be very relevant in the API. Instead, the API could specify the max number of results it wants to get. It should be possible to get all the results through the API. My two cents.

natacha-beck:
@prioux can we remove the pagination limitation if the format is not `HTML`. For me, we have no reason to put a limitation on other formats than HTML.

prioux:
We MUST have limits and pagination. There are models where we have way too many objects. Natacha, didn't you try the Ruby statements I sent you on Slack last week? It was:

```ruby
puts Time.now
a=Userfile.all.to_a
puts Time.now
b=a.for_api
puts Time.now
c=b.to_json
puts Time.now
puts "#{c.size} bytes"
```

This basically reproduces when a naive API user would trigger with a plain "GET /userfiles"

natacha-beck:
Ok... here is the output: 

> 2018-02-06 11:37:17 -0500
>  Userfile Load (1174.5ms)  SELECT `userfiles`.* FROM `userfiles`
> 2018-02-06 11:40:06 -0500
> 2018-02-06 11:44:22 -0500
> 2018-02-06 11:48:30 -0500
> 96931116 bytes

We are better to have limits.

BrainPortal/app/controllers/sessions_controller.rb
```diff
@@ -147,7 +147,7 @@ def create_from_user(user) #:nodoc:
     end
 
     self.current_user = user
-    session[:user_id] = user.id if request.format.to_sym == :html
+    session[:user_id] = user.id  if request.format.to_sym == :html
```
shots47s: extra space needed?

BrainPortal/app/controllers/sessions_controller.rb
```diff
@@ -51,7 +51,7 @@ def new #:nodoc:
   end
 
   def create #:nodoc:
-    if request.format.to_sym !~ /^(json|xml)$/i # JSON is used for API calls; XML not yet fully supported
+    if ! api_request? # JSON is used for API calls; XML not yet fully supported
```
shots47s: Is this comment still true, because it would api_request? will not screen agains it being XML or JSON... right?

prioux:
I am joining together two pull requests, because they depend on each other too. So this PR now also contains the following features:

* `_simple_filters=1` are now enabled by default without having to explicitly pass this param as a flag, whenever a request is made by an API call
* speaking of simple filters, unlike Remi's code, I validate which attributes are valid to add to the filters by asking the ActiveRecord for its official attribute list
* the LargeSessionObject encapsulated by CbrainSession will get its top entry `scopes` deleted completely at the end of the request, so that no scopes are persistent
* logging in with the API now puts a `api=true` key/value inside the LargeSessionObject
* all controllers now have access to the predicate `api_request?`
* Remi's scope subsystem used to default to naming scopes using the convention 'controller#action' (e.g. 'userfiles#index'), but the view code and controller code were most of the time supplying their own names with just the controller name (e.g. 'userfiles'). I adjusted all the code so that the default name is used from now on.
* api queries for userfiles index will default to the option `view_all=true` always; if users want their own files they'll have to supply the param `user_id=nnn`
* a Rails feature called 'wrap_parameters' has been disabled

BrainPortal/app/controllers/sessions_controller.rb
prioux: It's a reminder that there are controller actions where we properly return json results, but not in XML... I have tried as much as possible to provide both whenever I noticed the XML support was lacking.

BrainPortal/app/controllers/sessions_controller.rb
prioux: Yep, it's a mistake from me.

prioux: I noticed it too after I pushed the code.

BrainPortal/app/controllers/sessions_controller.rb
shots47s: Thanks.

prioux:
The KNOWN LIMITATIONS of the first pull request is no longer a problem.

NEW KNOWN LIMITATIONS:

* The userfiles controller code impose, by default, a filter where "hidden = false" to all request. There is no way to turn that off in the API
* Providing special filers or custom filter options at the same time as the simple filters won't work: it's one or the other, it's a limitation of Remi's scope update before_action method.

BrainPortal/spec/models/data_provider_spec.rb
```diff
@@ -145,9 +145,6 @@
       provider.not_syncable = true
       expect{provider.sync_to_cache(userfile)}.to raise_error(CbrainError, "Error: provider #{provider.name} is not syncable.")
     end
-    it "should raise an exception if sync_to_cache is called" do
```
natacha-beck: Just curious, why did this test was removed?

PR #702, opened 2018-02-02
force entrypoint override to `/bin/sh`

gkiar:
This enables containers with entrypoints other than a shell to work within cbrain. Similar to #696 but for Docker.

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2038,6 +2038,7 @@ chmod 755 #{docker_wrapper_basename.bash_escape}
     run                                          \\
     -u #{Process.uid}:#{Process.gid}             \\
     --rm                                         \\
+    --entrypoint /bin/sh                         \\
     -v "${PWD}":#{esc_cont_work_dir}             \\
     -v #{cache_dir.bash_escape}:#{cache_dir.bash_escape}         \\
     -v #{gridshare_dir.bash_escape}:#{gridshare_dir.bash_escape} \\


BrainPortal/app/models/cluster_task.rb
```diff
@@ -2038,6 +2038,7 @@ def docker_commands(command_script)
     run                                          \\
     -u #{Process.uid}:#{Process.gid}             \\
     --rm                                         \\
+    --entrypoint /bin/sh                         \\
```
prioux: Why `/bin/sh` instead of `/bin/bash` ?!?

prioux:
Please make your pull requests following the standard development policies:

1) from a distinct branch created for your feature in your fork
2) to the `dev` branch of aces

BrainPortal/app/models/cluster_task.rb
gkiar: Not all slimmed down linux containers have `/bin/bash`, all that I've found do have `/bin/sh`. We've had this discussion within Boutiques and was the conclusion we arrived at and has been successful.

prioux:
After discussion with Natacha this seems to be a 'hot patch'.

I'd liek to point out that this is not how CBRAIN development is supposed to happen: we're applying too often these hot patches that reveal that devs have not been testing their code locally  properly.

BrainPortal/app/models/cluster_task.rb
shots47s: alpine is an example, it is the canonical docker image for python programs and doesn't have /bin/bash available.  This is what precipitated this same change in boutiques.

prioux: OK. Anyway. /bin/sh is a buggy interpreter, very limited, it coudl cause serious problems if the script that you invoke was designed for /bin/bash ...

PR #705, opened 2018-02-09
Load tool_configs before singularity build.

natacha-beck:
In order to be able to use singularity-2.4.1 on some clusters, we need to be able to call the tool_config bash prologue before running the `singularity build` command. 

I fixed some comments in the meantime since we did not use anymore the pull command.  

I also add the fix for the permission in order to be able to mount directory by singularity. 

I made the PR on master after discussion with @prioux since both issues are essential for singularity usage.

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2180,6 +2180,9 @@ SINGULARITYJOB
 
 # Make sure it is executable
 chmod 755 #{singularity_wrapper_basename.bash_escape}
+# Other should have executable right on all components
+# of the path in order to be mounted by singularity.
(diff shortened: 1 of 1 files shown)


BrainPortal/app/models/cluster_task.rb
```diff
@@ -2180,6 +2180,9 @@ def singularity_commands(command_script)
 
 # Make sure it is executable
 chmod 755 #{singularity_wrapper_basename.bash_escape}
+# Other should have executable right on all components
+# of the path in order to be mounted by singularity.
+chmod o+x . .. ../.. ../../..
```
shots47s: This seems scary to me if it is not put in the absolute right place.  Just flagging to make sure you determined it wasn't so scary.

PR #708, opened 2018-02-17
Last updated indicator refactored. Fix #674

### LINKED ISSUE #674: New content timer should not bind a function to all elements
In the layout file `_section_account.html.erb` there's a piece of javascript code that binds a new_content event o each and every element of the DOM.

1) Move this out to a proper .js file
2) Instead of binding to everything, we only need to trigger the timer reset code
   to `ajax:complete` events (or even `ajax:start`).

prioux:
No longer binds a event handler to each and every element in the page!


prioux:
Will fix #674 .

natacha-beck:
Tested on my side, works properly.

PR #712, opened 2018-02-19
Better tool config index table. Fix #633

### LINKED ISSUE #633: ToolConfig index table should provide links, also new columns
- [x] Make execution server names links to their 'show' page.
- [x] Make the tool names link to their 'show' page.
- [x] Make the tool config names link to their 'show' page (just like the "Show" link already present on the right side of the table)
- [x] Add column with the group_id of the tool just after the tool name (and values linking to the groups)
- [x] Rename the column 'project access' to be 'Version Project", make entries be links to the groups
- [x] Add a column that links to the access report for that tool on that execution server

prioux:
Will fix #633


natacha-beck:
Tested, everything works fine.

PR #713, opened 2018-02-19
Refresh button in task show page near status. Fixes #688

### LINKED ISSUE #688: Tasks page to refresh itself of to have a refresh button on the page
For someone not familiar with the tool, it would easy to think nothing is happening, so we can put an autorefresh to the page or actually put a button that says "Update Status" on the page so that it is clear that something needs to be done to see the status change.

prioux:
You might find it ugly? This will fix #688


natacha-beck:
Tested it works fine but is not the best UI I never saw. 

I personally will prefer to have this button at the top of the show page will all the other button (`Terminate Task`, `Remove Task`).

prioux:
Will rebase and adjust.

PR #715, opened 2018-02-20
Plugins can provide a 'lib' directory. Fix #714

### LINKED ISSUE #714: Plugins to provide new Ruby code with a 'lib' directory.
Plugins could contain a "lib" directory with ruby code that could be integrated in the main CBRAIN framework.

- [x] The rake task would symlink the files into `installed-plugins/lib`
- [x] That directory woudl be added to the Rails app's load path

prioux:
Any CBRAIN plugins can now contain a 'lib' directory with
ruby files in them (*.rb), these will be symlinked by
the rake installation task and loaded by Rails.

### DIFF
--- a/BrainPortal/cbrain_plugins/installed-plugins/lib/.gitignore
+++ b/BrainPortal/cbrain_plugins/installed-plugins/lib/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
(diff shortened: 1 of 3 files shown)


natacha-beck:
Just tested with the new cbrain-plugins-loris. It works perfectly!

Issue #675, opened 2017-12-05
Prevent archiving of immutable file collections


prioux:
Add this at the top of `provider_archive()`

```ruby
cb_error "Error: file #{self.name} is immutable." if self.immutable?
```

prioux:
Maybe make sure the compress and uncompress methods for singlefiles also respect the immutable flag.

prioux:
Fixed by #716

PR #716, opened 2018-02-20
Immutable files can't be compressed or uncompressed.

prioux:


prioux:
This will fix #675

prioux:
Note to reviewers: you get a much more readable diff by using `?w=1` in GitHub.

PR #720, opened 2018-02-21
Ignore error when building singularity image

### LINKED ISSUE #719: Only check for singularity image existance when build.
To avoid some unexpected behavior after building a singularity image, it is better to only check for the presence of the singularity image after the `singularity build...` command. 
Only if the image is not present or is empty CBRAIN should raise an error.

natacha-beck:
Fixes #719

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2406,15 +2406,12 @@ chmod o+x . .. ../.. ../../..
       # of these blocks can be scheduled to run, but only one will execute at at any given time.
       next if File.exists?(cache_path.to_s) && File.size(cache_path.to_s) > 0
       # Run singularity build command
-      out,err = tool_config_system("umask 002; #{singularity_executable_name} build #{cache_path.to_s.bash_escape} #{singularity_index_location.bash_escape}#{singularity_image_name.bash_escape}")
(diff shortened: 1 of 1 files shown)


BrainPortal/app/models/cluster_task.rb
```diff
@@ -2406,15 +2406,12 @@ def load_singularity_image_from_repo #:nodoc:
...
+      tool_config_system("umask 002; #{singularity_executable_name} build #{cache_path.to_s.bash_escape} #{singularity_index_location.bash_escape}#{singularity_image_name.bash_escape}")
       # Singularity command can generate 'implausibly old time stamp' when pulling a docker image (due to tar), we ignore it.
       # Not sure if it is still true for `build` command, leave it just in case.
       # Remove all lines (use ^ and $) that contains this message.
-      err.gsub!(/^.*implausibly old time stamp.*$\n?/,"")
-      err.gsub!(/^.*WARNING:.+/, "")
       cb_error "Cannot build singularity image" if
-        err.present? ||
-        ! File.exists?(cache_path.to_s)
+        File.zero?(cache_path.to_s)
```
prioux: Bug: `File.zero?('does not exist')` returns false

prioux: `File.size?()` works better, it tests existence and size.

Issue #673, opened 2017-12-05
Easy way to access workdir.


natacha-beck:
When a task failed, is not easy to access the workdir in order to inspect the result: 

Accessing task work directories is possible but bulky: 
- needs to archive the task
- download
- then unarchive.

Should bring an easy way to do that.

shots47s:
Need to find all of the other issues that pertain to this an link them here @shots47s

natacha-beck:
This PR https://github.com/aces/cbrain/pull/721 allow saving the WorkDirectory as a TaskRawDirectory directly from the task pages. If it works with all of you (@shots47s , @prioux , @gkiar ) I will close this issue.

prioux:
Done by Natacha in #721

PR #721, opened 2018-02-25
Save Work Directory as FileCollection

natacha-beck:
My solution to the issue #673. 

On task show page the user will see a button `Save Work Directory`. 
It will allow the user to save the task Work directory as a FileCollection. 

*Note*: As a convention in CBRAIN the file viewer never shows files than begin by `.` or `..` so every single file present in the task work directory will not appear in the show page of the FileCollection even if they are present. 

@prioux I normally fixed what we discuss on Friday except for the `Message`. I do not see how I can send a message to the user. If I well understand I should use the `Message.send_message` method but I do not have access to the current_user in the `save_cluster_workdir` method. Any suggestion?

### DIFF
--- a/BrainPortal/app/models/cbrain_task.rb
+++ b/BrainPortal/app/models/cbrain_task.rb
@@ -371,14 +371,14 @@ class CbrainTask < ApplicationRecord
 
   # Returns an ID string containing both the bourreau_id +bid+
   # and the task ID +tid+ in format "bid/tid". Example:
-  #     "3/4"   # Bourreau #3, task #4
+  #     "3/4"   # Bourreau ID 3, task ID 4
   def bid_tid
     @bid_tid ||= "#{self.bourreau_id || '?'}/#{self.id || '?'}"
   end
 
   # Returns an ID string containing both the bourreau_name +bname+
   # and the task ID +tid+ in format "bname/tid". Example:
-  #     "Mammouth/4"   # Bourreau 'Mammouth', task #4
+  #     "Mammouth/4"   # Bourreau 'Mammouth', task ID 4
   def bname_tid
     @bname_tid ||= "#{self.bourreau.name || '?'}/#{self.id || '?'}"
   end
@@ -386,19 +386,24 @@ class CbrainTask < ApplicationRecord
   # Returns an ID string containing both the bourreau_name +bname+
   # and the task ID +tid+ in format "bname-tid" ; this is suitable to
   # be used as part of a filename. Example:
-  #     "Mammouth-4"   # Bourreau 'Mammouth', task #4
+  #     "Mammouth-4"   # Bourreau 'Mammouth', task ID 4
   def bname_tid_dashed
     @bname_tid_dashed ||= "#{self.bourreau.name || 'Unk'}-#{self.id || 'Unk'}"
   end
 
   # Returns an ID string containing both the task +tname+
-  # and the task ID +tid+ in format "tname/tid". Example:
-  #     "Civet-1234"   # Task 'Civet' #1234
+  # and the task ID +tid+ in format "tname-tid". Example:
+  #     "Civet-1234"   # Task 'Civet' with ID 1234
   def tname_tid
     @tname_tid ||= "#{self.name || '?'}-#{self.id || '?'}"
   end
 
-
+  # Returns an ID string containing the task +tname+
+  # and the run_id in format "tname-rid". Example:
+  #     "Civet-1234-1"   # Task 'Civet' with ID 1234 and run number 1
+  def tname_rid
+    @tname_rid ||= "#{self.name || '?'}-#{self.run_id || '?'}"
+  end
 
   ##################################################################
   # Run Number ID Methods
--- a/BrainPortal/app/models/cbrain_task.rb
+++ b/BrainPortal/app/models/cbrain_task.rb
@@ -401,8 +401,8 @@ class CbrainTask < ApplicationRecord
   # Returns an ID string containing the task +tname+
   # and the run_id in format "tname-rid". Example:
   #     "Civet-1234-1"   # Task 'Civet' with ID 1234 and run number 1
-  def tname_rid
-    @tname_rid ||= "#{self.name || '?'}-#{self.run_id || '?'}"
+  def tname_run_id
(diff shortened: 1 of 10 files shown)


prioux:
For the notification, add a "notify_user_ids" attribute to the RemoteCommand model, and have these user IDs propagate to the `save_cluster_workdir` method.

prioux:
As discussed in person, the Bourreau model will need to get a method changed:

```ruby
# Before
def send_command_alter_tasks(tasks,new_task_status,new_bourreau_id=nil,archive_dp_id=nil)
end

# After
def send_command_alter_tasks(tasks,new_task_status,extra_remote_commands = {})
end
```

BrainPortal/app/models/cbrain_task.rb
```diff
@@ -371,34 +371,39 @@ def short_description
...
+  #     "Civet-1234"   # Task 'Civet' with ID 1234
   def tname_tid
     @tname_tid ||= "#{self.name || '?'}-#{self.id || '?'}"
   end
 
-
+  # Returns an ID string containing the task +tname+
+  # and the run_id in format "tname-rid". Example:
+  #     "Civet-1234-1"   # Task 'Civet' with ID 1234 and run number 1
+  def tname_rid
```
prioux: I would call the method `tname_run_id` just to stay consistent with the other helpers.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1935,6 +1935,67 @@ def remove_cluster_workdir
...
+        :data_provider_id => data_provider_id.to_i
+      )
+
+    file_collection.cache_copy_from_local_file(full_cluster_workdir)
+
+    if file_collection.save
+      Message.send_message(user,
+        :header        => "Saved task work directory",
+        :description   => "Task work directory was saved as a FileCollection",
+        :variable_text => "For [[#{self.pretty_type}][/tasks/#{self.id}]] under [[#{file_collection.name}][/userfiles/#{file_collection.id}]]",
```
prioux: Change message:

  `For task {self.fullname} in file {file_collection.name}`

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1935,6 +1935,67 @@ def remove_cluster_workdir
...
+        :type          => :notice,
+      )
+      self.addlog("Saved work directory in FileCollection #{userfile_name}.")
+    else
+      Message.send_message(user,
```
prioux: I would add the content of `self.errors.full_messages` to the variable_text, so we can know why.

BrainPortal/app/models/remote_command.rb
```diff
@@ -100,7 +100,11 @@ class RemoteCommand < RestrictedHash
 
     # -------- CHECK DATA PROVIDERS PARAMETERS --------
 
-    :data_provider_ids,    # in input for the command, an array of DP ids
+    :data_provider_ids,    # an input for the command, a list of DP ids as string. "1,2,3"
+
+    # -------- KEEP USER IDS --------
+
+    :notify_user_id,       # an input for the command generaly the current_user
```
prioux: Please make a comment with proper grammar and spelling: (`generally` has two Ls)

`an input for the command, generally the current user's ID`

BrainPortal/app/views/tasks/show.html.erb
```diff
@@ -96,12 +96,17 @@
...
                                                   :datatype => :script,
-                                                  :class => "button",
-                                                  :confirm =>
+                                                  :class    => "button",
+                                                  :confirm  =>
                   "Are you sure you want to remove the task's work directory?"
       %>
+      <%= link_to "Save Work Directory", url_for(:action    => :operation, :operation => 'save_wd', 'tasklist[]' => @task.id),
+                                                  :method   => :post,
+                                                  :datatype => :script,
```
prioux: I don't think the `:datatype` is necessary at all ? Can it be removed form all these buttons?

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1935,6 +1935,67 @@ def remove_cluster_workdir
...
+      return
+    end
+
+    if self.share_wd_tid
+      self.addlog "Cannot save work directory: this task doesn't have its own work directory"
+      return
+    end
+
+    # Save under a new FileCollection
+    userfile_name = self.tname_run_id
```
prioux: Just using `tname_run_id` for the output name is too dangerous. I can imagine a task programmer choosing this for the name of the task's output, too. So your method here woudl crush the output.

I suggest renaming it:

```ruby
userfile_name = "TaskRawWorkdir-#{self.tname_run_id}"
```

prioux: Also, maybe it woudl be worth it create a userfile model in cbrain-plugins-base:

``` ruby
TaskRawWorkdir < FileCollection
```

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1935,6 +1935,68 @@ def remove_cluster_workdir
...
+    end
+
+    if self.share_wd_tid
+      self.addlog "Cannot save work directory: this task doesn't have its own work directory"
+      return
+    end
+
+    # Save under a new TaskRawWorkdir
+    userfile_name   = "TaskRawWorkdir-#{self.tname_run_id}"
+    file_collection = safe_userfile_find_or_new(TaskRawWorkdir,
```
prioux: The file collection will belong to the task's owner, not the person creating it.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1935,6 +1935,69 @@ def remove_cluster_workdir
...
+        :name             => userfile_name,
+        :data_provider_id => data_provider_id,
+        :user_id          => user_id
```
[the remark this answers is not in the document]
prioux: I checked the code, safe_userfile_find_or_new() will try to create a file with `user_id => nil` if user_id is nil, so it won't work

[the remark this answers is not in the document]
prioux: it should be

```ruby
:user_id => (user_id.presence || self.user_id)
```

Issue #670, opened 2017-11-30
CustomFilters data hash need updating for Rails 5.1


prioux:
In 5.1 the behavior of the 'data' hash will fail. See custom_filters_controller in create and update.

prioux:
I need to find out exactly how it will fail... I should have written this down. I think it was some sort of deprecation warning related to the permit() calls?

prioux:
It might be related to this:

```
DEPRECATION WARNING: Method to_a is deprecated and will be removed in Rails 5.1, as `ActionController::Parameters` no longer inherits from hash. Using this deprecated behavior exposes potential security problems. If you continue to use this method you may be creating a security vulnerability in your app that can be exploited. Instead, consider using one of these documented methods which are not deprecated: http://api.rubyonrails.org/v5.0.6/classes/ActionController/Parameters.html
```

prioux:
I now believe it's everything related to invoking hash-like methods with a Parameters object:

https://guides.rubyonrails.org/5_1_release_notes.html#action-pack-removals

We'll need to tweak controller code everywhere.

natacha-beck:
Tested with Rails 5.1.7 the PR #728 fix this issue.

Issue #628, opened 2017-05-15
Creating userfile custom filters that fail validation doesn't inform the user


prioux:
When creating a userfile custom filter, if it fails validation (like, with having illegal characters in the name) the popup form disappear and no information is shown to the user. The popup should stay right there, instead.

prioux:
I believe this has been fixed by my big refactoring in #817

Issue #446, opened 2016-08-02
New filters forms to appear in separate pages


prioux:
We have many 'new filter' forms that appear as overlays; we should make them nice separate pages. Right now I can't even create a new userfile filter because the overlap exceeds the browser's page size vertically.

Related to #397 and #411 .

crocodoyle:
This will be part of the Bootstrapification.

Issue #332, opened 2016-02-12
Scope mode 'replace' doesn't replace all filters, leaves CustomFilters in place.


prioux:
When a link is created on a page to point to some other index page with particular filters applied and `_scope_mode=replace` is in effect, the destination index page gets cleared of all standard column filters, but not the custom filters, which are still there.

E.g.

* apply a custom filter to the task index page
* go to a user's show page
* click on the link that sends to the set of task of that user
* the task index page shown properly filters by user ID, but the custom filter is still in effect.

remibernard:
This is due to custom filters being handled manually in the task & userfile controllers rather than via the Scope module; the helper generating the links does not consider them as they fall out of the Scope module's concerns.

To explicitly replace those filters as well, one needs to add explicitly something like `{ 'c' => { 'custom_filters' => [] } }` to the Scope hash used in creating the URL (maybe using the more generic `scope_link` helper instead of the more helpful one `scope_filter_link`).

As far as implementing the solution goes, the best way would be to rework custom filters into `Scope::Filter` subclasses, which would then naturally be handled by the Scope module. This solution is rather tedious to implement, however, as it requires (more or less) a complete refactor of custom filters. Another, simpler, solution would be to add another helper using `scope_link` which incorporates a controller check (similarly to how `default_filters_for` offers a nice set of defaults for `filter_values_for` and `scoped_filters_for`).

PR #728, opened 2018-03-15
Custom filter refactoring. 

natacha-beck:
This PR should fix a couple issues related to the Custom Filter. 

Here is a summary for each of them: 
- #670 and #309 -> use a whitelist of keys for the data hash of Custom Filter, not sure if the `data` and the `data=` method in `custom_filter.rb` should be protected (specification of #309). 
- #628 -> The Custom Filter is no more display in an overlay. So the validation error will be displayed properly now. 
- #446 -> open on a separate page for both Custom Filters. 

I was not able to reproduce this behavior locally #332. 

I think this PR cover all the related issue for the Custom Filter at this point.


natacha-beck:
@prioux  and @shots47s  I think I now have something stable with this PR. I will be happy if you took a look to this one and if you test it.

prioux:
POSTing a new filter with no name just discards the entire form instead of generating the validation messages and returning to the page

prioux:
No validations on the values seem to be performed: I used the devs tools in Firefox to edit the file type selection field and entered a file type 'xyz' and it accepted it and saved it int he database. Same with the SyncStatus. I entered a dummy tag ID and it accepted it even if the tag id didn't exist.

prioux:
I can create a filter by name that says "contains" and leave the field blank, which makes a filter that doesn't filter by name at all.

prioux:
The filter by name selection box says "Don't filter description" even though it's not about the description attribute, it's about the name attribute.

prioux:
If I create a filter that filters by a date range, then later on disable date filtering, the show page still show the message about the data range.

prioux:
When editing a userfile custom filter, the value for `child name like` is always filled with the value of `parent name like` by mistake

prioux:
Filtering by type shoudl say "by file type" not "file format"

prioux:
Filtering by exact size makes no sense, if I enter "19" it will only find files of exactly 19.000 kilobytes. We should disable this. Also the size units should be in specifyiable in other units than Kb.

prioux:
The filtering fields with selection boxes are inconsistent: some of them have "Don't filter by XYZ" as their top option, others just have blank space.

prioux:
`Archiving status` always reset to blank value between edits of the filter

prioux:
The data range panel is missing the 'Big Bang' option!

prioux:
Creating filters for tasks should allow the user to select bourreaux even if they are offline

prioux:
`Work Directory status` always reset to `don't filter` when re-editing a task filter

BrainPortal/app/views/custom_filters/_new.html.erb
```diff
@@ -25,23 +25,26 @@
 <% custom_filter ||= @custom_filter %>
 <% filter_type = custom_filter.class.name.underscore %>
 
-<h2>New Custom Filter</h2>
+<h2> <%= filter_type.capitalize.gsub("_"," ")%> </h2>
```
prioux: Why not `filter_type.humanize` instead of ugle gsub()

prioux: Also, we are missing the `New` keyword.

BrainPortal/app/views/custom_filters/_new.html.erb
```diff
@@ -25,23 +25,26 @@
...
 
-<div class="generalbox">
+<%= form_for @custom_filter, :as => :site, :url => { :action => "create" }, :datatype => "script" do |f| -%>
```
prioux: What is this? This is a form that edits a custom filter as a Site object?!?

BrainPortal/app/views/custom_filters/_new.html.erb
```diff
@@ -25,23 +25,26 @@
...
 
-  <%= ajax_form_for custom_filter, :as => :custom_filter, :url => {:controller  => "custom_filters", :action => "create" }, :datatype  => "script" do |f| %>
-    <%= hidden_field_tag :filter_class, filter_type %>
-    <%= error_messages_for(custom_filter, :object_name => "custom filter") %>
+  <div class="generalbox">
 
-    <p>
-    <%= f.label :name, "Filter name" %><br>
-    <%= f.text_field :name %>
+    <%= ajax_form_for custom_filter, :as => :custom_filter, :url => {:controller  => "custom_filters", :action => "create" }, :datatype  => "script" do |f| %>
```
prioux: And here you have another form embedded the first form? That's illegal

prioux: Please double-check the structure of the document with View Source because I found many errors...

BrainPortal/app/views/custom_filters/_new_task_custom_filter.html.erb
```diff
@@ -0,0 +1,69 @@
+
```
prioux: In all the view code for new custom filters and for editing custom filters, please use the attribute getters and setters that were generated with metaprogramming: that was the entire point of creating them, so that the form helpers use that as if they were standard attributes of the CustomFilter model.

BrainPortal/app/views/custom_filters/_userfile_custom_filter.html.erb
```diff
@@ -22,70 +22,100 @@
...
+
+   <% t.edit_cell(:data_parent_name_like, :header => "Parent name contains (lists children)") do %>
+    <%= text_field_tag "data[parent_name_like]", @custom_filter.data[:child_name_like] %>
+  <% end %>
+
+  <% t.edit_cell(:data_child_name_like, :header => "Child name contains (lists parents)") do %>
+    <%= text_field_tag "data[child_name_like]", @custom_filter.data[:parent_name_like] %>
+  <% end %>
+
+  <% t.edit_cell(:data_type, :header => "By file type", :content => (@custom_filter.data_type||[]).sort.join(", ")) do %>
```
prioux: Note: I needed to introduce a `.presence` after the getter method call to support cases with empty strings in the the data structure.

BrainPortal/app/controllers/custom_filters_controller.rb
```diff
@@ -51,43 +53,48 @@ def create #:nodoc:
...
-
-
     respond_to do |format|
-      format.xml  { render :xml => @custom_filter.errors, :status => :unprocessable_entity }
-      format.js
+      if @custom_filter.errors.empty?
+        flash[:notice] = "Custom filter '#{@custom_filter.name}' was successfully updated."
+        format.html { render :action => :show }
+      else
+        @custom_filter.reload
```
prioux: I'd remove the reload

BrainPortal/app/models/custom_filter.rb
```diff
@@ -51,26 +51,98 @@ class CustomFilter < ApplicationRecord
...
+    valid_status = self.is_a?(TaskCustomFilter) ? ["none", "cluster", "file"] : ["archived", "none"]
+    return true if valid_status.include? self.data_archiving_status
+    errors.add(:data_archiving_status, 'is not a valid archiving status')
+    return false
+  end
+
 
   # Do some validation on the date range filtering
-  def validate_date
+  def valid_data_date #:nodocs:
```
prioux: this method is still using the old way of accessing data: `self.data["date_attribute"]` instead of `self.data_date_attribute`

BrainPortal/app/models/userfile_custom_filter.rb
```diff
@@ -102,45 +178,45 @@ def scope_name(scope)
...
+    scope.where( :type => self.data_type )
   end
 
   # Return +scope+ modified to filter the Userfile entry's type.
   # This scope filters by all subclasses of the chosen class type.
   # Not used by interface yet.
   def scope_type_tree(scope)
     flatlist = []
-    Array(self.data["type"]).each do |klassname|
+    Array(self.data[:type]).each do |klassname|
```
prioux: I think we can use `self.data_type` here too?

natacha-beck:
Yeahh ! thanks for the review.

PR #730, opened 2018-03-16
Added tool show page and fixed a minor bug in scope

shots47s:


shots47s:
comments from Natacha's previous review: 

In your show page it miss the following information select_menu_text that is link to the following text Text for select box on the userfiles page in the edit page.

The show and the Edit page can become a single page as it is done for the majority of other models (it the case of the Site for example). In this way we can remove the edit page.

shots47s:
yes that is clearer, I will work on that.

BrainPortal/app/views/tools/show.html.erb
```diff
@@ -0,0 +1,139 @@
+
```
prioux: Please de-tab this file and remove trailing spaces, as per

https://github.com/aces/cbrain/wiki/Programming-Style

prioux:
This view code is not accessible at all from any page in CBRAIN; that is, there are no links anywhere to this the tool show page.

It is also a large duplication of other view files (tools/_form_fields.html.erb new.html.erb), so it's not DRY

BrainPortal/app/views/tools/show.html.erb
```diff
@@ -0,0 +1,139 @@
...
+		app_type = @tool.application_type.empty? ? "(Not Specified)" : @tool.application_type
+		t.edit_cell(:application_type, :header => "Application Type", :content => app_type) do %>
+		<%= text_field_tag "tool[application_type]", @tool.application_type %>
+		<% end
+	%>	
+
+	<%
+		app_tags = @tool.application_tags.empty? ? "(Not Specified)" : @tool.application_tags
+		t.edit_cell(:application_tags, :header => "Comma Separated Tags", :content => app_tags) do %>
+		<%= text_field_tag "tool[application_tags]", @tool.application_type %>
```
prioux: Bug here; it shows the application_type instead of application_tags

BrainPortal/app/views/tools/show.html.erb
```diff
@@ -0,0 +1,139 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.  
+#
+-%>
+
+<% title 'Tool Info' %>
+
+<div class="display_inline_block" style="min-width:50%">
```
prioux: The entire page is missing two attributes

- `:cbrain_task_class_name`
- `:select_menu_text` (already pointed out by Natacha)

prioux:
Cancelled after discussion with the author.

PR #731, opened 2018-03-16
Updated the S3 Data provider to use aws-sdk-s3 API v3

shots47s:
This now operates more or less like a normal flat data provider but on an S3 Instance.

I can provide you with a private API key to try it out if you like.

### DIFF
--- a/BrainPortal/app/models/s3_data_provider.rb
+++ b/BrainPortal/app/models/s3_data_provider.rb
@@ -28,7 +28,8 @@ class S3DataProvider < DataProvider
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
 
-  validates_presence_of :cloud_storage_client_identifier, :cloud_storage_client_token
+  validates_presence_of :cloud_storage_client_identifier, :cloud_storage_client_token, 
+                        :cloud_storage_client_path_start, :cloud_storage_client_bucket_name
 
   attr_accessor :s3_connection
 
@@ -40,48 +41,27 @@ class S3DataProvider < DataProvider
   # Connects to the S3 service using :cloud_storage_client_identifier and :cloud_storage_client_token;
   # the connection is maintained in a instance variable!
   def init_connection
-    @s3_connection = S3Connection.new(self.cloud_storage_client_identifier, self.cloud_storage_client_token)
+    @s3_connection = S3Connection.new(self.cloud_storage_client_identifier, 
+                                      self.cloud_storage_client_token,
+                                      self.cloud_storage_client_bucket_name,
+                                      self.cloud_storage_client_path_start)
+    puts @s3_connection.connected?
   end
 
-  # Hardcoded bucket name is "gbrain_{self.name}"
+  # Get the bucket name for the Data Provider specified at creation
   def bucket_name
-    "gbrain_#{self.name}"
-  end
-
-  # Mapping between a userfile's name and the name
-  # of the S3 file
-  def s3_filename(userfile,newname=nil)
-    namekey = newname.presence || userfile.name
-    ext = userfile.is_a?(FileCollection) ? ".TGZ" : ""
-    "#{userfile.id}_#{namekey}#{ext}"
-  end
-
-  # Informational: "bucket/s3_filename"
-  def provider_full_path(userfile)
-    "#{bucket_name}/#{s3_filename(userfile)}"
-  end
-
-  # Mapping between S3 filename and CBRAIN's userfile ID and filename.
-  def filename_from_s3_filename(s3_filename)
-    userfile_id,filename=s3_filename.split('_', 2)
-    return [ userfile_id,filename ]
-  end
-
-  # Create the bucket on S3
-  def create_base_bucket
-    init_connection
-    @s3_connection.create_bucket(bucket_name)
+    @s3_connection.bucket_name
   end
 
   def impl_is_alive? #:nodoc:
     init_connection
     @s3_connection.connected?
-  rescue
-    false
   end
 
   def is_browsable?(by_user = nil) #:nodoc:
-    false
+    return true if by_user.blank? || self.meta[:browse_gid].blank?
+    return true if by_user.is_a?(AdminUser) || by_user.id == self.user_id
+    by_user.is_member_of_group(self.meta[:browse_gid].to_i)
   end
 
   def allow_file_owner_change? #:nodoc:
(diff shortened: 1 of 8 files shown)


natacha-beck:
I would like to have the API Key to try it.

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,264 @@ def self.pretty_category_name #:nodoc:
...
   def impl_is_alive? #:nodoc:
     init_connection
     @s3_connection.connected?
-  rescue
-    false
   end
 
   def is_browsable?(by_user = nil) #:nodoc:
-    false
+    return true if by_user.blank? || self.meta[:browse_gid].blank?
```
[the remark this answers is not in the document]
shots47s: This is patterned directly after ssh_data_provider.rb, so the question applies to both if I understand it.  @prioux should weigh in.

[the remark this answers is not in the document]
natacha-beck: Oh really... @prioux  can you give us more explanation of this choice.

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,264 @@ def self.pretty_category_name #:nodoc:
...
+    by_user.is_member_of_group(self.meta[:browse_gid].to_i)
   end
 
   def allow_file_owner_change? #:nodoc:
     true
   end
 
   def content_storage_shared_between_users? #:nodoc:
-    false
+    true
```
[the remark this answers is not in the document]
shots47s: I am not sure that I understand this parameter then, again patterned after ssh_data_provider.rb.  This is essentially a "flat" data provider.  Advice on what to set this to would be helpful.

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,264 @@ def self.pretty_category_name #:nodoc:
...
 
+  def provider_full_path(userfile)
+    init_connection
+    @s3_connection.clean_starting_folder_path(userfile.name)
+  end
+  
+  def impl_provider_collection_index(userfile, directory = :all, allowed_types = :regular) #:nodoc:
+    list = []
+       
+     if allowed_types.is_a? Array
```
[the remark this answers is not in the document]
shots47s: Sometimes, I prefer explicit form but will comply to this compactness.

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,252 @@
...
-    @bucket   = Bucket
-    @s3object = S3Object
-
-    @base.establish_connection!(
-      :access_key_id     => (access_key || "blah"),
-      :secret_access_key => (secret_key || "blah")
-    )
+  def initialize(access_key, secret_key, bucket_name, path_start, 
+                 region='us-east-2',
+                 endpoint="http://s3.us-east-2.amazonaws.com")
```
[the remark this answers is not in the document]
shots47s: Eventually, for now, no.  I put as a placeholder, for the future.

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,252 @@
...
+    @bucket = @resource.bucket(@bucket_name)
+    @path_name = path_start
+    @endpoint = endpoint
   end
-
+  
+  # Method to translate Mime types to File types
+  # Redo more generally
+  def translate_content_type_to_ftype(ct)
+    if ct == "application/x-directory"
```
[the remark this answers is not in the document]
shots47s: to be fair, every if statement can do that, I would say it is a bit nitpicky to say that everyone one that can be should be.  But I will conform to the suggestion.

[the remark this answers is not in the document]
natacha-beck: Conform only when you feel it worst it. It was more a general comment and I do not put it everywhere.

BrainPortal/app/models/s3_data_provider.rb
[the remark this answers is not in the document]
natacha-beck: Seem fine I was just disturbed by the fact it changes from the previous version of the s3_dataprovider. I will let  @prioux  to take a second look to this point.

shots47s:
Made changes to address @natacha-beck comments to the review

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,243 @@ def self.pretty_category_name #:nodoc:
...
-  #  s3_connection.bucket.find(bucket_name).objects.map do |object|
-  #    file               = DataProvider::FileInfo.new()
-  #    filename           = filename_from_s3_filename(object.key)[1]
-  #    file.name          = filename
-  #    file.symbolic_type = :regular
-  #    file.mtime         = Time.parse(object.about()["date"]).to_i
-  #    file.size          = 0
-  #    file
-  #  end
+    if userfile.is_a?(FileCollection)
```
prioux: This entire code block can be rewritten as

```ruby
  if userfile.is_a?(FileCollection)
    @s3_connection.rename_path(oldpath,newpath)
  else
    @s3_connection.rename_object(oldpath,newpath)
  fi
  true
rescue
  false
```

where the rescue clause is at the method level

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,243 @@ def self.pretty_category_name #:nodoc:
...
+      end
+    else
+      entry = @s3_connection.get_object_stats(provider_full_path(userfile))
+      entry[:name] = userfile.name
+      entries << entry
+    end
+     
+    entries.each do |entry|
+      next if entry[:content_type] == "application/x-directory"
+      type = @s3_connection.translate_content_type_to_ftype(entry[:content_type])
```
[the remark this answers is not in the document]
prioux: I can't find the doc for translate_content_type_to_ftype() anywhere on the web?

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,243 @@ def self.pretty_category_name #:nodoc:
...
 
-end
+  def impl_provider_list_all(user=nil) #:nodoc:
+
+    init_connection
+    fileData = @s3_connection.list_objects_short()
+    list = []
+    
+    # First parse the files
+    fileData[:files].each do |f|
```
prioux: These two array code blocks are identical, except for the symbolic_type, can't we make them as one, and remove the redundancy?

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,243 @@ def self.pretty_category_name #:nodoc:
...
+  
+  
+#def impl_provider_readhandle(userfile, rel_path = ".") #:nodoc:
+#  init_connection
+#  
+#  full_path = File.join(provider_full_path(userfile),rel_path)
+#  cb_error "Error: read handle cannot be provided for non-file" unless userfile.is_a?(SingleFile)
+#  @s3_connection.list
+#  end
+end
```
prioux: Missing newline at end of file

BrainPortal/app/models/s3_data_provider.rb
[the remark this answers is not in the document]
prioux: Never mind, found it, it's in the S3Connection class...

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,244 @@
...
+require 'aws-sdk-s3'
+require 'fileutils'
 
 # A handler connections to Amazon's S3 service.
 class S3Connection
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
 
-  attr_accessor :base,:service, :bucket, :s3object
+  attr_accessor :resource, :bucket_name, :path_name, :region
```
prioux: It doesn't look like the class use any of these attributes using methods (e.g. `.bucket_name` or `.bucket_name=` ), so maybe there is no need for attr_accessor? Or on the other hand, the code could be modified such that it doesn't set and get using the instance variables directly.

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,244 @@
...
+    @bucket = @resource.bucket(@bucket_name)
+    @path_name = path_start
+    @endpoint = endpoint
   end
-
+  
+  # Method to translate Mime types to File types
+  # Redo more generally
+  def translate_content_type_to_ftype(ct)
+    ct == "application/x-directory" ? :directory : :regular
```
prioux: Other type supported? symbolic links at least? We have them in our data.

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,244 @@
...
+  def self.create_bucket(bucket_name)
     @bucket.create(bucket_name)
   end
 
   # Returns true if the connection is alive
+  # Need to figure out how to return the actual error message so that one can debug
   def connected?
-    @base.connected?
+    begin
+      testVar = @resource.client.list_objects(bucket: @bucket_name, delimiter: "/")
```
prioux: Isn't it costly to list a whole bunch of objects (can be thousands!) just to test the alivemness?

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,244 @@
...
+  # returns a list of each object
+  def list_objects_long(path=nil)
+    path = @path_name if path.nil?
+    
+    pathClean = clean_starting_folder_path(path).to_s
+    list_of_objects = Array.new()
+    
+    if object_exists?(pathClean)
+      cont_token = nil
+      while true
```
prioux: Don't use `while true`, I'm afraid of infinite loops

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,244 @@
...
+    keyPath = File.join(dest_head,trun_object_key).to_s
+    resp = @resource.client.put_object(body: srcFile,
+                                       bucket: @bucket_name,
+                                       key: keyPath)
+  end
+  
+  # Copies a directory to the bucket, will recursively create directory in the path
+  def copy_directory_to_bucket(srcDir, dest_head)
+    if object_exists?(dest_head)
+      Dir.glob("#{srcDir}/**/*").each do |x|
```
prioux: Will this work if we're uploading an empty directory?

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,244 @@
...
+    end
+  end
+  
+  # Renames an objects path
+  def rename_object(srcPath,destPath)
+    src_path_clean = clean_starting_folder_path(srcPath).to_s
+    dest_path_clean = clean_starting_folder_path(destPath).to_s
+    if object_exists?(src_path_clean)
+      src_path_w_bucket = "/#{File.join(@bucket_name,src_path_clean)}"
+      resp = @resource.client.copy_object(bucket:@bucket_name,
```
prioux: Renaming something requires copying?!? Just a comment on AWS.

BrainPortal/lib/s3_connection.rb
```diff
@@ -19,52 +18,244 @@
...
+                                          copy_source: src_path_w_bucket,
+                                          key: dest_path_clean)
+      delete_path_from_bucket(src_path_clean)
+    end
+  end
+  
+  # Renames a Folder by renaming each object underneath
```
prioux: There's no way to rename just a top folder?!?

BrainPortal/app/models/s3_data_provider.rb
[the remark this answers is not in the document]
shots47s: I will consider this resolved then and correct.

BrainPortal/app/models/s3_data_provider.rb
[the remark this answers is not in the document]
shots47s: I will consider this resolved and correct then.

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,264 @@ def self.pretty_category_name #:nodoc:
...
   end
 
   def content_storage_shared_between_users? #:nodoc:
-    false
+    true
   end
 
+  def provider_full_path(userfile)
+    init_connection
+    @s3_connection.clean_starting_folder_path(userfile.name)
```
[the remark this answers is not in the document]
shots47s: I followed the previous S3 Data Provider's convention of the @s3_connection object not being persistent and so init_connection must be called before any use of the @s3_connection object.  There is no persistent connection to the S3 store, the connection just holds variables needed to use the rest interface to find data. Personally, I would not have done this in this manner, but I was following what was done before and the ssh providers where this is not the case.

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -28,7 +28,8 @@ class S3DataProvider < DataProvider
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
 
-  validates_presence_of :cloud_storage_client_identifier, :cloud_storage_client_token
+  validates_presence_of :cloud_storage_client_identifier, :cloud_storage_client_token,
```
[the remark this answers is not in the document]
shots47s: I understand, do you think we should rename this or make the old DP something named differently?

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -40,137 +41,243 @@ def self.pretty_category_name #:nodoc:
...
+  def impl_provider_collection_index(userfile, directory = :all, allowed_types = :regular) #:nodoc:
+    list = []
+    types = allowed_types.is_a?(Array) ? allowed_types.dup : [allowed_types]
+    
+    types.map!(&:to_sym)
+    
+    init_connection
+    entries = []
+    if userfile.is_a? FileCollection
+      if directory == :all
```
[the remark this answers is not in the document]
shots47s: Chalk this under "I forgot to do this, but thought I was going to do this :( . I will implement this.

BrainPortal/lib/s3_connection.rb
shots47s: Can only be 1000 and this is only one API call to the Rest interface, it was the preferred way that was described by the developers of aws-s3-sdk to do this.  You actually have to make an api call for the gem to connect to the s3 store.  So this should not present significant overhead.

BrainPortal/lib/s3_connection.rb
shots47s: You should learn to embrace you infinite loop while true overlords.

BrainPortal/lib/s3_connection.rb
shots47s: Well, there is no such thing as a directory in an AWS S3 store, just paths, but I think so, I will test directly.

BrainPortal/lib/s3_connection.rb
shots47s: I am not sure what you mean, the bucket name should never be changed.  But you can change the name of any folder.

BrainPortal/app/models/s3_data_provider.rb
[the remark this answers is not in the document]
shots47s: There should be no reason that both gems could not be loaded.

BrainPortal/lib/s3_connection.rb
shots47s: Copying a renaming a path is the same operation, it is just renaming the object keys, so it is not really a flaw in S3, it is just how object store is done.

BrainPortal/lib/s3_connection.rb
shots47s: No, there is no way outside of the S3 interface to use the API to create "empty directories".  Is there is a usecase in CBRAIN where we will be doing that?  So we can:

1) Try to make it do something that will allow it.
2) Make a check to throw an error if one tries to do this.

shots47s: What it will currently do is create a Userfile to the database as an empty filecollection, and then nothing appears on the S3 Store

shots47s:
It also now handles symbolic links correctly.

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -0,0 +1,293 @@
...
+
+  # This returns the category of the data provider
+  def self.pretty_category_name #:nodoc:
+    "Cloud"
+  end
+
+  # Connects to the S3 service using :cloud_storage_client_identifier and :cloud_storage_client_token;
+  # the connection is maintained in a instance variable!
+  def init_connection
+    @s3_connection = S3Sdkv3Connection.new(self.cloud_storage_client_identifier,
```
prioux: Shouldn't we have

```ruby
@s3_connection ||= S3Sdkv3Connection.new(blah)
```

instead, so that we don't have to re-open the connection every time if we do several operation side by side?

BrainPortal/app/views/data_providers/_dp_browse_table.html.erb
```diff
@@ -111,11 +111,12 @@
         )
       end
 
-      t.column('Last modified', :mtime,
-        :sortable => true
-      ) { |f| to_localtime(Time.at(f.mtime), :datetime) }
+      t.column('Last modified', :mtime, :sortable => true)  do |f|
+       mtime = f.mtime
+       next mtime.nil? ? '-' : to_localtime(Time.at(f.mtime), :datetime)
+	  end
```
prioux: You have a TAB here, the indentation is all screwed up

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -0,0 +1,293 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# Implements a DataProvider that stores its files
```
prioux: This description is wrong, it still refers to the old `S3DataProvider`

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -0,0 +1,293 @@
...
+# in a Amazon S3 bucket. Single files are saved
+# as-is, FileCollections are tar'ed and untar'ed
+# as needed.
+class S3FlatDataProvider < DataProvider
+
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  validates_presence_of :cloud_storage_client_identifier, :cloud_storage_client_token,
+                        :cloud_storage_client_path_start, :cloud_storage_client_bucket_name
+
```
prioux: Add Rails validations for these attributes

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -0,0 +1,293 @@
...
+  attr_accessor :s3_connection
```
prioux: You define an accessor method for the variable, yet everywhere you use the instance variable directly, EXCEPT at lines 97 and 101. Chose one or the other, but not a mix... (I suggest removing the accessor completely, unless you code external to this class will need to access it).

BrainPortal/lib/s3_sdkv3_connection.rb
```diff
@@ -0,0 +1,346 @@
...
+        xStat = get_object_stats(x[:key])
+        next if xStat[:content_type] == 'application/x-directory'
+        mod_dest_path = x[:key].dup.sub! src_path_clean,dest_path_clean
+        rename_object(x[:key],mod_dest_path)
+      end
+    end
+  end
+
+  # List the buckets available
+  def list_buckets
```
prioux: This method doesn't seem to be used by anything?

shots47s: Correct, it is included in the protocol library for completeness.

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -0,0 +1,299 @@
...
+  end
+
+  # Connects to the S3 service using :cloud_storage_client_identifier and :cloud_storage_client_token;
+  # the connection is maintained in a instance variable!
+  def init_connection
+    @s3_connection = S3Sdkv3Connection.new(self.cloud_storage_client_identifier,
+                                           self.cloud_storage_client_token,
+                                           self.cloud_storage_client_bucket_name,
+                                           self.cloud_storage_client_path_start)
+    @s3_connection.create_bucket(@s3_connection.bucket_name) unless @s3_connection.bucket_exists?(@s3_connection.bucket_name)
```
prioux: These two lines mean whenever we perform ANY action, we test again and again for the bucket, and we also create and recreated the dummy placeholder file? Since these things only need to be performed ONCE in the lifetime of the DP, why not put them in the `is_alive?()` method instead?

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -0,0 +1,299 @@
...
+  def impl_provider_collection_index(userfile, directory = :all, allowed_types = :regular) #:nodoc:
+    list = []
+    types = allowed_types.is_a?(Array) ? allowed_types.dup : [allowed_types]
+    types.map!(&:to_sym)
+
+    init_connection
+    entries = []
+    if userfile.is_a? FileCollection
+      if directory == :all
+        entriesTmp = s3_connection.list_objects_long(userfile.name)
```
prioux: Crash here with `NameError: undefined local variable or method 's3_connection'`

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -0,0 +1,299 @@
...
+        entriesTmp.each do |e|
+          x = @s3_connection.get_object_stats(e[:key])
+          x[:name] = e[:key]
+          entries << x
+        end
+      else
+        if directory == :top or directory == "."# Think this indicates the whole bucket
+          entriesTmp = s3_connection.list_objects_short(userfile.name.to_s + "/")
```
prioux: Again, crash here.

BrainPortal/app/models/s3_data_provider.rb
```diff
@@ -106,9 +114,7 @@ def impl_sync_to_cache(userfile) #:nodoc:
       else
         dest_fh = File.new(local_full,"w:BINARY")
       end
-      @s3_connection.s3object.stream(remote_filename, bucket_name) do |chunk|
-        dest_fh.write chunk
-      end
+      s3_connection.copy_object_from_bucket(remote_filename,dest_fh)
```
prioux: Crash here with `NameError: undefined local variable or method 's3_connection'`

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -0,0 +1,299 @@
...
+      fileinfo = FileInfo.new
+      fileinfo.name          = entry[:name]
+      fileinfo.symbolic_type = type
+      fileinfo.size          = entry[:content_length]
+      fileinfo.mtime         = entry[:last_modified]
+      fileinfo.owner         = "s3"
+      fileinfo.group         = "s3"
+      list << fileinfo
+   end
+   list.sort! { |a,b| a.name <=> b.name }
```
prioux: Bugs:
* When asking for files, all returned entries are prefixed with the bucket name
* When asking for directories, always got an empty array

shots47s:
I have installed and deployed this on my local CBRAIN, and tested it with Amazon S3, works as expected.

PR #739, opened 2018-04-18
Parallelizer recovery. Fixes #651

### LINKED ISSUE #651: Parallelizer to mark subtasks as 'Data Ready' even if it terminates improperly
There are cases where the parallelizer is terminated improperly (or rather, properly if it's the cluster that decides the CPU time's been exceeded) and the parallelizer is marked as `Failed On Cluster` while the subtasks stay stuck at `Configured`. I think the Parallelizer should register a status transition callback to handle the transitions to `Failed On Cluster` so that the subtasks can be marked as 'Data ready' in case they are. Those subtasks that have completed will proceed normally, while the others that were also terminated at the same time as the parallelizer will individually be marked as `Failed On Cluster`.

prioux:
Alright this is complex code change with several enhancements.

**Situation:** the Parallelizer normally runs sveral other subtask's bash scripts in parallel on their behalf, using a script more or less like this

```bash
# This is the parallelizer's script submitted to the cluster
cd subtask1_dir; bash subtask1.sh &
cd subtask2_dir; bash subtask2.sh &
wait
```

**The problem:** if the subtasks take too long, the cluster can terminate the entire parallelizer script and all its subprocess. Some of them might be finished, some of them might be incomplete. What we were left with was

* A parallelizer in state 'Failed On Cluster'
* Each subtask stuch in state 'Configured'

**The solution in this PR:** Clicking 'recover' on the Parallelizer now enables it to look at the state of its subtasks are reset things appropriately:

* Subtasks that had completed are simply marked as 'Data Ready' and the parallelizer is configured to no longer 'control' it
* Subtasks that were interrupted are marked as 'Recover Cluster' and they go through their normal cluster recovery code, if any
* The Parallelizer waits for them to recover and to return to 'Configured', and then resubmits its parallelizer script (with this time only the tasks a subset of tasks, nto recomputing tasks that were completed the prevoius time around)

The state monitoring of subtasks is performed by having each subtask create a touchfile in the parallelizer's work directory when they complete.

### DIFF
--- a/Bourreau/app/models/bourreau_worker.rb
+++ b/Bourreau/app/models/bourreau_worker.rb
@@ -418,15 +418,8 @@ class BourreauWorker < Worker
         else # OK, we have the go ahead to retry the task
           task.addlog("Successful recovery from '#{fromwhat}' failure, now we retry it.")
           if fromwhat == 'Cluster' # special case, we need to resubmit the task.
-            begin
-              Dir.chdir(task.full_cluster_workdir) do
-                task.instance_eval { submit_cluster_job } # will set status to 'Queued' or 'Data Ready'
-                # Line above: the eval is needed because it's a protected method, and I want to keep it so.
-              end
-            rescue => ex
-              task.addlog_exception(ex,"Job submit method raised an exception:")
-              task.status_transition(task.status, "Failed On Cluster")
-            end
+            task.meta[:submit_without_setup] = true
+            task.status_transition(task.status, "New") # basically, we will go directly to submit_cluster_job()
           else # simpler cases, we just reset the status and let the task return to main flow.
             task.status_transition(task.status, "New")                    if fromwhat    == 'Setup'
             task.status_transition(task.status, "Data Ready")             if fromwhat    == 'PostProcess'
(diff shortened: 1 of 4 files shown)


prioux:
This will fix #651 by the way.

prioux:
Help in testing this:

1. In your dev env, create a Diagnostics tool config with parallelization set to 5
2. launch a diagnostics with the number of copies of the task set to 5 and cluster delay set to 20 seconds
3. Wait until all is finished; this is a base setup 
4.Manually each of the diagnostics object's params such that cluster dalay is 20 for the first task, 40 for the second task, 60 for the third, 80 for the fourth, and 100 for the fifth. That will give you time to kill the process during the next relaunch and have some subtasks that will compelte and some that will not (if you kill the stuff between 20 and 80 seconds after the next restart)
5. In the console issue this:
```ruby
p=CbrainTask::Parallelizer.last;st=CbrainTask.find(p.params[:task_ids_enabled].keys).to_a;all=[p]+st;p.bourreau.send_command_alter_tasks(all, "RemoveWorkdir");all.each { |t| t.meta[:configure_only] = t.is_a?(CbrainTask::Parallelizer) ? nil : true;t.status = 'New';t.save;t.destroy_log };st.each { |s| p.enable_subtask(s) };p.save
```
This will reset everything as if just created from scratch
6. Then issue this to start the reprocessing: `p.bourreau.send_command_wakeup_workers`
7. monitor your processes and manually kill all the process that have "bash .qsub" in the command line
8. You should get the setup in **The Problem** above in the PR description
9. Click recover on the Parallelizer, watch what happens

PR #742, opened 2018-04-23
Restart postprocess; fix 736

### LINKED ISSUE #736: Restarting Post Processing fails when checking for END token
The code that checks that a task's wrapper has finished completely by looking for the `CBRAIN Task Exiting` keyword fails when performing a Restart At PostProcessing.

The reason is that the run number is increased and the check tries to find the output of the wrapper using a filename based on that run number. When restarting at post-processing, there is no such output.

prioux:
This will fix #736 .

Also added one more option to the diagnostics tasks to test other situations. The behavior is changed in that in the past the stdout and stderr were copied when doing a restart at post processing, but now the user must ask for this explicitely. To test this PR one must leave the new option disabled (by default) and also check the `Do not actually submit job script on cluster` checkbox at the moment of restart (because Diagnostics is too intelligent and checks for its own keywords in the outputs).


natacha-beck:
Everything is fine, just point some typo that can be fixed before merging.

PR #763, opened 2018-08-08
Upgraded Boutiques schema

natacha-beck:


glatard:
I guess you used the schema from the latest Boutiques release (0.5.12), right?

natacha-beck:
Oupss, take the descriptor from master.... Will check if it has changed. 
Must be the same if I looked well.

glatard:
yes, the master branch is at the latest tag.

Issue #771, opened 2018-09-04
Show table helpers: add ability to have a single FORM wrapping several show tables.


prioux:
Right with each show table helper invocation, we get one form surrounding one table:

```erb
<%= show_table(@obj) do |tb, f| %>
   <% tb.edit_cell { f.text_field(:field1) } %>
<% end %>
<%= show_table(@obj) do |tb, f| %>
   <% tb.edit_cell { f.text_field(:field2) } %>
<% end %>
```

produces this:

```html
<form>
  <table>
     <text_field1/>
  </table>
</form>
<form>
  <table>
     <text_field2/>
  </table>
</form>
```

If we want to create a single form with multiple tables in it, for the same `@obj`, we can't.

So let's add a capability to create a table WITHOUT the surrounding form element, and the capability to generate the form context separately. We'll need to pass the form helper around.

Something like

```erb
<% show_table_context(@obj) do |cf| %>
  <% show_table(@obj, :form_context => cf) do |tb, f|
      <% tb.edit_cell { f.text_field :field1 } %>
  <% end %>
  <% show_table(@obj, :form_context => cf) do |tb, f|
      <% tb.edit_cell { f.text_field :field2 } %>
  <% end %>
<% end %>
```

which would produce:

```html
<form>
  <table>
    <text_field1/>
  </table>
  <table>
    <text_field2/>
  </table>
</form>
```

prioux:
This should be integrated into show page for ToolConfig objects for creating brand new ones in #765 .

prioux:
Done in #772

PR #765, opened 2018-08-14
Toolconfig new fields, ux / refactoring to table helper for config/version show and new page and tool show #650

MontrealSergiy:

Issue #650, opened 2017-09-18
ToolConfig "show" page should provide more info about newest attributes


prioux:
We've added a lot of attribtues to the ToolConfig models, yet the "show" page only lists the `version_name`, `group`, and `ncpus` .
Let's maybe make it nicer and clearer; keep the full bash prologue well below the information.

We could have an option to show the info just for the TC, or the full set of inherited TCs (like it is now).

MontrealSergiy:
also need more work on harmonisation with other pages (e.g. use show table)

MontrealSergiy:
As discussed changed controller, routes, create new. Doing same to tool show/new

MontrealSergiy:
sorry forgot to push one last corrections

prioux:
Fixed by #765

PR #772, opened 2018-09-04
Show table context

prioux:
It is now possible to generate a page with several `show_table()` in it, such that all of them are within the same form. See description of #771 for an example. This PR also includes an example of using this in the AccessProfile `show` page.

This should be very useful to Sergiy to make the new and edit page of ToolConfigs finally be one and the same (see #765 and #650 ).

### DIFF
--- a/BrainPortal/app/helpers/show_table_helper.rb
+++ b/BrainPortal/app/helpers/show_table_helper.rb
@@ -79,12 +79,20 @@ module ShowTableHelper
       # Appearance
       @header          = options[:header].presence || "Info"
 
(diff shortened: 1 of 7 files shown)


prioux:
Clarification: this PR also includes some unrelated files that unfortunately were also modified in my dev branch:

* BrainPortal/app/views/tasks/show.html.erb
* BrainPortal/config/console_rc/lib/pretty_view.rb
* BrainPortal/config/console_rc/lib/wirble_hirb_looksee.rb

You can review that too.

Issue #785, opened 2018-11-28
API not-found error code


MontrealSergiy:
If we follow usual HTTP conventions, API should return error code 404 for nonexisting items (page/resource not found), and not 500 (server error). Use case: correct return code helps API user, that is front-end developer, generate a correct error message, or diagnose an issue in his own code (e.g. mixing  tools vs bourreaux or file ids), 

**To reproduce**
1) generate a random number 
2) procure an api tocken say 3333333333
3) send an http request (using curl, postman or whatever your prefered client)  `localhost:3000/bourreaux/random_number?token`
 where token is authorization token and randon_number where obtained in previous steps

**Observed**
HTTP status  code 500

**Expected**
HTTP status code 404

Using correct codes helps API to diagnose problem.

prioux:
I get a proper 404 error code. I think maybe your test environment isn't set up properly? Trying to find a record that doesn't exist raise a ruby `ActiveRecord::RecordNotFound`, which triggers the exception handler 'record_not_found' in lib/exception_helpers.rb  ; that in turn returns the 404 code properly.

That being said I got some HTML output in the body, even when requesting application/json, so that needs to be fixed.

MontrealSergiy:
Yup on production I see correct 404 code ( on my PC Rack error <ActiveRecord::RecordNotFound>)

MontrealSergiy:
I see no HTML in response, just a fine JSON message do you see an issue with it?

MontrealSergiy:
(Actually not so fine) but still meaningfull

{
    "error": "Couldn't find RemoteResource with 'id'={{id}}"
}

prioux:
Side note: I think in `exception_helper.rb` we could consider changing all lines such as:

```ruby
raise unless Rails.env == 'production' #Want to see stack trace in dev.
```

to

```ruby
raise if Rails.env == 'development' # Want to see stack trace in dev.
```

Not sure of the consequences, though.

MontrealSergiy:
https://portal.cbrain.mcgill.ca:4000/bourreaux/33333

also giving 404

prioux:
I just tested again, and all seems fine:

* In development environment, we get 404 with a HTML page with stack traces etc, which is expected
* In a production environment, we also get 404 with a JSON error message

I wasn't able to get a 500 ever.

PR #787, opened 2018-12-06
fixing the 500 api error and adding richer error details

MontrealSergiy:
Hi I worked around the bug #785 and, additional provided more data to json to faciliate usage front-end/third party. Could you check it out?

### DIFF
--- a/BrainPortal/app/controllers/userfiles_controller.rb
+++ b/BrainPortal/app/controllers/userfiles_controller.rb
@@ -306,7 +306,7 @@ class UserfilesController < ApplicationController
     rescue ActionView::Template::Error => e
       exception = e.original_exception
 
-      raise exception unless Rails.env == 'production'
+      raise if Rails.env == 'development'
       ExceptionLog.log_exception(exception, current_user, request)
       Message.send_message(current_user,
         :message_type => 'error',
--- a/BrainPortal/app/controllers/userfiles_controller.rb
+++ b/BrainPortal/app/controllers/userfiles_controller.rb
@@ -306,7 +306,7 @@ class UserfilesController < ApplicationController
     rescue ActionView::Template::Error => e
       exception = e.original_exception
 
-      raise if Rails.env == 'development'
+      raise exception unless Rails.env == 'production'
       ExceptionLog.log_exception(exception, current_user, request)
       Message.send_message(current_user,
         :message_type => 'error',
--- a/BrainPortal/lib/exception_helpers.rb
+++ b/BrainPortal/lib/exception_helpers.rb
@@ -113,7 +113,7 @@ module ExceptionHelpers
 
   # Anything else is serious.
   def generic_exception(exception)
-    raise unless Rails.env == 'production' #Want to see stack trace in dev. Also will log it in exception logger
+    raise if Rails.env == 'development' #Want to see stack trace in dev. Also will log it in exception logger
 
     # Note that send_internal_error_message will also censure :password from the params hash
     exception_log = ExceptionLog.log_exception(exception, current_user, request) # explicit logging in exception logger, since we won't re-raise it now.
--- a/BrainPortal/lib/exception_helpers.rb
+++ b/BrainPortal/lib/exception_helpers.rb
@@ -24,6 +24,9 @@ module ExceptionHelpers
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
 
+  NOT_FOUND_MSG = "The object you requested does not exist or is not accessible to you." #:nodoc:
+  CANNOT_DELETE_MSG = "The requested object could not be deleted." #:nodoc:
+
   def self.included(includer) #:nodoc:
     includer.class_eval do
       rescue_from StandardError,                        :with => :generic_exception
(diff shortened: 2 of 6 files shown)


prioux:
Please cleanup the PR, it seems you left a lot of debugging code in there.

BrainPortal/lib/exception_helpers.rb
```diff
@@ -34,40 +36,59 @@ def self.included(includer) #:nodoc:
     end
   end
 
+  NOT_FOUND_MSG = "The object you requested does not exist or is not accessible to you."
```
prioux: Constants should be defined at the top of the file, after Revision_info, separated by one blank line before and after the constants (and not in between). If no documentation is provided, or they must not be documented, append #:nodoc: . Look at `scir.rb` for examples.

BrainPortal/lib/exception_helpers.rb
```diff
@@ -34,40 +36,59 @@ def self.included(includer) #:nodoc:
...
+  CANNOT_DELETE_MSG = "The requested object could not be deleted."
+
+
   protected
 
-  # Record not accessible.
```
prioux: Comment was deleted here, why?

BrainPortal/spec/controllers/users_controller_spec.rb
```diff
@@ -586,7 +588,7 @@
         end
 
         it "should not allow site manager to destroy a user not from the site" do
-          expect { delete :destroy, params: {:id => user.id }}.to raise_error(ActiveRecord::RecordNotFound)
+          expect { delete :destroy, params: {:id => user.id }}
```
prioux: This is an error: this expect() does not test anything, the test will never fail.

BrainPortal/spec/controllers/users_controller_spec.rb
```diff
@@ -22,6 +22,8 @@
 
 require 'rails_helper'
 
+
```
prioux: Why are there new blank lines here?

BrainPortal/spec/spec_helper.rb
```diff
@@ -13,6 +13,11 @@
 # it.
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
+#
```
prioux: Why are there new blank lines here too?

BrainPortal/test_api/req_files/groups/badget/out
```diff
@@ -0,0 +1,2 @@
+404 application/json  \[.*
```
prioux: Why is there a regex for a elimination pattern that will never happen in the output?

BrainPortal/lib/exception_helpers.rb
```diff
@@ -20,6 +20,8 @@
 #
 
 # Controller helpers to elegantly handle and log runtime exceptions.
+#
```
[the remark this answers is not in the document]
MontrealSergiy: Ok, it just looked better for me with empty line. That is not necessary for the bugfix so I am removing, but I do not think it a line between commentary and code would breaks rdoc.

BrainPortal/spec/controllers/users_controller_spec.rb
MontrealSergiy: fixing

BrainPortal/spec/controllers/users_controller_spec.rb
MontrealSergiy: removing

BrainPortal/spec/controllers/users_controller_spec.rb
MontrealSergiy: fixing

BrainPortal/test_api/req_files/groups/badget/out
MontrealSergiy: This one no idea, actually I think you suggested it. Should I delete?

prioux: Yes please, no need to confuse future testers about why that's there, especially since there are no square brackets in the output anymore.

Yes I did suggest it, but that was appropriate at the time, it allowed comparisons of the outputs before you changed the outputs to something more deterministic.

BrainPortal/lib/exception_helpers.rb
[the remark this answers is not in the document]
prioux: Please do address all my comments, you still have spurious blanks lines that were not in the original code.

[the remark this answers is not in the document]
MontrealSergiy: ok

BrainPortal/spec/spec_helper.rb
MontrealSergiy: removed

prioux:
Please rebase this because it's undoing some changes that have happened since you created that PR.

MontrealSergiy:
rebased

BrainPortal/lib/exception_helpers.rb
```diff
@@ -38,36 +41,51 @@ def self.included(includer) #:nodoc:
 
   # Record not accessible.
   def record_not_found(exception)
-    raise unless Rails.env == 'production' #Want to see stack trace in dev.
-    flash[:error] = "The object you requested does not exist or is not accessible to you."
+    raise if Rails.env == 'development' #Want to see stack trace in dev.
```
prioux: At two places you changed the exception forwarding conditional from

```ruby
raise unless Rails.env == 'production'
```

to

```ruby
raise if Rails.env == 'development'
```

The difference in behavior is only while in the `test` environment. Is there a good reason to have this change? This file contains two other methods with similar checks and they have been left with the original check (`== 'production'`). I don't like that we have two methods with one way, and two methods the other way. Can you double-check that please? It's particularly weird because the original code is 7 years old (from the original file commit made by @natacha-beck in April 2012) so I don't see why it would suddenly no longer be appropriate.

MontrealSergiy:
The main motivation for this is testing API error codes. Probably in 2012  were no API so having same error codes in testing and production environments was less important.

The developer/tester might want to know that API produces error codes such as 400, 401, 404 etc appropriately.  I skipped the 'generic_exception' method because it returns 500 in either case of dev or test. Another method, display, I probably missed because it also returns 500 on all the environments, or may  be because it was in a different file (userfiles_controller.rb).  I can change these for greater uniformity if you like. Probably there are a few other unrelated env `unless`es here and there.

prioux:
Yes please do change all places in this file so it says:

```ruby
raise if Rails.env == 'development'
```

that way our test environment works like in production. I think there is only one place left (in `generci_exception()` )

Make a quick commit and I'll merge this.

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -306,7 +306,7 @@ def display
     rescue ActionView::Template::Error => e
       exception = e.original_exception
 
-      raise exception unless Rails.env == 'production'
+      raise if Rails.env == 'development'
```
prioux: That's a bug

PR #788, opened 2018-12-06
Swagger format

MontrealSergiy:
swagger improvements

### DIFF
--- a/BrainPortal/public/swagger/cbrain-5.1.0-swagger.json
+++ b/BrainPortal/public/swagger/cbrain-5.1.0-swagger.json
@@ -1215,12 +1215,12 @@
         },
         "last_connected_at" : {
           "type" : "string",
-          "format" : "dateTime",
+          "format" : "date-time",
           "description" : "time of last connection by the user. (can be empty)"
         },
         "account_locked" : {
           "type" : "string",
-          "format" : "boolean",
+          "enum": ["true", "false"],
           "description" : "Whether or not the account is locked."
         }
       }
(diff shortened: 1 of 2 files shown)


prioux:
Please move this PR to the dev branch, it's properly based on dev but by merging into master we'd get commits that do not belong there yet (as can be seen in https://github.com/aces/cbrain/network )

prioux:
All is good except one thing: I am not sure about that `enum: ["true", "false"]` is actually ok in all the cases of the spec. We have not been consistent in our handling of such attributes, and I fear sometimes we have `"0", "1"` or `true, false` (which is different from `"true", "false"`)... these need to be checked.

prioux:
I suggest looking at the CURL outputs. Just running all the tests with `-v5` should provide all the information.

MontrealSergiy:
Looks like it returns actual boolean, and swagger 2 does not seem support oneOf ( several types) .

Loris actually uses an empty string for account_locked, so switching to type: boolean would make his client invalid

prioux:
The swagger spec is now in files with "5.1.1", and also I am still not convinced the proper values are specified for the boolean returns. We need to be consistent, including modifying the controllers as necessary. The JSON output shoudl have `true` and `false` everywhere.

shots47s:
Is this pull request still necessary?

prioux:
@shots47s Well I wanted Sergiy to perform an audit of all the true/false/"true"/"false" values of our JSON outputs, and fix the controllers and the swagger spec as necessary. The PR itself is useless because it modifies swaggers specification files that are not supposed to be touched (we normally create new ones as we improve them).

prioux:
So I'll reject it and create an issue.

PR #795, opened 2019-02-14
Generator subclasses

prioux:
This will support the BidsAppHandler abstract class in the cbrain-plugins-neuro.

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
@@ -89,6 +89,9 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
   # directories, prepare symlinks to input files, set environment variables,
   # etc. Returns true if the task was correctly set up.
   def setup #:nodoc:
+
+    return false unless super
+
 % unless files.empty?
     params = self.params
 
@@ -113,7 +116,7 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
     end
 
     # And make them all available to <%= name %>
-%   make_available = format_call('make_available', files) { |file| [
+%   make_available = format_call('multi_make_available_to_dir', files) { |file| [
 %     "params[:'#{file['id']}']",
 %     "'.'"
 %   ] }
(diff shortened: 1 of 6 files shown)


BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -163,6 +166,14 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
   # in params.
   def cluster_commands #:nodoc:
 
+    # OK so this is crazy stuff, if we have a superclass
+    # that is not the default ClusterTask, we trust that it knows
+    # how to generate the commands better than we do, so we
+    # just delegate all the work to it.
+    if self.class.superclass < ClusterTask
```
shots47s: I have never seen this kind of syntax before, does this really mean "less than" the ClusterTask, does it imply that it is lower on the hierarchy of class structure?

prioux: Yeah, the `<` and `>` comparison methods (they are methods!) are defined on class object (classes are objects!) so that you compare classes within their hierarchy. `if A < B` test that class A inherits from B (and not necessarily directly). E.g. in CBRAIN we have a class hierarchy with `Userfile < SingleFile < TextFile < LogFile`

prioux: Oops correction

MontrealSergiy: So the above expression checks that class inherits, yet indirectly?

prioux:
Unusually, there was no official issue associated with this feature.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
prioux: Well, `self` is an instance of a particular class `Abc` for the bidsapp. So `self.class` is Abc. And `self.class.superclass` would be whatever immediate superclass is above.

The test then is `if my object's class' superclass happens to be a subclass of ClusterTask`.

Two examples: if self is an instance of SuperDuper then:

* when `ClusterTask > SuperDuper`, then if condition is false
* when `ClusterTask > Youppi > SuperDuper` then the condition is true
* when `ClusterTask > Youppi > GlopFoo > SuperDuper` then it is true too, etc.

PR #800, opened 2019-04-24
Datalad provider stb

shots47s:
An updated Datalad provider that does all of the necessary things to cache the datalad repo for maintaining the file information inside the datasets.

This is only capable of reading from a datalad dataset now, writing comes next.

### DIFF
--- a/BrainPortal/app/controllers/data_providers_controller.rb
+++ b/BrainPortal/app/controllers/data_providers_controller.rb
@@ -133,6 +133,7 @@ class DataProvidersController < ApplicationController
            remote_user remote_host remote_port remote_dir
            not_syncable cloud_storage_client_identifier cloud_storage_client_token
            cloud_storage_client_bucket_name cloud_storage_client_path_start
+           datalad_repository_url datalad_relative_path
          )
       )
       meta_flags_for_restrictions = (params[:meta] || {}).keys.grep(/\Adp_no_copy_\d+\z|\Arr_no_sync_\d+\z/)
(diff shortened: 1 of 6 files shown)


shots47s:
I put it in the main CBRAIN directories as datalad is technically not a neuroimaging tool.

prioux:
Shawn, please return this code to the neuro plugin repo, we are losing all the history of the datalad DP code with this new commit. Also, you have left lots of debug code in here and changed the behavior of the platform (see comments soon).

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -468,10 +469,12 @@ def register
     succeeded, failed = [], {}
 
     CBRAIN.spawn_with_active_records_if(
-      [:html, :js].include?(request.format.to_sym),
+      #[:html, :js].include?(request.format.to_sym),
```
prioux: This makes the Rails app block forever

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -468,10 +469,12 @@ def register
...
+      false,
       current_user,
       "Register files DP=#{@provider.id}"
     ) do
+      binding.pry
```
prioux: This makes the rails app block forever expecting a programmer to enter the console, which doens't exist in production

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -617,6 +620,7 @@ def unregister
       current_user,
       "Unregister files DP=#{@provider.id}"
     ) do
+      puts "FUCK ME I AM HERE"
```
prioux: Bad

shots47s:
This doesn't belong  in cbrain-plugins-neuro, can it be moved after we finalize?

BrainPortal/db/schema.rb
```diff
@@ -96,19 +96,21 @@
     t.string   "remote_host"
     t.integer  "remote_port"
     t.string   "remote_dir"
-    t.boolean  "online",                                        default: false, null: false
-    t.boolean  "read_only",                                     default: false, null: false
-    t.text     "description",                     limit: 65535
+    t.boolean  "online",                                         default: false, null: false
```
prioux: Fix so that no chanes are made to the schema when unnecessary

PR #801, opened 2019-04-30
Updated the DataProvider Model and view pages to accommodate new variables for Datalad Dataprovider.

shots47s:


BrainPortal/app/views/data_providers/show.html.erb
```diff
@@ -153,6 +153,15 @@
         <%= text_field_tag "data_provider[cloud_storage_client_path_start]", @provider.cloud_storage_client_path_start, :size => 80 %>
       <% end %>
     <% end %>
+
+    <%= show_table(@provider, :as => :data_provider, :header => "Datalad Configuration", :edit_condition => true) do |t| %>
```
prioux: (Not a request for change, just a comment)

This section of the page could be put inside a `if` block and not shown when the data provider we're currently viewing is not a Datalad type.

PR #804, opened 2019-05-06
During boot, duplicated SyncStatus are cleaned

prioux:
In very rare circumstances, it seems some SyncStatus
objects are created while violating the uniqness
rules for (:bourreau_id, :userfile_id). The boot
process now identifies them and remove the duplicates.
Otherwise, all synchronization actions fail on
the affected userfiles.


MontrealSergiy:
seems ok. I wonder is it lastest status which is most relevant? Could be a situation when the latest status is wrong?

PR #805, opened 2019-05-06
A bit of robustness-related code tweaks.

prioux:
CbrainMailer won't even attempt to send
emails if the From address is missing.

Avoid race condition is caching info for remote resource.

Handle some bad historical task filter object's data.

CSV file size viewer limit increased.


MontrealSergiy:
Not sure that I understand the origin of all these issues in full details, but seems ok

PR #806, opened 2019-05-06
Boutiques bug. Fixes #794

### LINKED ISSUE #794: Boutiques-integrated tasks do no always addlog() messages for the files they create
It seems the Bourreau code only add the provenance record when the params contain a _parent_ userfile (which I'm not too sure what it is). From the `bourreau.rb.erb` file:

```ruby
%   if (single_file = files.first if files.count == 1 && ! files.first['list'])

        # As all output files were generated from a single input file,
        # the outputs can all be made children of the one parent input file.
        parent = Userfile.find_by_id(params[:'<%= single_file['id'] %>'])
        output.move_to_child_of(parent) if parent
        self.addlog_to_userfiles_these_created_these([parent], [output]) if parent
%   end
```

This need to be adjusted so that we always get a addlog() message on the created files.

prioux:
Adds messages to the logs when multiple inputs
are provided to the descriptor.


glatard:
Looks good to me!

PR #811, opened 2019-05-07
Added page and per_page params in Swagger. Fixes #789

### LINKED ISSUE #789: Unable to get full queries of objects through the API without paging
Currently, if you query returns more than the allowed number of objects (which appears to be set at 50), there is no way to continue to get the rest of the query.  We should implement some way to either request further pages or a continuation token so that the client can make subsequent calls to get all of the useful data.

prioux:


prioux:
Note: I have also updated accordingly the client library:

https://github.com/aces/cbrain-client-gem

prioux:
Fixed #789

PR #812, opened 2019-05-08
More robust extracting files from a collection. Fix #781

### LINKED ISSUE #781: Extract to zombie file
Extracting files from a FileCollection sitting on a DataProvider which CBRAIN do not have the write permission creates new file records that can't be deleted.

Error message when deleting: 
```
Error when deleting/unregistering file(s)
[<datetime>] For 1 file, Error: provider <dataprovidername> is read_only.:
<filename>
```

prioux:
Also, we try to guess the file type now.


prioux:
Unit tests failed, will investigate why tomorrow. But you can still review the code.

Issue #779, opened 2018-10-30
Boutiques integrator should support shells other than bash


prioux:
A boutiques descriptor can provide a property that says the command to run is from a specific interpreter, e.g.

```json
{
  "command-line": "blah blah |& whoopsie",
  "shell": "tcsh"
}
```

In CBRAIN we run the command under `bash`, always, which can be a problem. In the example above, `|&` is a csh piping that redirects both stdout and stderr, while in bash this will only cause a bash syntax error.

MontrealSergiy:
I think integrating bosh is big enough to be considered as a full fledged feature (add full boutiques support) rather than a bugfix

prioux:
I'm thinking of adding just two line around the generated command:

```bash
theshell <<'QUOTED_COMMAND'
normal_command here
QUOTED_COMMAND
```

there theshell is what the descriptor specifies. I would add these two lines only if theshell is not already 'bash'.

prioux:
Fixed by #815

PR #815, opened 2019-05-09
Boutiques descriptors can use other shells.

prioux:
If the descriptor contains the "shell" entry, and the
value is something other than bash, then the
command will be wrapped by two more lines:

theshell <<'BoutiquesShellWrapper'
the-command-here
BoutiquesShellWrapper

### DIFF
--- a/Bourreau/spec/boutiques/boutiques_tester_spec.rb
+++ b/Bourreau/spec/boutiques/boutiques_tester_spec.rb
@@ -224,6 +224,19 @@ describe "Bourreau Boutiques Tests" do
           expect( task.cluster_commands[0].strip ).to eq( '/minimalApp -a value' )
         end
 
+        # Test that optional "shell" in descriptor trigger the creation of a wrapper
+        it "should create wrapper commands for alternate shells" do
+          @descriptor['shell'] = 'myFakeShell'
+          task = @generateTask.( { a: 'value' } )
+          expect( task.cluster_commands[0].strip ).to match(
+            /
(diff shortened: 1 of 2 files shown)


prioux:
I'm providing an example of a JSON descriptor (excerpt) and what Ruby code is created from the template, and what Bash code is generated in the final script:

```json
{
    "command-line": "use English;print \"I am Perl $PERL_VERSION\\n\";print [STRING],\"\\n\";", 
    "shell": "/usr/bin/perl",
}
```

```ruby
    # Command template from the descriptor.
    # The descriptor specified that it is to be run with: /usr/bin/perl
    command_template = <<-'COMMAND_TEMPLATE'
/usr/bin/perl <<'BoutiquesShellWrapper'
use English;print "I am Perl $PERL_VERSION\n";print [STRING],"\n";
BoutiquesShellWrapper
    COMMAND_TEMPLATE
```

```shell
# CbrainTask 'PerlVers' commands section
/usr/bin/perl <<'BoutiquesShellWrapper'
use English;print "I am Perl $PERL_VERSION\n";print 'No I Will Not',"\n";
BoutiquesShellWrapper
```

prioux:
Darn. Tests failed. BRB.

prioux:
This will fix #779

prioux:
@tta9 yeah, I did include you on purpose; not really expecting you to review the code as much as just to say “hi!” and “look, your code is still in use!” :-)

ttaa9:
Thanks @prioux, I'm actually quite happy to see my code still alive and well! :) It's very nice to know it continues to be useful.

Issue #768, opened 2018-08-21
Bearer token to only be accepted when it comes from same IP


prioux:
Now that we have support for HTTP Bearer authentication, we could make things more secure by making sure the controllers only accept it if the request comes from the very same IP address that requested it originally.

It's a nice security feature. The IP address would have to be stored in the persistent `LargeSessionInfo` object associated with the session.

prioux:
Fixed by #820

PR #820, opened 2019-05-15
Better handling of IP addresses

prioux:
We use Rail's `request.remote_ip` method instead of our own.

When users roam, the IP address and host name will be updated.
In the past, we always showed the IP address that the user
initialy used when logging in.

Also, requests made with API tokens are checked for their
IP addresses, and if they change, the session is de-activated.


prioux:
This will fix #768

PR #823, opened 2019-05-17
Instances of abstract models won't ever be created.

### LINKED ISSUE #822: Module AbstractModelMethods no longer does anything
That module use to flag classes and prevent new() from working, but it might have been too harsh in the past. The code for new() is commented out. It should still do its job of preventing objects from abstract classes from being saved, at least. To fix.

E.g. in the console

```
g=Group.new
g.type = 'Group'
g.name = "aaa"
g.save! # this works, but shouldn't.
```

prioux:
Fixes #822


prioux:
And of course a bunch of tests failed because the test writers willfully created objects of abstract classes. Let me fix that.

PR #832, opened 2019-06-27
Several fixes all bundled together.

### LINKED ISSUE #25: Double check framework interception of userfile view code exceptions
When there are exceptions in a userfile model's view code or in
the valid_for? tests, handle gracefully.

Example: in minc_file.rb, we have:

``` ruby
has_viewer :name => "Volume Viewer",  :partial => :volume_viewer, 
           :if => Proc.new { |u| u.class.has_minctools?([2,0,0]) && u.is_locally_synced? && u.size < 80.megabytes }
```

but what happens if the size attribute is unset? Make the framework tolerant of such a Proc.

### LINKED ISSUE #45: No warnings when removing a task can't also remove the workdir archive
If a Bourreau doesn't have access to the DP where a task's archive is kept, and the task is deleted, no warning is sent. The task's archive will keep being registered in the DB.

Possible solutions:

1- have the portal remove them? (Prob: time taken, portal might be prevented to access this DP too)
2- warn user, don't do anything else
3- warn admin instead
4- refuse destroying task, tell user

Comment by Marc: 
I would tend to use option 3.
CBRAIN could maintain a list of "things that should be checked by an admin"?

### LINKED ISSUE #55: Implement Remote Resource command 'stop_yourself'
A remote Rails application (such as a Bourreau)
is currently stopped by sending it a signal (SIGTERM
by default) from an external process (the ruby script
cbrain_remote_ctl when remote control is configured).
A new command 'stop_yourslf' on the control channel
could do the trick even more cleanly: the remote
resource receiving it could:
- stop its workers
- stop the SSH master channels for the data providers
- stop the SSH master channels for the
- Bourreaux (N/A for the Bourreaux themselves)
-  send itself a TERM signal using spawn_fully_independent()

Related: 690 || #54 
#### Comments:
- Updated by Marc Rousseau almost 5 years ago

Although cleaner, I feel this is a low-priority back-end improvement which can be administered manually (as we do currently).

### LINKED ISSUE #830: Lists in boutiques task forms are ordered wrong
A boutiques descriptor can contains a description for an input of type list, which contains several items. E.g.

```json
        {
            "command-line-flag": "--block-window-length", 
            "default-value": [
                10, 
                1, 
                30
            ], 
            "id": "block_window_length", 
            "integer": true, 
            "list": true, 
            "max-list-entries": 3, 
            "min-list-entries": 3, 
            "minimum": 1, 
            "optional": true, 
            "type": "Number" 
        }, 
```

What happens is that when the form is generated, three input fields are shown, but the values are inserted in the wrong order:

image (1)

Observed the order is distorted (e.g. 30, 10, 1 instead of 10,1,30)

### LINKED ISSUE #831: File lists and Boutiques-based tools
**Intro**: 
CBRAIN supports import of tools in Boutiques format. Regretfully, input parameters of FIle type in Boutiques tool descriptors, are not properly supported in CBRAIN, while, I think users and tool developers could benefit from full file list support.

An example of the file list input in Boutiques format:
```
        {
            "id": "list",
            "value-key": "[list]",
            "list": true,
            "command-line-flag": "--list",
            "description": "Some file list. Could be e.g. the list of files with MRI data to analyze.",
            "type": "File",
            "name": "files (with MRI data_"
        },
```
(see https://raw.githubusercontent.com/MontrealSergiy/cbrain-plugins-eeg/bugs/cbrain_task_descriptors/filelistbug.json for full descriptor)
CBRAIN seemingly imports boutiques descriptors with file lists without any warning or error message, configuring and executing tool's task is ridden with several issues.

**Steps to reproduce issues**:
1. Create boutiques with a file list input, download from above GitHub link, ( or copy one from Boutiques tests).
2. Import boutiques into CBRAIN;
3. Create a task and a) try to populate the input with more than one file b) run it (with one file list).

**Observed**: User can only supply one file per input, and, anyway, the command line will be malformed and (usually) fail

**Expected**: User should be able to provide multiple files and put them into file list, perhaps in the same way he can create integer or string lists

**Note 1**: The lists of numbers or strings are supported by CBRAIN, the problem is with file lists

**Note 2**: Older versions of boutiques had a few issues with file lists, which are fixed by now

prioux:
1- The controls channel for Bourreau has a new 'StopYourself'
   command, which is much more efficient than the old way
   of stopping the rails app using an external script.
   Fixes #55 .
2- The 'ibc' utility provides access to it with the 'J' command
3- Tooltips in the interface could never be positioned
   at offset_x set to 0 because 0 is falsy in javascript. Fixed.
4- We no longer ever keep a backup of database.yml on remote sites.
5- Userfiles viewers that have conditions which crash no
   longer prevent the 'show' page from rendering; instead the
   viewer is ignored and removed from the available list.
   Fixes #25
6- Task forms for boutiques tasks will present list
   inputs in proper order.
   Fixes #830
7- Support for lists for File inputs in boutiques
   Fixes #831
8- When erasing a task, if the associated archive cannot be erased too
   it will be tagged with the admin-owned Tag 'TaskDestroyed', so the
   admin can find them and deleted them separately later.
   Fixes #45

### DIFF
--- a/BrainPortal/app/models/bourreau.rb
+++ b/BrainPortal/app/models/bourreau.rb
@@ -353,6 +353,15 @@ class Bourreau < RemoteResource
 
   public
 
+  # Utility method to sends a +stop_yourself+ command to a
+  # Bourreau RemoteResource, whether local or not.
+  def send_command_stop_yourself
+    command = RemoteCommand.new(
+      :command     => 'stop_yourself',
+    )
+    send_command(command)
+  end
+
   # Utility method to send a +get_task_outputs+ command to a
   # Bourreau RemoteResource, whether local or not.
   def send_command_get_task_outputs(task_id,run_number=nil,stdout_lim=nil,stderr_lim=nil)
@@ -392,6 +401,18 @@ class Bourreau < RemoteResource
   ############################################################################
 
   protected
+
+  # This comaand is a bit dangerous: it will first trigger a +stop_workers+
+  # command to be sent locally, and then the rails app will kill itself with
+  # a TERM signal, thus exiting right after the current request. The user should
+  # really make sure the workers are inactive and there is no other background
+  # activity.
+  def self.process_command_stop_yourself(command)
+    process_command_stop_workers(
+      command.dup.tap { |com| com.command = "stop_workers" }
+    )
+    Process.kill('TERM',Process.pid) # the rails server will shut down after the current request
+  end
 
   # Starts Bourreau worker processes.
   # This also triggers a 'wakeup' command if they are already
--- a/BrainPortal/config/console_rc/lib/interactive_bourreau_control.rb
+++ b/BrainPortal/config/console_rc/lib/interactive_bourreau_control.rb
@@ -52,8 +52,8 @@ class InteractiveBourreauControl
     @bourreaux = bourreaux_list
     @width     = term_width
     if term_width.blank? || term_width.to_i < 1
-      numrows,numcols = Readline.get_screen_size rescue [25,120]
-      @width           = numcols
+      _,numcols = Readline.get_screen_size rescue [25,120]
+      @width          = numcols
     end
     @selected = {}
   end
@@ -138,7 +138,7 @@ Operations Mode : #{@mode == "each_command" ?
   def process_user_letter(letter) #:nodoc:
 
     # Validate the user input
-    if letter !~ /^([haombwitukygsrczqx]|\d+|exit|quit)$/
+    if letter !~ /^([haombwitukygsrczqxj]|\d+|exit|quit)$/
       puts "Unknown command: #{letter} (ignored)"
       return false
     end
@@ -157,7 +157,10 @@ Operations Mode : #{@mode == "each_command" ?
         W = starts workers
         T = stops workers
         U = stops workers and waits to make sure
-        K = stops bourreaux
+        K = stops bourreaux (using shell commands)
+        J = stops workers and bourreaux (using command control
+            messages; use this only if you're sure no workers
+            are active and the SSH masters are opened)
         Y = cycle: stop workers and bourreaux then
             start bourreaux and workers (replace operation queue)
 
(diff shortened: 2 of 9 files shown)


BrainPortal/app/models/bourreau.rb
```diff
@@ -393,6 +402,18 @@ def send_command_alter_tasks(tasks,new_task_status,extra_remote_commands = {})
 
   protected
 
+  # This comaand is a bit dangerous: it will first trigger a +stop_workers+
+  # command to be sent locally, and then the rails app will kill itself with
+  # a TERM signal, thus exiting right after the current request. The user should
+  # really make sure the workers are inactive and there is no other background
+  # activity.
```
shots47s: Is this to do in the future, or is there a reason you are not making this check?

BrainPortal/config/console_rc/lib/interactive_bourreau_control.rb
```diff
@@ -157,7 +157,10 @@ def process_user_letter(letter) #:nodoc:
         W = starts workers
         T = stops workers
         U = stops workers and waits to make sure
-        K = stops bourreaux
+        K = stops bourreaux (using shell commands)
+        J = stops workers and bourreaux (using command control
+            messages; use this only if you're sure no workers
+            are active and the SSH masters are opened)
```
shots47s: Oh, I see, so still dangerous, but only done by an experienced admin.

BrainPortal/app/models/bourreau.rb
prioux: Because we can't really make the check: once background activity is started it's forked into a separate process completely disconnected from the rails server (using double fork). So the rails server doesn't know if a background process is also using the DB.

BrainPortal/config/console_rc/lib/interactive_bourreau_control.rb
prioux: Well, an admin who can run the 'ps' command first.

prioux:
I added on more feature: when a singularity build fails, we get a capture of the failed command. Fix #833 .

BrainPortal/config/console_rc/lib/interactive_bourreau_control.rb
MontrealSergiy: and fast one :)

(seems fine thought theoretically with same effort some signaling or mutex mechanism can be implemented to avoid the issue)

PR #836, opened 2019-07-04
Added a code of conduct to the .github directory for CBRAIN 

shots47s:
adapted from the CONP code.

Please feel free to add or suggest changes.


dlq:
This is a good idea.

MontrealSergiy:
Good. I personally prefer Python's Code of Conduct to the Covenant, which is short up to the point than an averaged actually defines expected behaviour rather than giving some vague "examples".

Issue #834, opened 2019-07-02
Implement CARMIN API


prioux:
As described in https://github.com/CARMIN-org/CARMIN-API

Several model adaptations will need to be implemented and documented.

prioux:
NOTE: I created this issue to track the development, although I've already implemented 90% of the code already.

prioux:
Implemented in #837

PR #837, opened 2019-07-07
Carmin

prioux:
This PR contains a full implementation of the CARMIN API.

The reference document is:

 [GitHub/CARMIN](https://github.com/CARMIN-org/CARMIN-API/blob/master/carmin.yaml)

and you can also review the pretty SWAGGER layout at

  [SwaggerHub/CARMIN](https://app.swaggerhub.com/apis/CARMIN/carmin-common_api_for_research_medical_imaging_network/0.3.1)

Some notes about the implementation:

* CARMIN's `pipelines` are created from a join of CBRAIN's `Tools` and `ToolConfigs`; their IDs are the `ToolConfig` IDs
* CARMIN's `executions` are CBRAIN's `CbrainTasks`
* CARMIN's `paths` are stored in a new CBRAIN  `DataProvider`, a `CarminPathDataProvider`. This provider is basically a `IncomingVault` type, which means it can conveniently be configured to be the target of a SFTP server.
* CARMIN users will automatically load and save paths to the first CarminPathDataProvider that they have access too, unless they have configured a different prefered one using CBRAIN's interface (assuming more than one is available)
* All CARMIN paths such as `abc/def/xyz.txt` are stored in CBRAIN `FileCollections` named according to the first component (in this case, `abc`).
* All CARMIN paths for top-level files, such as `hello.txt`, will be saved in a CBRAIN `SingleFile`
* Given that most CBRAIN task do no work with paths, it means that to successfully launch a CARMIN task, the user must get the CBRAIN ID of the associated SingleFile or FileCollection. This is a consequence of the fact that CBRAIN is a distributed systems, and paths of data files vary from one system to another.
* All CARMIN path actions typically trigger a sync_to_cache() and/or sync_to_provider(), which means that building a large directory structure component by component can be very slow and expensive.
* The auto-extract feature of carmin is ambiguous, so I made the following decisions:
  * tarball (.tag.gz) only is supported
  * given a path `/abc/def`, the tarball is extracted inside it (it means the path is ALWAYS interpreted as a path to an existing directory)
* A full set of API test commands for curl is available for the CurlReqTester framework.


prioux:
This implements #834

dlq:
If all your tests are passing in CI what tests can I do to check this out?

prioux:
Very good point. But I think you could just try it at least once? With curl? Just to have one more confirmation that it deploys properly. 

Also maybe the PR contains some code changes not directly related to CARMIN?

dlq:
Locally I'm getting:
```
Some tests failed:
 => 077/090 : userfiles/06-content           : Exception: IOError closed stream
```

This seems to be working in Travis CI but the tests aren't executed in the same order there as they are locally for some reason.
```
Running test 55/90 : userfiles/06-content
```

prioux:
It's an interaction with a more recent pull request, you can safely ignore that and I'll fix it myself.

The tests are always executed in random order (more or less)

prioux:
PR #860 is interfering with this merge, and I'll fix that.

dlq:
I don't understand the changes made to `BrainPortal/app/controllers/sessions_controller.rb` and `BrainPortal/app/controllers/tasks_controller.rb` and `BrainPortal/lib/cbrain_extensions/active_record_extensions/abstract_model_methods.rb` and what they have to do with CARMIN.

prioux:
Right. It's not all obvious but they are all related to CARMIN.

In the case of sessions_controller, the code there was too monolithic and I refactored methods so that I could call them from the carmin_controller (e.g. create_from_user() was too integrated tightly and now it returns true/false).

In the case of task_controller, again a lot of refactoring was performed but it's mostly just taking large pieces of code that were in the create() method and putting them into separate methods. Again it's so I could call these separately from the carmin_controller (e.g. CarminController, in method exec_create(), I call create_initial_task_from_form() and create_tasklist_from_initial_task() that are in the TasksController; these last two methods didn't exist before)

The change in abstract_model_methods is to allow a programmer to create an object of class that cannot be saved in the database, then change its class to a model that can.

E.g. create a `u=Userfile.new` in memory (objects of class Userfile cannot be saved, only subclasses can), fill it with data, then change its type with `u.type = SingleFile'.

In the past such code would have refused to save it because it was thinking this is a Userfile, but we are in fact now trying to save it as a SingleFile so it should be ok. I needed this in Camrin because I am doing it somewhere.

prioux:
I'm ready to fix the conflicting files whenever you're satisfied, @dlq

dlq:
Ok.  Given my limited testing and code review and given your assurances on my comments I'll ok this PR.

prioux:
Ah gee. The GitHub's way of resolving the conflict is the merge dev back into carmin, before we merge THAT back into dev... I thought it would be easier that the manual mrge on the command line (which I did on my work computer) but it's making the graph a bit of a mess.

PR #838, opened 2019-07-11
Bourreau workers propagate stop to subworkers; fix #741

### LINKED ISSUE #741: Stop signal not propagated from Workers to Subworkers
On the Bourreau side, when things need to be shut down, a stop signal is sent to the Workers; ay of them that have active subworkers are supposed to propagate the signal to those subworkers. It seems the propagation is not happening, because the logs don't show the signal being received.

prioux:
There was a bug that prevented the stop signal from
propagating: basically if logger was invoked from
the signal handler, it crashed and the rest of the signal
handler code was ignored.

The main worker class already had code to mitigate this,
with @trap_log a variable holding log messages that
are sent at the end. So the subworker use that too now.

### DIFF
--- a/Bourreau/app/models/bourreau_worker.rb
+++ b/Bourreau/app/models/bourreau_worker.rb
@@ -156,7 +156,10 @@ class BourreauWorker < Worker
   # In the subworker it will do nothing.
   def stop_signal_received_callback #:nodoc:
     if @process_task_list_pid
-      worker_log.info "Propagating STOP to subprocess #{@process_task_list_pid}"
+      # NOTE: in a signal handler it seems we can't invoke the logger.
+      # The superclass has @trap_log to capture and log instead.
+      #OLD: worker_log.info "Propagating STOP to subprocess #{@process_task_list_pid}"
+      @trap_log << [ :info, "Propagating STOP to subprocess #{@process_task_list_pid}" ]
       Process.kill('TERM',@process_task_list_pid) rescue nil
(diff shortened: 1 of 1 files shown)


MontrealSergiy:
@prioux How to test it? Any steps to reproduce? Why no unittest?

prioux:
Start a diagnostics task with a setup delay of say 100 seconds. Then you check with 'ps' regularly. When you see the subworker is alive, you send a kill signal to the worker (NOT the subworker itself) and check the logs to see if it's been properly propagated. I suggest setting the worker logs to 'verbose'.

Do this twice: once with the old code, once with the new code.

MontrealSergiy:
another take on ruby's signal handling issue https://www.mikeperham.com/2013/02/23/signal-handling-with-ruby/

MontrealSergiy:
From web UI perspective did not notice any difference - if I kill subworker, task still fails to exit setup state

MontrealSergiy:
never mind... I did it wrong will try again

MontrealSergiy:
so far results with new code are inconclusive, on one test the killed workers respawn, so test task terminate, on other did not (task stuck in queue). I guess might depend on typing speed.

prioux:
@MontrealSergiy Can you please finish reviewing this? It should be obvious that it works or not. Thanks

prioux:
Well, I can't wait any further, so I'm moving forward with this.

PR #839, opened 2019-07-22
Added JSON descriptor for BoutiquesDemo task.

### LINKED ISSUE #828: Create a boutiques descriptor for a template task
The CBRAIN codebase already comes with a few standard tasks in the base package, but we should also include a boutiques descriptor as an example too. Maybe a descriptor that has many of the integration features that CBRAIN support, which could help developers debug them, or at least demonstrate them.

prioux:
This PR includes a single file, a boutiques descriptor for a sample CBRAIN task that runs the UNIX `du` command, with a few options.

The base CBRAIN installation didn't even have a single Boutiques descriptor, so I figured I might provide a demo one.  I wrote it from scratch, if you look at it you can see how elegantly I aligned everything in the JSON file. This will fix #828 .

The task form interface looks like this:

Screen Shot 2019-07-22 at 18 43 30

### DIFF
--- a/BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task_descriptors/boutiques_demo.json
+++ b/BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task_descriptors/boutiques_demo.json
@@ -1,0 +1,103 @@
+{
+    "name":           "BoutiquesDemo",
+    "tool-version":   "5.1.2",
+    "schema-version": "0.5",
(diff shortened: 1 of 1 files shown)


MontrealSergiy:
Agree, good idea (for adding a boutiques example. not so sure for beatifying json files vertical alignment - imho an automated formater or one of standard/current json style guides should be used. Json is machine-oriented format, if needed to be read/edited by a human transpiling to YAML is always possible)

Issue #545, opened 2016-11-04
Add ExitOnForwardFailure to SSH options for control masters


prioux:
SSH control masters for Bourreaux would benefit from using the `ExitOnForwardFailure` option.

crocodoyle:
There's a branch in my fork where I attempted to implement this, but I have no idea how to test it. I don't think it broke anything though; I tested by simply stopping and starting a Bourreau.

https://github.com/crocodoyle/cbrain/commit/23abf2c5bf60042f928023810db329888dbee8b3

prioux:
I looked at it, it's wrong, the option must be set in the proper place in the code, in `ssh_shared_options()`. Also your code is potentially buggy (you didn't leave spaces around your option, it could get concatenated with previous or next arguments)

prioux:
We will postpone this feature as the option is not supported on all versions of `ssh`. That is the case on our old CentOS servers, but it might be a common situation out there depending on when the SSH dev team introduced the option.

prioux:
Let's revisit this in 2020.

prioux:
I just couldn't wait for 2020.

prioux:
Fixed by #840 

Note: the check will only work in a strict IPv4 env, it seems when IPv6 is also available, SSH is still able to 'listen' to the port even when another process has allocated it, and so doesn't exit.

PR #840, opened 2019-07-23
SSH masters now fail if tunnels can't be opened.

prioux:
This requires the ssh option ExitOnForwardFailure which by now
is supported on most UNIX systems.

Also:
* slight improvments to CustomFilters create()
* scratch userfiles get logging info when created/updated

### DIFF
--- a/BrainPortal/app/controllers/custom_filters_controller.rb
+++ b/BrainPortal/app/controllers/custom_filters_controller.rb
@@ -56,13 +56,7 @@ class CustomFiltersController < ApplicationController
 
     filter_class   = Class.const_get(filter_param)
     @custom_filter = filter_class.new(custom_filter_params(filter_class))
-
     @custom_filter.user_id = current_user.id
-
-
-    if @custom_filter.errors.empty?
-      flash[:notice] = "Filter successfully created."
-    end
 
     respond_to do |format|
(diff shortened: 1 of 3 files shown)


prioux:
This will fix #545

prioux:
To test this:

1) Make sure all your Bourreaux are OFF
2) Find the ID of a Bourreau, any of them, that you want to start. Call this `i`
3) Make the difficult calculation `3090 + i`. Call this `p`
4) Start the command `nc localhost {p}` where p is that value
4) Try to start the bourreau; in the codebase BEFORE this patch it will kind of work but still fail, in the codebase after this patch it won't

You don't even need a portal to test it all, you can use the console:

```ruby
b = Bourreau.find(2)
m=b.ssh_master
# here start the nc command in a separate window
m.start
m.stop (to clean up between tests)
```

prioux:
Ah never mind, I just discovered that because we establish our tunnels without specifying a binding address, the ssh option `ExitOnForwardFailure` doesn't get triggered... :-( I'll fix this next week.

prioux:
Well, the option works, as long as only IPv4 is involved. I'll let the code keep its traditional behavior when IPv6 is present (no failure, just like it was in the past).

PR #843, opened 2019-07-24
Cleaned up many forms (mostly field_explanations)

prioux:
There are no issues associated with the PR.

The view code was still full of dead code and badly placed DIVs instead of SPANs. Cleaned up a bunch of them.

### DIFF
--- a/BrainPortal/app/views/bourreaux/new.html.erb
+++ b/BrainPortal/app/views/bourreaux/new.html.erb
@@ -31,22 +31,23 @@
 
 <%= form_for @bourreau, :as => :bourreau, :url => { :action => "create" }, :datatype => "script" do |f| -%>
 
-  <div class="generalbox">
-      <p><%= f.label :name %><br/>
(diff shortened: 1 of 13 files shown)


prioux:
Note: before reviewing wait a bit, I just discovered that SPANs cannot be given a maximum-width.

prioux:
Hi @candicecz well `display: flex;` doesn't seem to work. It starts columnizing my text.

Here are three screenshots, one with the original DIV elements, then with my SPANs, then with your FLEX. I want the same look as the original one, but without the divs. I think I'll just cancel the PR.

Screen Shot 2019-08-30 at 15 37 46

Screen Shot 2019-08-30 at 15 36 11

Screen Shot 2019-08-30 at 15 36 17

prioux:
I am cancelling this PR; I already re-applied the other esthetic changes to the dev branch in this commit: https://github.com/aces/cbrain/commit/dcf764b91ceb7a61c01dc9dfeb6a3b189f06eef8

PR #845, opened 2019-08-01
Adding new SingSquashDataProvider

prioux:
This is a new data provider class that can connect to a set of squashfs files and browse and extract files out of it. This DP is read-only.

Requirements for installation:

In the DP's configuration:

a) enter values for `remote_host` and `remote_dir`; if these happen to be local, the DP will automatically act 'smart' and not do its operations through SSH.
b) in the attribute `cloud_storage_client_path_start`, configure the full path of the data root inside the container; this woudl be what you prepared in your squashfs files (see below).

In a directory configured as `remote_dir` of the DP, put these files:

1) one or several squashfs files (all with .squashfs extensions)
2) a single singularity image named `sing_squashfs.simg`

The image in 2) above can be the simplest linux image you can build, as long as it has the 'rsync' command inside it.

### DIFF
--- a/Bourreau/app/models/sing_squashfs_data_provider.rb
+++ b/Bourreau/app/models/sing_squashfs_data_provider.rb
@@ -1,0 +1,1 @@
+../../../BrainPortal/app/models/sing_squashfs_data_provider.rb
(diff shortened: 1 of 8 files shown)


prioux:
@shots47s I made all the improvements we talked about.

shots47s:
Ok, it all works now, when I browse, I am given a series of squashfs files.  How do I register those as a real FileCollection?

prioux:
You're joking right? You're not supposed to see the squashfs files...

Issue #749, opened 2018-07-04
Bourreau Workers can die if a task object is destroyed while active


prioux:
Not a big problem, but it woudl be nice if a worker could just handle the case of a disappearing task object.

```
2018-07-04 11:56:13 FATAL SubWorker-94426: Some object has disappeared while processing task 'Cedar/566697': ActiveRecord::RecordNotFound Couldn't find CbrainTask::Civet with 'id'=566697 [WHERE `cbrain_tasks`.`type` IN ('CbrainTask::Civet')]
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/relation/finder_methods.rb:353:in `raise_record_not_found_exception!'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/relation/finder_methods.rb:479:in `find_one'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/relation/finder_methods.rb:458:in `find_with_ids'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/relation/finder_methods.rb:66:in `find'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/querying.rb:3:in `find'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/core.rb:151:in `find'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/persistence.rb:448:in `block in reload'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/scoping/default.rb:35:in `block in unscoped'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/relation.rb:351:in `scoping'
/home/prioux/.rvm/gems/ruby-2.4.1@cbrain/gems/activerecord-5.0.6/lib/active_record/scoping/default.rb:35:in `unscoped'
/home/prioux/CBRAIN/Bourreau/lib/cbrain_extensions/active_record_extensions/single_table_inheritance.rb:47:in `unscoped'
```

[...]

PR #846, opened 2019-08-04
Fix for issue #749

prioux:
The code originally checked the exception's error
message and was trying to make sure it contained
something like:

`Couldn't find CbrainTask::Abcd with id=1234`

The problem is that in the message is really:

`Couldn't find CbrainTask::Abcd with 'id'=1234`

Not sure if it's something that was changed in Rails.
Anyway, I made the check a bit less strict.


prioux:
Link for issue #749

PR #860, opened 2019-08-30
Fixes to CSRF check when POST to userfiles download

### LINKED ISSUE #858: Userfile download action checks CSRF token even with API requests
The download() action is special in that

1) it is a `POST` action
2) it receives JSON as well as www-urlencoded forms
3) it sends back `application/octet-stream` (or at least Accepts it)

This means Rails will apply its CSRF checks on it. That check needs to be disabled when the request comes from an API.

I already have the code to fix all that, I will make a PR tomorrow.

prioux:
Basically, when it's an API request, we don't do the check.

This will fix #858 .


prioux:
To test this:

1) Try to trigger the bug before the PR: Use the API or curl to POST to /userfiles/download with a header `Accept: application/octet-stream` and verify that rails raise an exception about the CSRF token
2) Switch to this branch then try again, the content of your file(s) should appear.

prioux:
This is the result of a bug discovered by Arna.

prioux:
I'm adding some more API checks, stay tuned.

prioux:
Okay so in Travis we get a weird failure of the API test for binary download. It doesn't happen on my Mac. Will investigate next week.

PR #861, opened 2019-08-30
Boutiques: Fix for inputs 'multi' with dropdown

### LINKED ISSUE #848: Boutiques integration not working for with list=true and a select set of options
The following Boutiques descriptor implement fails to produce a useful UI element in CBRAIN:

```
       {
            "command-line-flag": "--steps",
            "description": "Longitudinal pipeline steps to run.",
            "id": "steps",
            "list": true,
            "name": "steps",
            "optional": true,
            "type": "String",
            "value-choices": [
                "cross-sectional",
                "template",
                "longitudinal"
            ],
            "value-key": "[STEPS]"
        },
```

This should produce a list with each element being a select box to set.  Right now one only gets a single select box and then if it is set, the task receives a single value and fails to validate against the schema.

image

Fails in the task with the following error:

```
[2019-08-08 11:30:56 EDT] submit_cluster_job() Launching job on cluster.
[2019-08-08 11:31:23 EDT] cluster_commands() Container image prepared: "CbrainTask::BIDSAppFreesurferReconAll.simg" (ID=2032289)
[2019-08-08 11:31:23 EDT] cluster_commands() Running bosh simulation invocation
[2019-08-08 11:31:27 EDT] cluster_commands() Simulation command failed: got status 99 and STDERR:
[2019-08-08 11:31:27 EDT] cluster_commands() [ ERROR ] u'cross-sectional' is not of type 'array'

Failed validating 'type' in schema['properties'][u'steps']:
[2019-08-08 11:31:27 EDT] setup_and_submit_job() Exception raised while setting up: CbrainError: Simulate command failed
```

prioux:
When list is 'true' in a dropdown with a provided list of acceptable value.

Problem still: can't select more than one value. For that we need to
rewrite (or add a new) dropdown widget.

This will partially fix #848 .

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
@@ -319,7 +319,8 @@
   <%- enum_vals = param["value-choices"] %>
   <%- if not enum_vals.nil? -%>
         # Enum value dropdown
-        dropdown.(id, id,
+        name = id <%= list ? '+ "[]"' : '' %>
+        dropdown.(id, name,
           <%= enum_vals.map { |v| [v, v] } %>,
           optional: <%= opt %>,
           default:  <%= "\"#{defaults.find { |d| d['id']==id }['default-value']}\"" rescue "nil" %>
--- a/script/update_cb_all.sh
+++ b/script/update_cb_all.sh
@@ -42,7 +42,7 @@ A) for both Bourreau and BrainPortal:
 B) for BrainPortal only:
 
   - rake db:migrate
-  - rake db:sanity:check 
+  - rake db:sanity:check
   - rake assets:precompile
   - chmod -R a+rX BrainPortal/public
 


prioux:
@shots47 The whole code is in `BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb`

The standard dropdown is a ruby lambda with a layout defined in line 222-255 ; the HTML has class elements that inherit JQuery behavior defined in lines 435-end; in particular the dropdown code is in several function at around lines 783-819.

The ideal woudl be to leave most of that code alone and duplicate it to implement a 'multiselect_dropdown'. In appearance it woudl be the same except multiple entries in the drop down can be clicked and they would be shown in the text box, space-separated. Select and deselect woudl be supported. The problem is that we need on HTML INPUT element per selected entry, so you'd have to add/remove those hidden INPUT elements. Their `name` parameter must be according to the rails convention, something that ends with `[]`, e.g. `steps[]`.

shots47s:
Thanks, I will take a look

PR #864, opened 2019-09-13
Tool info

natacha-beck:
Add information about Boutiques in 'Tool Version' table.

### DIFF
--- a/BrainPortal/app/views/tool_configs/_tool_configs_table.html.erb
+++ b/BrainPortal/app/views/tool_configs/_tool_configs_table.html.erb
@@ -113,6 +113,7 @@
       tc_basename = tc_self_update.basename
       tc_commit   = tc_self_update.short_commit
 
+      # tc_basename match *.json if cbrain use a Boutiques descriptor
       "#{tc_commit} -- #{tc_basename}" if tc_basename =~ /\.json$/
     }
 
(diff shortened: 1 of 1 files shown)


MontrealSergiy:
good job, do you think we need also detect bids tool? (maybe in futures, I do not think we have many yet)

natacha-beck:
@MontrealSergiy did you have any example in mind?

MontrealSergiy:
https://portal.cbrain.mcgill.ca/tool_configs/1153

BrainPortal/app/views/tool_configs/_tool_configs_table.html.erb
```diff
@@ -105,6 +105,18 @@
       :filters  => default_filters_for(@base_scope, :ncpus)
     ) { |tc| tc.ncpus.to_s }
 
+    t.column("Boutiques", :boutiques,
+    ) {|tc|
+      tc_self_update = tc.tool.cbrain_task_class.revision_info.self_update rescue nil
```
prioux: Two comments.

1) the `next if` is wrong, you're not in an iterator at all, and conceptually you're supposed to always return something from that block; if you don't want o display anything, have the block implicitely return the empty string

2) `self_update` is a verb, and a mutator method too. The value you're fecthing is a RevisionInfo object. So the code should really be

```ruby
rev_info = tc.tool.cbrain_task_class.revision_info
rev_info.self_update
# At this point the rev_info object has its content filled in.
```

prioux: Also the revision info you're getting is not from the ToolConfig, it's from the task class....

PR #865, opened 2019-09-14
Added resource usage tracking.

prioux:
We have a now a generic structure for recording resource usage changes.
It is an ActiveRecord of class ResourceUsage, with several subclasses:

```
  ResourceUsage (abstract)
  |
  +- SpaceResourceUsage (abstract)
  |   |
  |   +- SpaceResourceUsageForCbrainTask
  |   |
  |   +- SpaceResourceUsageForUserfile
  |
  +- TimeResourceUsage (abstract)
      |
      +- CputimeResourceUsageForCbrainTask
      |
      +- WalltimeResourceUsageForCbrainTask
```

* Whenever a file's size is updated, one SpaceResourceUsageForUserfile
  is created.

* Whenever a task transitions to 'Data Ready', one
  WalltimeResourceUsageForCbrainTask will be created and one
  CputimeResourceUsageForCbrainTask will be created.

* Whenever a task reach a final state, one SpaceResourceUsageForCbrainTask
  record will be created.

All records contain optional associations to the userfiles, tasks,
tools, tool_configs etc. For all of these assocations, because we
want to keep the usage information even after the associated abject
is deleted, we also keep a full copy of some attributes. E.g. for
a task, we keep its type, its status, its tool name, etc.

For the moment, the migration doesn't include the creation of any indexes.

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -2611,7 +2611,7 @@ chmod o+x . .. ../.. ../../..
     false
   end
 
-  protected
+  public # the callbacks below are handled by after_status_transtion()
 
   # Add up CPU usage when a task goes to "Data Ready" state
   def track_resource_usage_cpu(prevstate) #:nodoc:
@@ -2621,6 +2621,8 @@ chmod o+x . .. ../.. ../../..
     qsub_file = Pathname.new(task_workdir) + qsub_stdout_basename
     num_seconds_info = extract_cpu_times_from_qsub_wrapper(qsub_file)
     return unless num_seconds_info.present?
+
+    #TODO sacct -n -p -o JobID,UserCPU,CPUTime,TotalCPU,CPUTimeRAW,Elapsed -j 2896567
 
     cpu_time  = num_seconds_info[:user_tot] + num_seconds_info[:syst_tot]
 
@@ -2651,7 +2653,7 @@ chmod o+x . .. ../.. ../../..
   # a final status.
   def track_resource_usage_final(prevstate) #:nodoc:
     SpaceResourceUsageForCbrainTask.create(
-      :value              => self.cluster_workdir_size, # just FYI
+      :value              => self.cluster_workdir_size || 0, # just FYI
       :user_id            => self.user_id,
       :group_id           => self.group_id,
       :cbrain_task_id     => self.id,
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1195,7 +1195,7 @@ class ClusterTask < CbrainTask
     return nil unless File.exists?(filename.to_s)
     content = File.read(filename.to_s).tr("\n"," ")
     return nil if content !~
-       /After\s(\d+)\sseconds
+       /CBRAIN\sTask\sEnding\sWith\sStatus.*After\s(\d+)\sseconds
         .*
         CBRAIN\sTask\sCPU\sTimes\sStart
         \s+
@@ -1827,7 +1827,7 @@ date "+CBRAIN Task Ending With Status $status After $SECONDS seconds, at %s : %F
 
 echo ''
 echo 'CBRAIN Task CPU Times Start'
-times
+times # this is very important and the output is captured and parsed
 echo 'CBRAIN Task CPU Times End'
 
 echo "CBRAIN Task Exiting"       # checked by framework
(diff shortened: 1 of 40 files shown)


prioux:
Question for you all: is there anything else we need to track permanently? For reports, grant proposals, etc.

prioux:
My next pull request will be a report generator so an admin user can query the table and obtain numbers (e.g. how many CIVET tasks were terminated between date A and B?)

BrainPortal/app/models/resource_usage.rb
```diff
@@ -0,0 +1,91 @@
...
+  belongs_to            :data_provider,   optional: true
+  belongs_to            :cbrain_task,     optional: true
+  belongs_to            :remote_resource, optional: true
+  belongs_to            :tool,            optional: true
+  belongs_to            :tool_config,     optional: true
+
+  before_save :record_names_and_types
+
+  # If any of the "_id" attributes are provided,
+  # the we also automatically fill in the duplicated
```
natacha-beck:
```suggestion
  # then we also automatically fill in the duplicated
```

prioux:
Yep, I saw that typo (and others too), will correct in a future commit

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2562,6 +2611,56 @@ def task_is_proper_subclass #:nodoc:
...
+      :cbrain_task_id     => self.id,
+      :remote_resource_id => self.bourreau_id,
+      :tool_id            => self.tool.id,
+      :tool_config_id     => self.tool_config_id,
+    )
+  end
+
+  # Add a task workdir size and the status of a task that reach
+  # a final status.
+  def track_resource_usage_final(prevstate) #:nodoc:
```
natacha-beck: Should be a public method

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2562,6 +2611,56 @@ def task_is_proper_subclass #:nodoc:
     false
   end
 
+  protected
+
+  # Add up CPU usage when a task goes to "Data Ready" state
+  def track_resource_usage_cpu(prevstate) #:nodoc:
```
natacha-beck: Should be a public method.

prioux:
Please stand by, adding new features.

prioux:
Don't wanna brag, but even after I remove all the copyright headers, I still have 900 new lines in this PR. And I started it Friday afternoon. OK I'm bragging now.

natacha-beck:
Let me know when I can test it.

prioux:
@natacha-beck I added my final commit just 1 minute ago. As soon as you see it has passed Travis, start testing as much as you want!

BrainPortal/app/views/resource_usage/_resource_usage_table.html.erb
```diff
@@ -0,0 +1,297 @@
...
+    is not garanteed to be accurate, as this is a expensive resource to compute
+    and is only provided FYI. The main purpose of this report is to compare the
+    final <strong>status</strong>.
+  <% end %>
+  <p class="medium_paragraphs">
+  Columns labeled <strong>Cached</strong> record pieces of information
+  about resources <em>as they were at the time the record was made</em>.
+  If a resource was destroyed (e.g. a non-<strong>Cached</strong> column is empty), these stay behind
+  and provide filtering options. It makes it possible to compute resource usage reports for
+  deleted resources.
```
natacha-beck: report

I played with the interface. It is not clear that the report takes into account the filter applied to the table.

BrainPortal/app/views/tasks/_resource_usage.html.erb
```diff
@@ -0,0 +1,44 @@
...
+          <% else %>
+            <td></td>
+          <% end %>
+
+          <% if type =~ /Disk/ %>
+            <td><%= pretty_resource_usage_value(ru) %></td>
+          <% else %>
+            <td></td>
+          <% end %>
+
```
natacha-beck: Why did we have 2 similar if/else block?

prioux: it's for the alignment in the table. So we get:

| [cputime] | emptycell |
| emptycell | [diskspace] |

BrainPortal/app/views/resource_usage/_resource_usage_table.html.erb
prioux: Can you be more specific? Please give an exact example?

BrainPortal/app/views/tasks/_resource_usage.html.erb
natacha-beck: Thanks for the explanation.

BrainPortal/app/views/resource_usage/_resource_usage_table.html.erb
prioux: By design, whatever filer you apply sticks around between reports. That's so you can say for example "I want to see user=abc and task_type=xyz" and switch between reports with these two filters applied.

natacha-beck: report1

If I have no filter apply to my table I get: 
> 6 usage records: + 25.6 Kb - 319.5 Mb (Total: - 319.5 Mb) (Average: - 53.2 Mb) 
And if I have `Remote resource` NaN selected I get: 
> 1 usage record: + 0 bytes - 8.5 Kb

For me, it just not clear that the report applies the table filter.

prioux: I don't get it but we'll talk about this when we meet at the office...

prioux: Got the explanation in person, will work on it now.

Issue #866, opened 2019-09-16
Add JSON model support as a base userfile feature


prioux:
I think it would be nice.

Also, since browsers have built-in JSON and XML viewers, we should create CBRAIN viewers that send the raw content to  the browser.

prioux:
I'm already done part 1

prioux:
Implemented in #868

PR #868, opened 2019-09-18
Added JSON file and inline content-disposition

prioux:
This add support for JSON files in CBRAIN.

Also both XML and JSON files can be viewed in the browser's own viewer (Firefox supports both, Chrome only seems to support XML).

### DIFF
--- a/BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/json_file/views/_json_raw.html.erb
+++ b/BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/json_file/views/_json_raw.html.erb
@@ -5,5 +5,5 @@ Some modern browsers have a built-in JSON viewer.
 <p>
 
 <%= link_to "Browser-based viewer for JSON content",
-    content_userfile_path(@userfile.id, :disposition => 'inline') %>
+    content_userfile_path(@userfile.id, :disposition => 'inline'), :target => :_blank %>
 
--- a/BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/json_file/views/_json_raw.html.erb
+++ b/BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/json_file/views/_json_raw.html.erb
@@ -1,0 +1,9 @@
+<p>
+The link below will open this JSON file directly in your browser.
+<p>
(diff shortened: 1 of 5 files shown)


prioux:
This will fix #866 by the way.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/json_file/views/_json_raw.html.erb
```diff
@@ -0,0 +1,9 @@
+<p>
+The link below will open this JSON file directly in your browser.
+<p>
+Some modern browsers have a built-in JSON viewer.
+<p>
+
+<%= link_to "Browser-based viewer for JSON content",
```
natacha-beck: I personally prefer that the link opens in a new tab, and it is what it´s done for raw Text file for example. We should decide which behaviour is better.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/xml_file/views/_xml_raw.html.erb
```diff
@@ -0,0 +1,9 @@
+<p>
+The link below will open this XML file directly in your browser.
+<p>
+Some modern browsers have a built-in XML viewer.
+<p>
+
+<%= link_to "Browser-based viewer for XML content",
```
natacha-beck: Same comment as before regarding, maybe it should open in a new tab.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/json_file/views/_json_raw.html.erb
prioux: I didn't think of it, but it's a good idea. It should be easy to do...

natacha-beck: yes  `:target => "_blank"` and that's it.

prioux: I'll do it tomorrow.

prioux:
@natacha-beck Added the target=_blank; Travis CI passed.

Issue #333, opened 2016-02-13
Show tasks 'raw' parameters in a more beautiful way


prioux:
The 'raw' tab for the params could use a format that's different than YAML. Something prettier, yet still raw.

prioux:
Fixed in #869 (more or less).

Issue #867, opened 2019-09-18
Add optional cpu busy loops (user and system) to Diagnostics task


prioux:
The diagnostics tasks already has several places where delays (sleep) can be configured, but I would like to add two new optional params, for doing N seconds of busy looping (user side only, system side only).

I was pretty successful in getting busy system-side loops but reading 65K chunks from /dev/urandom in this test perl script I was using for slurm. These loops could be implemented as one-liners in the diagnostic's task bash script.

```perl
my $PURE   = 120;
my $SYSTEM = 120;

print "Pure CPU\n";

my $start = time;
while ((time - $start) < $PURE) {
  my $x = 0;
  for(my $i=0;$i < 1_000_000;$i++) {
    $x += $i;
    $x -= ($i - 1);
  }
}

print "System CPU\n";

$start = time;
my $buffer = "x" x 65536;
while ((time - $start) < $SYSTEM) {
  open(READ,"</dev/urandom");
  read(READ,$buffer,65536);
  close(READ);
}
```

prioux:
Implemented in #869

PR #869, opened 2019-09-18
Added options for busy loops in diagnostics.

prioux:
This will implement #867 .

Works on Linux and macOS (yep, macOS has `/dev/urandom`).


prioux:
This will also, more or less, fix #333

prioux:
Two birds with one stone.

PR #878, opened 2019-10-09
Remove new in UserfilesController

### LINKED ISSUE #735: UserfilesController#new is no longer used
When in 2016 the upload form was transformed from a standalone page to a overlay in the userfiles index page, the controller action `new` was rendered useless.

If found this bug because I wanted to add a restriction to which data providers can be used for uploading, and the mod there did nothing; instead the list of DPs for upload is determined in `views/userfiles/_dialogs.html.erb` (yuk!)

That dialog partial needs to get its list of DPs from a controller action...

natacha-beck:
Close #735

### DIFF
--- a/BrainPortal/app/helpers/select_box_helper.rb
+++ b/BrainPortal/app/helpers/select_box_helper.rb
@@ -592,5 +592,24 @@ module SelectBoxHelper
     ordered_by_lock_status
   end
 
+  def writable_data_providers(user=current_user) #:nodoc:
+
+    writable_data_providers =
+      DataProvider
+        .find_all_accessible_by_user(current_user)
+        .all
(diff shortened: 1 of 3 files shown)


BrainPortal/app/helpers/select_box_helper.rb
```diff
@@ -592,5 +592,24 @@ def regroup_users_by_lock_status(users) #:nodoc:
     ordered_by_lock_status
   end
 
+  def writable_data_providers(user=current_user) #:nodoc:
```
prioux: If this is a helper method then it shouldn’t be :nodoc: ; please provide a few short lines of documentation about what the method returns. Also for the other method below

Issue #855, opened 2019-08-15
Publish to Zenodo button for exporting the full task to Zenodo for provenance capture


shots47s:
Thought I had created this, but did not.

Feature would be to give the user the ability for a completed Boutiques task to publish enough information to Zenodo to reproduce the result of the pipeline in CBRAIN (and potentially other platforms).

Current list of things that need to be published would be at minimum:

1) The Singularity Image the pipeline used to run.
2) The Boutique descriptor that was used to define the pipeline.
3) The input parameters or invocation for the pipeline.
4) The input data (potentially sensitive, so need to make optional, at least then some representation so that we can get it back from CBRAIN if needed).
5) The output data (potentially sensitive, so need to make optional, at least then some representation so that we can get it back from CBRAIN if needed).
6) A JSON capturing the information that is related to the execution that may cause the singularity image to have an altered execution on different computational hardware (tricky).

There would be a button that automates this and provides the user a new interface component to control the process including the option to publish to Zenodo as "Closed", "Open", "Registered", or "Embargoed".  

Each component should be uploaded separately and uniquely identified with a DOI.

Conversely, we could provide an import from Zenodo button that gives the user the ability to re-import the information into CBRAIN. 

All of this should have a "caching" mentality, i.e. if the data is already where it needs to go (i.e. the Singularity image is already in Zenodo), then the existing data is used.

MontrealSergiy:
I guess for now button goes on task (or task list).

I guess ideally the cache of singularity image should be controlled by cache sum or smth.
Also caching leaves possibility one user decision to unpubish task  deleteing tool image shared by several tasks.

shots47s:
For Zenodo API Key, we make these a part of the user profile.

glatard:
It would be great if points 2, 3, and perhaps 5 could be compatible with what is tracked and published by `bosh data`. The PR in https://github.com/boutiques/boutiques/pull/532/files adds the possibility to append an arbitrary object to it.

prioux:
Any idea how we can handle gracefully entire large task arrays? E.g. civet over 1000 subjects, or a bids app with a bids dataset with 1000 subject? Each individual task contain only the params for that task (and doesn’t know it is part of a larger whole). Also, possibly, individual task might have been reatarted with slightly different params (eg because of a failure with the standard params), so we can’t assume all of them have identical params.

We also don’t want to upload the same input dataset 1000 times.

[...]

PR #880, opened 2019-10-15
Zenodo publishing

prioux:
This is operational, provided a task has been extended with the two methods that CBRAIN expects to create the Zenodo records. Two base CBRAIN tasks have this capability: Diagnostics and BashScriptor.

Soon to be added: automatic support for Boutiques tasks (in a separate PR?)

This will implement part of the requirements for #855 

Note: while testing it seems that Zenodo just refused all large files (greater than a few dozen megabytes).


prioux:
Addendum:

The three commits I made today added zenodo support to most (all?) Boutiques tasks. Also, we now record and publish a lot more information about the runtime environment of the task.

To test with CIVET or ReconAll (traditional tasks) you'll need to deploy the PR for the cbrain-plugins-neuro : https://github.com/aces/cbrain-plugins-neuro/pull/146

BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -771,6 +771,395 @@ def apply_operation(operation, taskids, do_in_spawn = false, options = {})
...
+    # with another DOI are just refered to as such.
+    related = @zenodo_userfiles
+      .select { |u| u.zenodo_doi.present? }
+      .map    { |u| ZenodoClient::RelatedIdentifier.new(:identifier => u.zenodo_doi, :relation => 'hasPart') }
+    if related.present?
+      @zenodo_metadata.related_identifiers ||= []
+      @zenodo_metadata.related_identifiers  += related
+    end
+
+    # Validate here; TODO
```
natacha-beck: Should be done or remove no ?

BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -771,6 +771,395 @@ def apply_operation(operation, taskids, do_in_spawn = false, options = {})
...
+  end
+
+  def init_zenodo_client(zsite) #:nodoc:
+    zsite  = zsite == 'main' ? 'main' : 'sandbox' # sanitize
+    token  = token_by_zsite(zsite)
+    server = zenodo_server_by_zsite(zsite)
+    ZenodoClient.configure do |config|
+      config.api_key['access_token'] = token
+      config.host                    = server
+      #config.debugging               = true # will print info on stdout; WARNING binary junk!
```
natacha-beck: Could be removed?

BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -771,6 +771,395 @@ def apply_operation(operation, taskids, do_in_spawn = false, options = {})
...
+    is_col = userfile.is_a?(FileCollection)
+
+    # Identify the content
+    userfile.sync_to_cache
+    content_path = userfile.cache_full_path
+    content_name = userfile.name
+    content_path = create_tmp_tar_for_filecollection(userfile) if is_col
+    content_name = userfile.name + ".tar.gz"                   if is_col
+
+    # Upload
```
natacha-beck: Code duplication from 1105 to 1115 and from 1133 to 1143.

BrainPortal/app/helpers/zenodo_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  ZenodoSandboxDOIPrefix = "10.5072/"
+  ZenodoMainDOIPrefix    = "10.5281/"
+
+  # Creates a pretty link to a DOI, usually for Zenodo
+  def link_to_zenodo_doi(doi)
+    link_to "<img src=\"https://zenodo.org/badge/DOI/#{doi}.svg\" alt=\"DOI\">".html_safe,
+            "https://doi.org/#{doi}", :target => '_blank'
+    #("<a href=\"https://doi.org/#{doi}\">" +
```
natacha-beck: Is it useful?

BrainPortal/app/controllers/tasks_controller.rb
prioux: Agreed. I'll add some validations, at least one.

BrainPortal/app/controllers/tasks_controller.rb
prioux: No, this is really useful, and someone who doesn't know how the ZenodoClient library works would probably be happy to know how to turn on debugging. I was using it all the time while developing, and I can imagine wanting to uncomment it when doing future work.

BrainPortal/app/controllers/tasks_controller.rb
prioux: Yes, but I'm not sure how to handle refactor it... will look into it.

BrainPortal/app/helpers/zenodo_helper.rb
prioux: Nope, I'll remove that, thanks.

PR #887, opened 2019-10-25
Signup

natacha-beck:


BrainPortal/app/models/signup.rb
```diff
@@ -60,15 +60,11 @@ def generate_token
...
+  def full_name #:nodoc:
     "#{title} #{first} #{middle} #{last}".strip.gsub(/  +/, " ")
   end
 
-  def user_full #:nodoc:
-    "#{first} #{last}".strip.gsub(/  +/, " ")
-  end
-
-  alias full_name full
+  # alias full_name full
```
prioux: Why leave the alias commented out? Just delete it. Not only that, but if the alas was re-insterted now, it would crash, because there is no longer a method called `full`. So the alias itself has been made invalid by your code changes.

BrainPortal/app/views/signups/new.html.erb
```diff
@@ -80,7 +80,7 @@
     <tr>
       <th><%= f.label(:position, "* Current position or role:") %></th>
       <td><%= f.select :position, options_for_select(
-              [ [ "(select one)", "" ], "Faculty", "PostDoc", "Researcher", "Student", "Industry", "Other" ],
+              [ [ "(select one)", "" ], "Undergraduate Student", "Postgraduate Student", "Post Doctoral Student", "Faculty", "Researcher", "Staff", "Industry", "Other" ],
```
prioux: I like the new names, but why change the order? Put Faculty first and then go down the list of academic levels (including just "student", not everybody is at a university or college). I'll come up wityh a better list in a follow up comment.

BrainPortal/app/models/signup.rb
prioux: Also `BUG!` You forgot to change the method call in `BrainPortal/app/views/cbrain_mailer/signup_notify_admin.text.erb`

BrainPortal/db/migrate/20191025141308_add_position_to_users.rb
```diff
@@ -0,0 +1,20 @@
+class AddPositionToUsers < ActiveRecord::Migration[5.0]
+
+  def up
+    add_column :users, :position, :string, :after => :full_name
+
+    User.all.each do |user|
```
prioux: I'm not getting any `position` information transfered to the User records, not sure why. All I get is 'Unknown' for everything.

BrainPortal/db/migrate/20191025141308_add_position_to_users.rb
```diff
@@ -0,0 +1,20 @@
...
+      user.position = extract_position_from_log(user.getlog || "") || "Unknown"
+      user.save
+    end
+  end
+
+  def extract_position_from_log(log)
+    log[/Position:\s*(\.+?)\s*\z/, 1]
```
prioux: OK I found the error: I made the mistake when I created this regex. It should be:

```ruby
  log[/Position:\s*(.+?)\s*$/m, 1]
```

There are three differences: no backslash before the period (duh), use `$` instead of `\z` (double duh) and use the multiline modifier `m` (optional but sensical).

Issue #786, opened 2018-12-04
Cookie affects the API (it should not)


MontrealSergiy:
According to Swagger CBRAIN API should rely on authentification token, not cookies. 
Yet presently POST session/?user=...?pwd returns token according to the cookie if any rather than the token.

I observed the error with Postman API testing tool, which stores the cookies for more realistic testing.
Probably the same can be observed using any browser-based API testing tool  and other API endpoints.

The issue might confuse users who use several logins for different projects (e.g. depending on affiliations) or share same PC.

prioux:
I don't understand what is to fix here... if a program has a token, CBRAIN will use it to verify authentication. If cookies are exchanged, that becomes a fallback (what browsers do). Does the API in the cbrain-client gem ceased to work? Do browsers cease to work? Can you please give me an exact reproducible example of a problem?

prioux:
@MontrealSergiy Please explain precisely and give me an example, like I asked.

MontrealSergiy:
Generally speaking, REST API's are supposed to be 'stateless' https://restfulapi.net/statelessness/ .

meanwhile I am trying to make up more clear scenario soon.

prioux:
I know what REST is and the CBRAIN API has never had a requirement to be totally restful. In fact we don't want it to be.

MontrealSergiy:
Ok, if it is on purpose I do not mind cancelling the ticket.

Looks like there are a few big name api's with cookies support.

If you have time I can show you some thing from my desktop
but I do not have a good reproducible scenario so far.

MontrealSergiy:
Do you think two different 3-rd party apps can use cbrain api on same browser with no issues?

prioux:
Probably not, but one way to be sure is to disable sending back the cookie whenever we are working with API requests. It should be pretty simple.

prioux:
I wrote the code by before committing it I wanted to check how oftne API requests generate cookies. Adding the "-c" option to curl revealed that NONE of the API request generate a cookie! The only exception was when testing the specific `users/badnormpatch` test, and I'm not sure why either.

MontrealSergiy:
For me /session?login=username&password=wrongpassword generates a cookie even if password is wrong

prioux:
Actually I'm getting a cookie for all PATCH and POST and PUT

prioux:
I made a commit, if TRAVIS CI is satisfied I'll create a PR.

[...]

PR #888, opened 2019-10-25
API requests never send or accept cookies.

prioux:


prioux:
@MontrealSergiy Please pull and test before and after. including with the API tests.

prioux:
This will fix #786

prioux:
I suspect this PR can affect negatively AJAX, JS or JSON requests made by the web interface, if such requests expect session information to be updated. Please test before and after for several such requests (you can find them by scanning the controllers and finding `format.js`)

MontrealSergiy:
While works as intended in simple scenarios, it is not exactly "never" no cookies 100%. If previously created,  before the suggested change cookie is provided, a new cookie might be generated. Even if cookie is from different environment, it will be updated (though request would fail in 401). Not sure is thee issue or benign particularity, as after a while old cookies will expire. But cookies are not 100% ignored. Better than before anyways.

Steps to reproduce: 
0. create user, say called `goodusername` using GUI, console or recycle an existing account.
1. run old code to generate a cookie for a user. `baseurl/session?login=goodusername&password=goodpwd` 
2. Update to new branch.
3. Create session using wrong password but above cookie `baseurl/session?login=goodusername&password=badpwd`
Observed: 200, both json and a new cookie are created
Expected: requires fails with no new cookies returns.

prioux:
I don't understand. User's aren't logging in from the web form when using the API. It's normal for actions triggered by the web browser to generate cookies. We even want that.

Please check that:

* in a series of actions using PURELY the API, cookies aren't created and sent back, or needed.
* in a series of actions using purely the web interface, the new code does not introduce new bugs, e.g. where the session is not updated properly because an action was sent with a JS or JSON content type and the action was expected to change session info

If you find any counter examples of any of these two scenarios, document them.

prioux:
And if you do find counterexamples, give us the test code samples that we can run directly.

MontrealSergiy:
Note, in the example above, login is from API. However, creation of user is not. Use case existing CBRAIN users can benefit from a 3rd party tool that compliments/improves/ CBRAIN features e.g. with sophisticated UX, data sets, visualization etc.

MontrealSergiy:
Disregard above example. Works as intended.

prioux:
@MontrealSergiy Have you tried the make sure the interface works properly when session params are changed? In particular I suspect the group_controller#switch action won't work anymore, can you verify please?

MontrealSergiy:
sure

MontrealSergiy:
It seems not this PR issues, dev branch has same problem. Production is fine. Will double check.

MontrealSergiy:
It seems like something related to test mode, development is fine

MontrealSergiy:
Albeit above the Monday issue seems minor, to make sure you understand what I reported on Monday here is more easy to follow steps in bash

from bash run line by line
```cd ~/cbrain/BrainPortal # go to cbrain portal folder to start the server
git stash
git checkout dev
kill -9 $(lsof -i tcp:3000 -t) # kill a (server) process to free port 3000
rake db:seed:test:api 
rails server puma -e test -p 3000 &
curl -v -X  POST   'http://localhost:3000/session?login=norm&password=norm_123'  -H 'Accept: application/json' -c ~/cookies.txt
git checkout api_no_cookies #  or whatever is this new feature branch name in your repo
kill -9 $(lsof -i tcp:3000 -t) # kill the server
rails server puma -e test -p 3000 & 
curl -v -X  POST   'http://localhost:3000/session?snorm_123'  -H 'Accept: application/json' -c ~/new_cookies.txt -b ~/cookies.txt
tail ~/new_cookies.txt```

the file ~/cookies represents the consumed cookies
the file ~/new_cookies.txt - the returned


PS. curiously you can change the password to whatever you want

PPS. If above scenario is too long a shorter one is below

1. set test secret to 
```test:
  secret_key_base: 34da37faeff4c6cbd17d014f2a20dc4f195b2eb1954fe1b92e41ad500797db0ec02540f5c7e50616a383a67577ed9f568bc074ac5bba087b9f63098b4441da9c
```
and start test server

2. run

curl -v -X  POST   'http://localhost:3000/session?snorm_123='  --cookie "BrainPortal5_Session=UEpnVmgvanh6cXZCWW0vaU1BcklnNTJtNDQ1bzc2T2diV200NlpNUEFRVVpsOTVSazJqRERQWG0yZjR3TjlXOStvK3BYNWxFdW5obEhJYThBWm5JZjhNR2NDWnlMTFBaMDBRN0dCb28xbEhpWlpYdEZrYUgwTmhWcnFkNEZNVitQWkR3MWVsZ3hnWS94L2xEcHkzblRHRERHeGJZYW01alBPRkdGeXdtZ3FiUS9BaGM3QnpqYjV3UWRGWjdFZWtzTkJxdVV3S3l3SlBiWEFrM2RmaWZ4WGw5dTBGNTNCR3M5TzI1eU5VeXI0QT0tLUJZRktIM3pWNGozbTNRUHo0anhvL3c9PQ%3D%3D--e7f251a866be47d8bfd1e70dd63483705e038dbc; path=/; domain=localhost; HttpOnly; Expires=Tue, 19 Jan 2038 03:14:07 GMT;"  -H 'Accept: application/json, application/xml'   -H 'Postman-Token: 9900e43f-7d49-4ceb-815d-eea44c3440f3'   -H 'cache-control: no-cache'

prioux:
I checked your action sequences (the one above the PS and PPS), I have some observations. Two of them about the commands themselves (not all that important) and one about the behavior of the code we're trying to evaluate.

About the commands:

1) you mangled the URL in the second CURL request, I assume it is just a cut and paste error and it was supposed to be the exact same one as the first CURL command
2) as a general rule, never use `kill -9`, it never lets a process do its proper tear down (you are left with spurious PID files etc). Myself I just start the rails server in a separate window and hit CTRL-C when I want to stop it, it's more reliable that your lsof trick

About the behavior:

Indeed, we get a cookie in both `cookies.txt` and `new_cookies.txt`, but if you look carefully at the request's replies' header using `-D -` on the curl command-line you'll see that the server does NOT send back a Set-Cookie header; I am pretty certain that teh content of `new_cookies.txt` was created by curl itself just as a way to maintain the type of persistency that a web browser would do (a browser doesn't erase a cookie if a request doesn't send back the cookie every time)

prioux:
No please check the behavior of `groups#switch`

MontrealSergiy:
Yup, switch passes gets 200 status yet `GET {{baseurl}}/userfiles?cbrain_api_token={{auth_norm}} ` return all files,

prioux:
OK, I suspected as much. So I have two solutions:

1) disable the switch action for API requests. This only affects the old API libraries (in Perl and Ruby) because the switch action is not exported in our Swagger API
2) change the active_group_id session attribute so it goes into cbrain_session instead

I'll think about it.

MontrealSergiy:
Third Options: add some (optional) options_n_filters query or json 
paramer to all list objects commands and let frontend decide how to handle that (though might be reinveting the cookies). Switch then is not really needed.

prioux:
@MontrealSergiy Made a fix for switch, try it again please.

MontrealSergiy:
The url in my example was mangled intentionally. (Probably should use something more absurd to make it clear).

MontrealSergiy:
Not sure of the switch but token authentification seems fail.

MontrealSergiy:
Does not work. To reproduce on test data

1) create session for `norm` user `POST {{baseurl}}/session?login={{user}}&password={{pwd}}`
2) switch to branch 3 (`norm`) `{{baseurl}}/groups/switch?cbrain_api_token={{auth_norm}}&id=3` where auth_norm is the token obtained in the step 1
3) `POST {{baseurl}}/userfiles?cbrain_api_token={{auth_norm}}` returns all the files, not just group 3

MontrealSergiy:
switch works now

prioux:
I think we're ok at this point: I checked all the cases of `format.js` in the controller code, none of them seem to store state information in the cookie-based session (we put everything into the DB-backed session store)

PR #896, opened 2019-11-13
Viewer

### LINKED ISSUE #760: Provide feedback to user on visualization when file is too large to visualize
When a file is too large to visualize (i.e. exceeds 400mb, which I am not sure is actually a limit anymore given RAM on computers), the user is just presented either with a message that says it is unable to visualize a file type, or a reduced visualization with no information on what they should do next to view the file.  This should be changed to potentially provide them a link to information on how they can visualize it (i.e. a link to a visualization tool such as Brain Browser or a wiki page on how to download and visualize the file.

natacha-beck:
Close #760


natacha-beck:
Should fix test first.

PR #899, opened 2019-11-15
Added support for path prefix in S3 Flat DP.

### LINKED ISSUE #898: Add support for cloud_storage_client_path_start in S3 Flat DP
Right now it just browses the top of the bucket.

It should be easy to add the constant prefix when creating the `S3Sdkv3Connection` object. Then all its methods would use it.

prioux:
This will fix #898


prioux:
Note that this PR is on MASTER

Issue #900, opened 2019-11-19
Bug in QC panel


natacha-beck:
Bug description: 

- Open the QC panel with multiple NIFTI files for example.
- Click on `previous` on the left panel.
- Will end up with 2 BrainBrowser panel on the right side. 

It's due to redundant id for BrainBrowser panel.

PR #901, opened 2019-11-19
Fixed QC panel bug

natacha-beck:
Fixed for #900


prioux:
I thought we had agreed this PR would include a copy of the patch made in master to fix the javascript code?

prioux:
I'm talking about this: https://github.com/aces/cbrain/commit/b933ae8ed8b39cf11f68693aca318b62e35f5682

prioux:
I am not able to reproduce the original bug... :-( Please see me so we can have a look.

natacha-beck:
> I thought we had agreed this PR would include a copy of the patch made in master to fix the javascript code?

Done.

PR #902, opened 2019-11-21
Improved QC page layout (control section)

prioux:


prioux:
This PR also includes a QC viewer for text files, so it's easy to experiment with them instead of more complex viewers.

Please try it with different browsers and try to break it.

PR #903, opened 2019-11-28
Allow switch between 1 or 2 QC panels

### LINKED ISSUE #553: QC main page should allow the user to only see one QC panel instead of two.
Right now we show two QC panels side by side; it would be easy to add a little close button so that the rightmost panel is discarded and the user is only shown one QC panel, full size. Drop the entire div from the DOM. User shoudl be told there is no going back, however.

natacha-beck:
Close #553

### DIFF
--- a/BrainPortal/public/javascripts/cbrain.js
+++ b/BrainPortal/public/javascripts/cbrain.js
@@ -243,6 +243,18 @@
     });
 
 
+    loaded_element.find(".hide_qc_panel").click(function(event) {
+      var qc_right  = loaded_element.find(".qc_right_panel");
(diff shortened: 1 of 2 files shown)


BrainPortal/public/javascripts/cbrain.js
```diff
@@ -243,6 +243,18 @@
     });
 
 
+    loaded_element.find(".hide_qc_panel").click(function(event) {
```
prioux: Please add a comment section explaining what the new code does; no need to go into details but indicate clearly what part of the user interface is affected an how.

Issue #663, opened 2017-11-15
ToolConfig project selection


glatard:
The project selection drop-down list in ToolConfigs is very difficult to navigate. For instance, it took me several minutes to find the PSOM project. It would be nice if a search box could be added.

![screenshot from 2017-11-15 09-29-32](https://user-images.githubusercontent.com/5174953/32841156-9f87e530-c9e7-11e7-92ba-fc2aad1e5d9f.png)

prioux:
Fixed by #906

Issue #217, opened 2015-09-24
Group selections needs to be less awful (Redmine 3017)


prioux:
(From Tarek, March 2012)

At least for admins. It's just one giant, ugly select box, and you usually need to scroll down to the bottom to get what you want.

prioux:
Ajax dynamic selection box.

crocodoyle:
Fixed in Bootstrapification

prioux:
Duplicate of #663

prioux:
Fixed by #906

PR #906, opened 2019-12-11
Improve long select-box with chosen

natacha-beck:
This PR is related to issues #663 and #217


prioux:
I would prefer it if the the stylesheets, javascript and images were configured under the Rails assets management pipeline (app/assets/*) instead of the public directory.

prioux:
I'm starting to think maybe we should do this to ALL select inputs. E.g. here's the task control section, with two Chosen select and two normal ones:

Screen Shot 2019-12-11 at 14 24 55

natacha-beck:
> I'm starting to think maybe we should do this to ALL select inputs. E.g. here's the task control section, with two Chosen select and two normal ones:
> 
> Screen Shot 2019-12-11 at 14 24 55

Agree.

prioux:
The chosen library seems to disable the actual `select` element and builds its own set of divs and spans for the new selection box. This means we lose all the functionaly of auto-refresh (on change triggers), e.g. when selection a new tool version in the task params

make sure all select box affected work properly and send their result properly.

prioux:
Is this all ready now?

PR #910, opened 2020-01-10
Fix for #848

natacha-beck:


natacha-beck:
@prioux I need to check the test before you review it.

PR #915, opened 2020-01-21
Fix for Select Tool panel UI

natacha-beck:


MontrealSergiy:
scales up and down well for me (Firefox 72.0.1 for Ubuntu canonical - 1.0)

MontrealSergiy:
Chromium only overlaps on tiniest font/size 
image - but anyways at that sizes content is not discernible without a good electron microscope, so not was the effort. Vivaldi do not show any tools at all, I guess not RoR friendly browser.

Issue #819, opened 2019-05-15
DP consistency checker interface improvements


prioux:
The DP consistency checker could benefit from some improvements:

1) direct links to it from the DP index page
2) not to show files that are unregistered as this is not much of a problem really (but maybe keep it as an option?)
3) when triggering the 'repair' process, spawn it in background instead of blocking the rails request
4) better categories for the different possible errors
5) filtering of the report table

natacha-beck:
Some of this issues are the same mentioned here: #593

prioux:
Fixed by #918

Issue #593, opened 2017-01-20
Several issues with the DP consistency validator


prioux:
* Separate columns for the type of issue and the instance of the issue (e.g. 'Missing File', 'abc.txt'),
* More filtering in the columns (by user, by action, 
* Fixing operation to happen in background (with tracking in the process command line, see how $0 is modified elsewhere)
* For VAULT DPs, errors are reported for the directories of the user vaults themselves, and these are not errors.

natacha-beck:
> Another bug: when switching page, the report DP is accessed and the full report is regenerated anew because the GET option `?reload=true` is added to the links! The reload option shoudl be implemented as a POST anyway.

Seems to be done: 
https://github.com/aces/cbrain/blob/master/BrainPortal/app/views/data_providers/report.html.erb#L29

prioux:
Closed by #918

PR #918, opened 2020-01-31
Dataprovider report

natacha-beck:
Fix the point 1, 2  and 3 of the following issue #593 + 
Fix the point 1, 3 and 5 of the following issue #819

### DIFF
--- a/BrainPortal/app/controllers/data_providers_controller.rb
+++ b/BrainPortal/app/controllers/data_providers_controller.rb
@@ -767,6 +767,9 @@ class DataProvidersController < ApplicationController
     @view_scope = @scope.apply(@issues, paginate: true)
 
     respond_to do |format|
+      if params[:reload]
+        format.html { redirect_to :action => :report }
+      end
       format.html
       format.js
       format.xml  { render :xml  => { :issues => @issues } }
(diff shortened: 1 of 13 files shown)


BrainPortal/app/views/data_providers/_one_data_provider_table.html.erb
```diff
@@ -124,9 +124,15 @@
...
 
     %>
+
     <% t.column("Operations", :operations) do |dp| %>
       <% if dp.is_browsable?(current_user) && dp.can_be_accessed_by?(current_user) && dp.online? %>
         <%= link_to 'Browse', browse_data_provider_path(dp), :class => 'action_link' %>
       <% end %>
     <% end %>
+
+    <% t.column("Inconsistency", :inconsitency) do |dp| %>
```
prioux: The symbol is misspelled, how did it work?

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -779,20 +779,33 @@ def repair
...
     end
 
-    # Display a message reporting how many issues were repaired
-    unless @issues.empty?
+    Message.send_message(current_user,
+      :message_type  => :notice,
+      :header        => "Repair #{@issues.size} inconsistencies in background.",
+    )
+
+    CBRAIN.spawn_with_active_records(:admin, "Repair DP inconsistencies") do
```
prioux: Change string to `"Repair DP #{self.name}"` so that if we start several in parallel, we can tell which process is repairing which DP

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -779,20 +779,33 @@ def repair
...
+      # Try to repair the inconsistencies (or issues)
+      failed_list  = []
+      success_list = []
+
+      @issues.each do |issue|
```
prioux: Change to `each_with_index_and_size` then within loop add

```ruby
Process.setproctitle "Repair DP=#{@provider.id} #{idx+1}/#{size}"
```

(see other examples in the same controller)

PR #923, opened 2020-03-13
Neurohub public

natacha-beck:
For remote code review only for @prioux .


natacha-beck:
@prioux I have made several new commits since Friday. Fixing tools, ad task permission for example. 
I think the main functionality of public group is implemented now (for Neurohub and CBRAIN). When you have time can you look at it. 

I still have issues with the test, I do not know why yet but Travis did not start to run the spec at all. And in the meantime, I have some tests to fix in order to test access restriction. 

Let me know how do you want to proceed with the code review.

PR #930, opened 2020-04-17
Naive repaging

MontrealSergiy:
Hi I tried to go as simple as I could with. 

Not sure is it right place on project and did not removed current project paging mockup, @candicecz might wish to clarify here

PS Another known issue:
If on the last or so page and page size is increased, it might land the user to the first page. I do not think there is a lot of ux value in fixing but tell me if I am wrong


prioux:
I created a commit based on the end result of this PR, which fixes several bugs and adds many other improvements:

https://github.com/aces/cbrain/commit/34686f1ac6f8f1c748250a896fdc0c697a099f34

So I'm closing this PR instead of merging.

PR #939, opened 2020-05-11
Adding instruction and command line

xlecours:
There is probably better ways to do this but this is better than <instructions goes here>

### DIFF
--- a/BrainPortal/app/views/nh_storages/_show_key.html.erb
+++ b/BrainPortal/app/views/nh_storages/_show_key.html.erb
@@ -36,7 +36,7 @@
       <div class="label">Command line</div>
       <div class="text">
         <pre>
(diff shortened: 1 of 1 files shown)


xlecours:
> Hi, may be mention that this works best/tested for Linux and Mac host? As not all version of windows support bash and ssh at the least out of box

Good point. Maybe it should be starting at "Once login to the host, ..."

MontrealSergiy:
Technically, windows android and most have login. May be connect to the remote unix/linux compatible server/file host, for example ```ssh username@filehostname .```

natacha-beck:
> Good point. Maybe it should be starting at "Once login to the host, ..."
> 
> Event there, the host might not be unix based.

In CBRAIN we assume that some of the resource should be on UNIX system (Portal and Bourreau) adding `Once login to the host` and a note about the fact it should be unix based, should be fine.

Issue #614, opened 2017-03-29
Task show page should provide link to all tasks in the same batch


prioux:
When a task is part of a batch, it would be nice to have a link in that task's show page that sends back to the task index page with a filter set for just the tasks in the batch.

blackknight75:
@prioux I would like to take a shot at this if you dont mind.

prioux:
Okay, go right ahead. It should be pretty easy, my only concern is that the helpers we have that might do the trick are not super easy to understand. Thanks!

uzorjchibuzor:
Is this issue still open?

prioux:
Yes.

uzorjchibuzor:
Okay, I will take a stab at it. Thank you

PR #940, opened 2020-05-19
Fix for Issue #614

candicecz:
[Task show page should provide link to all tasks in the same batch #614](https://github.com/aces/cbrain/issues/614)

### DIFF
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -1601,7 +1601,7 @@ class TasksController < ApplicationController
       "Status: #{@value.to_s.humanize}"
     end
 
-    # The methods below are TagFilter specific versions of the Scope::Filter
+    # The methods below are StatusFilter specific versions of the Scope::Filter
     # interface. See Scope::Filter for more details on how these methods
     # operate and for detailed parameter information.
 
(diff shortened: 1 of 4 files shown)


BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -1681,4 +1683,93 @@ def self.compact_hash(hash)
...
+    def initialize
+      @attribute = 'batch_id'
+    end
+
+    # Nice string representation of this filter for +pretty_scope_filter+.
+    def to_s
+      "Batch Id: ##{@value.to_s.humanize}"
+    end
+
+    # The methods below are TagFilter specific versions of the Scope::Filter
```
natacha-beck, line 1707: Should be `BatchFilter` not TagFilter. Same error line l1604 where it should be `Status Filter`

BrainPortal/app/views/tasks/show.html.erb
```diff
@@ -47,6 +47,17 @@
 <% if @task.bourreau.online? %>
 
   <div class="menu_bar">
+    <!-- TASKS IN BATCH -->
```
natacha-beck, line 50: Can you put the link into the Info section.

candicecz: I moved the link in the summary section instead because that's where it was in the other files that contained the broken links.

Issue #908, opened 2020-01-06
Add checkbox 'all locked users/all unlocked users' to report maker


prioux:
We have a general report maker (path `/report`)that is useful for administrative tasks. I was trying to review locked users with lots of disk space, but selecting all the locked users using the new `Chosen` select input is painful.

Add a checkbox that says `Add all locked users` or `Add all unlocked users`.

prioux:
This was closed by #941

PR #941, opened 2020-05-19
Fix for Issue#908

candicecz:
[Add checkbox 'all locked users/all unlocked users' to report maker #908](https://github.com/aces/cbrain/issues/908)

### DIFF
--- a/BrainPortal/app/views/portal/report.html.erb
+++ b/BrainPortal/app/views/portal/report.html.erb
@@ -418,10 +418,8 @@ in a tabular layout. Proceed as follow:
 
           <% if @model_atts.include?(:user_id) %>
             <tr class="report_fix_attributes">
-            <th>Users:</th>
-            <td colspan="2">
-              <%= user_select            'user_id',         { :selector => params[:user_id], :include_all => { :users => true, :locked => true, :active => true} }, :multiple => true, :size => 5 %>
-            </td>
+              <th>Users:</th>
+              <td><%= user_select          :user_id,          { :selector => params[:user_id] },                                  :multiple => true, :size => 5 %></td>
             </tr>
           <% end %>
 
--- a/BrainPortal/app/views/portal/report.html.erb
+++ b/BrainPortal/app/views/portal/report.html.erb
@@ -422,10 +422,9 @@ in a tabular layout. Proceed as follow:
             <td colspan="2">
               <%= user_select            :user_id,          { :selector => params[:user_id] },                                  :multiple => true, :size => 5 %>
               <div>
-                <%= check_box_tag(:show_locked, 1, params[:show_locked] ) %>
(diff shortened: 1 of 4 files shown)


BrainPortal/app/views/portal/report.html.erb
```diff
@@ -419,7 +419,15 @@ in a tabular layout. Proceed as follow:
           <% if @model_atts.include?(:user_id) %>
             <tr class="report_fix_attributes">
             <th>Users:</th>
-            <td><%= user_select          :user_id,          { :selector => params[:user_id] },                                  :multiple => true, :size => 5 %></td>
+            <td colspan="2">
+              <%= user_select            :user_id,          { :selector => params[:user_id] },                                  :multiple => true, :size => 5 %>
+              <div>
+                <%= check_box_tag(:show_locked, 1, params[:show_locked] ) %>
```
natacha-beck, line 425: As discussed please replace with 3 radio buttons: 
1. For all Users (default)
2. For unlock user only
3. For lock user only.

natacha-beck:
I found a pretty nice solution: 

- http://jsfiddle.net/jphase/A6LBv/ 

The code just need to be added in the `cbrain.js` a good place to do that is around l1010

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -1016,6 +1016,27 @@
     }
   });
 
+    // Add select/deselect all toggle to optgroups in chosen
```
prioux, line 1019: This seems to work but I suspect this code was added from some source on the internet? Then please add a comment with the source and make sure we're allowed to use that code.

natacha-beck: Yep see my comment 4 days ago:

> I found a pretty nice solution:
> 
> * http://jsfiddle.net/jphase/A6LBv/
> 
> The code just need to be added in the `cbrain.js` a good place to do that is around l1010

prioux: Yes, when I wrote 'I suspect' I really meant to say I'd seen you had provided a URL before, though I did no follow the link myself. Anyway, if Jeff Hays is the author, ask him if we can use the snippet, and credit him in the code.

jphase: Hey Team,

This is Jeff Hays, the author of the aforementioned snippet. You are more than welcome to use my code and I’m happy to see it being put to use! Let me know if there’s anything else I can help contribute to.

prioux: @jphase Hi Jeff! We'll include one comment line with your name and GitHub handle directly in the code. Thanks!

@natacha-beck since Candice is away on vacation can you do it please? Thanks.

PR #950, opened 2020-06-23
Editor in CBRAIN. Resolves #932

### LINKED ISSUE #932: [ADD] Information about Project Editor
At this point in CBRAIN we do not have information about Project Editor.

MontrealSergiy:
Adding editor table to classic project page

### DIFF
--- a/BrainPortal/app/views/groups/show.html.erb
+++ b/BrainPortal/app/views/groups/show.html.erb
@@ -58,8 +58,6 @@
         <span class="warning">Warning</span>:
           If you change the maintainer to someone else you won't be able to retrieve control over the project
       </div>
-      <%# the text outdate. at the moment any editor can set a new maintainer (and remove old maintainer from ) %>
-      <%# to do may be block editors from this the text update. at the moment any editor can set a new maintainer (and remove old maintainer from ) %>
     <% end %>
 
     <% t.edit_cell(:site_id, :content => link_to_site_if_accessible(@group.site), :disabled => !current_user.has_role?(:admin_user)) do %>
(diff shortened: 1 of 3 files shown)


prioux:
This code cannot be added until the NeuroHub code base has been merged in CBRAIN, because the database schema changes to support editors is only in NeuroHub.

natacha-beck:
This works well. 

But I found it weird, that the `Members and Their Project Roles` block will contain the redundant information with information present in `Quick Membership Control`. We can think to another way to present the information.

MontrealSergiy:
> https://portal.cbrain.mcgill.ca/report
@natacha-beck 
Is redundancy issue concerned with group editing or viewing only? In the latter case I can say just put the total number of member there or some brief description of the functionality. I maybe did not give enough thought in reducing redundant info on Quick Membership Control because it is only visible to admin, basically it is you and Pierre, normal users would not see it. I did not wanted to add much to that panel edit view because it is already large, because it has list of all users inside. I do not follow either what you suggest with report page, it seems like https://portal.cbrain.mcgill.ca/report is builder for wide square reports where both columns and rows defined by db content, not sure how it relates here

MontrealSergiy:
I think table is better and tidier UX than user box, but as a compromise I can add it to admin only - Quick member management panel.

prioux:
Eh guys, remember we already have neat select box helpers, don't use bare `select` when you can use `group_select` (see BrainPortal/app/helpers/select_box_helper.rb)

natacha-beck:
Just a comment about the number of user in groups, some groups have more than 1000 Users in CBRAIN. 
So with the current version, you will have 2 times the list of users: 

Screen Shot 2020-07-28 at 11 33 08 AM

So you will have 2 times the complete list of user one in `Members and Their Project Roles` and one in `Members`. 
This comment is related to my previous comment about information redundancy. At least you can just have the Editors list instead of `Members and Their Project Role`.

prioux:
Yes, please, always keep in mind the cases where thousands of units of something is present.

BrainPortal/app/views/groups/show.html.erb
```diff
@@ -110,28 +113,65 @@
...
   <%
     group_members  = @group.users
+    group_editors  = @group.editors || []
+    group_editors << @group.creator unless group_editors.include? @group.creator
+    group_viewers  = group_members - group_editors
     open_invites   = Invitation.where(sender_id: current_user.id, group_id: @group.id, active: true).to_a
-  %>
 
-  <%= show_table(@group, :as => :group, :header => 'Members', :edit_condition => @group.can_be_edited_by?(current_user) && (!current_user.has_role?(:normal_user) || group_members.count > 1 || open_invites.count > 0 )) do |t| %>
+    if @group.is_a?(WorkGroup) then # for now editors are only allowed on WorkGroups
```
prioux, line 121: Please redesign this layout, it's super ugly.

I suggest you list the members only once, with an asterisk beside them when they are editor (and a proper note about this).

BrainPortal/app/views/groups/show.html.erb
```diff
@@ -110,28 +113,65 @@
...
+              <%= link_to_user_with_tooltip u %>
+            </td>
+            <%  if @group.is_a?(WorkGroup) %>
+              <td>
+                <%= check_box_tag("group[editor_ids][]", u.id, @group.editor_ids.include?(u.id),
+                      :checked => u.id == @group.creator.id, :disabled=> u.id == @group.creator.id ) %>
+              </td>
+            <% end %>
+            <td>
+              <%= link_to("Remove", group_path(@group, "group[user_ids]" => (@group.user_ids - [u.id]), update_users: true ), :method => :put, :class => "action_link") unless u == current_user %>
```
prioux, line 166: Please don't do that kind of POST, when we have 1000+ users you get 1000 links each with 999 user ids in them, it's ridiculous

prioux:
Also you have your code comment showing up IN THE PAGE:

Screen Shot 2020-07-30 at 11 36 32

BrainPortal/app/views/groups/show.html.erb
MontrealSergiy: @prioux That's Tarek's code. I can refactor that of course.

prioux: Yes that's just awful. Have a look at what the HTMl looks like in production using your admin account, I bet it's horrific.

prioux:
This branch seems to have been abandoned, I'm going to close it without merging in one week if issues aren't fixed.

MontrealSergiy:
@prioux @natacha-beck a) At the time I was more inclined to use a more traditional table panel to work with project member privilege's because existing cbrain members panel seems already to big and intimidating for average user. It is easy to change if other project related roles or attributes are introduced. Probably paging or scroll are required for larger projects. To address Natacha comment on duplication I would put editors table on  dedicated page. Of course  I still can try stick existing style cbrain UI sprit and modify Members framework, say either b) having two checkboxes in the existing user rectangles - one for editor, perhaps one checkbox followed by editing emoji and another not editors; or c) add Editing Members rectangle to Members panel, I think it simpler to list all member. What do you think?

prioux:
This is stale. I'm closing it. We can revisit this later.

Issue #953, opened 2020-06-29
NeuroHub Alpha user feedback - technical issue - collaborator email mismatch and notification


bryancaron:
Reference survey containing raw data feedback in spreadsheet format:
https://docs.google.com/spreadsheets/d/1yt3iKKlMoRUgZl9E_fomAVgj38CgvJ2K3AKIeB3tMDQ/edit#gid=675520657

Feedback from Jérôme Dockès regarding the following technical experience issues (to split as needed into separate items):

- when inviting collaborators, If I enter one email address of a registered user and one that doesn't match a user (separated by spaces), the message says "successfully sent" and doesn't tell me that the invitation failed for one user.

MontrealSergiy:
@prioux should I fix this, unless there is a security issue?

MontrealSergiy:
I think adding `No user was found with email(s) {wrong_emails}, we are not able to invite them. At the moment only invitation of already boarded users is supported. Please ask them first to join NeuroHub, or if you expect a typo in the email address, ask them with which exactly email they are registered to NeuroHub. `

PR #963, opened 2020-07-06
Error message for unmatched collobarators emails issue #953

MontrealSergiy:
Note I not sure is there security risk. While neurohub already make it easy to check is user email in the system, the bugfix makes it easier further. May be not big deal may be something should be done to alleviate, like allow users to keep their emails confidential, not sure.

### DIFF
--- a/BrainPortal/app/controllers/nh_invitations_controller.rb
+++ b/BrainPortal/app/controllers/nh_invitations_controller.rb
@@ -40,12 +40,20 @@ class NhInvitationsController < NeurohubApplicationController
     user_emails     = (params[:emails].presence.try(:strip) || "").split(/[\s,]+/)
     user_emails     = user_emails.map(&:presence).compact
     cb_error "Please specify at least one email address", :redirect => nh_project_path(@nh_project) if user_emails.empty?
-    user_ids        = User.where(:email => user_emails).pluck(:id)
+    users           = User.where(:email => user_emails)
+    user_ids        = users.pluck(:id)
+    found_emails    = users.pluck(:email)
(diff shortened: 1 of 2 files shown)


prioux:
Yep, I plan to merge this once I'm done with the main neurohub cbrain merge

prioux:
I have change the merge bass to be 'master' just right now.

BrainPortal/app/controllers/nh_invitations_controller.rb
```diff
@@ -40,12 +40,20 @@ def create #:nodoc:
...
     cb_error "Please specify at least one email address", :redirect => nh_project_path(@nh_project) if user_emails.empty?
-    user_ids        = User.where(:email => user_emails).pluck(:id)
+    users           = User.where(:email => user_emails)
+    user_ids        = users.pluck(:id)
+    found_emails    = users.pluck(:email)
+    wrong_emails    = user_emails - found_emails
+
+    flash_messages = []
+    if wrong_emails.present?
+      flash_messages << "\nWe are not able to invite user(s) with email(s) #{wrong_emails.join(", ")}. At the moment users are matched by email that they are used to register in NeuroHub. Please confirm with them which email they provided to NeuroHub. "
```
natacha-beck, line 50: This PR is fine, I just would like to see some more '\n' in the message, it give a really long message that becomes not readable. 

The message regarding the failing invitation is actually in green (notice) I would like prefer to have it in red (error).

MontrealSergiy: I would call it pink but resolved. I also added yellow color flash, hope it is fine @prioux @candicecz ?

PR #973, opened 2020-07-28
Added new flag to groups: not_assignable. Fixes #962

### LINKED ISSUE #962: Implement groups that are viewable yet not assignable
I'm not sure we have this right now.

* If we flag a group as 'invisible', resources associated with the group can be accessed but not assigned to.
* However, in NeuroHub, the group will not be shown in the project list

I need a group to be visible in the project  list, yet not assignable. E.g. UK BioBank access project.

prioux:
The migration in here will also set all groups previously
'invisible' to also be 'not_assignable'. Them in the future we can
make some groups visible yet not assignable separately (e.g.
for datasets)

### DIFF
--- a/BrainPortal/app/models/admin_user.rb
+++ b/BrainPortal/app/models/admin_user.rb
@@ -36,6 +36,12 @@ class AdminUser < User
     Group.where(nil)
   end
 
+  # List of groups that the user can list in the interface. Normally, groups that are invisible
+  # are not listed
+  def listable_groups
+    Group.where(nil)
+  end
+
   # List of groups that the user can assign to resources.
   # The user must be a member of one of these groups. Subset
   # of viewable_groups
--- a/BrainPortal/app/models/admin_user.rb
+++ b/BrainPortal/app/models/admin_user.rb
@@ -31,20 +31,16 @@ class AdminUser < User
   end
 
   # List of groups which provide view access to resources.
-  # It is possible for the user not to be a member of one of those groups.
   def viewable_groups
     Group.where(nil)
   end
 
-  # List of groups that the user can list in the interface. Normally, groups that are invisible
-  # are not listed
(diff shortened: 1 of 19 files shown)


prioux:
@natacha-beck @MontrealSergiy If you would please review this?

The purpose is this: groups now have a "not_assignable" boolean flag. When set, a group will not appear in any selection boxes where people can assign them to resources (files etc)

In the past, the 'invisible' flag used to do this AND hide the group from the list of groups that people can see in their group list (index page, project page).

So the migration included will also make all invisible groups automatically 'not_assignable'.

But in the future we can no make a project that is visible in the group list yes people cannot assign them (invisible=false, not_assignable=true)

This PR also adds a new group listing method for users, `listable_groups`. This is the method to invoke to get the list of groups a user can see. It could contain LESS groups than the groups they can actually use (as returned by 'viewable groups`).

I created some testing code too, added in the misc directory. Follow the README. It will create stuff in your dev environment, please verify that the lists all seem OK to you. Some combinations are a bit weird but we shodul simply not create them (e.g. invisible public groups....)

BrainPortal/app/models/admin_user.rb
```diff
@@ -36,6 +36,12 @@ def viewable_groups
     Group.where(nil)
   end
 
+  # List of groups that the user can list in the interface. Normally, groups that are invisible
+  # are not listed
```
natacha-beck, line 40: The comment: `Normally, groups that are invisible are not listed` is misleading here since it is not the case for the admin.

prioux:
@MontrealSergiy I don't understand what you're saying: `why NeuroHub cannot be assigned to a resource` ? What does that mean?

MontrealSergiy:
@prioux Never mind. I withdraw that question, I just misunderstood the feature.

PR #978, opened 2020-08-04
NeuroHub invitations can be sent using login names

prioux:
This allows NeuroHub users to invite other users using their login names as well as their email addresses.

I also fixed/improved some messages and invitations.


MontrealSergiy:
works as intended 
image

PR #979, opened 2020-08-04
Better navigation between CBRAIN <-> NeuroHub

prioux:
Also:

* we record if the user submitted a new account request
using the CBRAIN form or the NeuroHub form. The info
is shown to the admin, that's all.

* we record if the user connected using the CBRAIN login
page or the NeuroHub login page; this info is not used right now
but could be turned into a "return home" link, as generated
by the new helper code (see login_dashboard_path(), not
used right now)

### DIFF
--- a/BrainPortal/app/assets/stylesheets/neurohub.scss.erb
+++ b/BrainPortal/app/assets/stylesheets/neurohub.scss.erb
@@ -1,35 +1,35 @@
 @import "<%= asset_path './normalize.css' %>";
 /*
   LEGEND
-  ├── BREAKPOINTS
-  ├── SPACE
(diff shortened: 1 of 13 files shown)


MontrealSergiy:
Albeit not directly relating I think cbrain can a message to forum or direct message. "A new experimental emerging portal neurohub.ca is build directly upon cbrain infrastructure. At the moment all the cbrain user from McGill are welcome to try it using cbrain credentials. Use this oppotity to try the new portal and share your experience. OR may be add 'breaking news" section to cbrain just for that message. I would prefer that to forum because we can filter out mcgill users only by chekcing email patterns.

Issue #977, opened 2020-08-03
No upload file progress panel in NH


natacha-beck:
When people upload file in NH no panel show up to tell progress to user.

natacha-beck:
Done by #983

PR #982, opened 2020-08-05
Add message in NH for file upload

### LINKED ISSUE #954: NeuroHub Alpha user feedback - technical issue - single file upload limit
Reference survey containing raw data feedback in spreadsheet format:
https://docs.google.com/spreadsheets/d/1yt3iKKlMoRUgZl9E_fomAVgj38CgvJ2K3AKIeB3tMDQ/edit#gid=675520657

Feedback from Seun Jeon regarding the following technical experience issues (to split as needed into separate items):

- Only one file at a time can be up/downloaded to/from the web.

natacha-beck:
This fix #954, by adding warning message about file size and pattern of filename if it is not an allowed pattern. 

I still need to fix the #977 

@candicecz will give some style to this page.


BrainPortal/app/views/nh_projects/new_file.html.erb
```diff
@@ -24,56 +24,89 @@
...
-                switch_groups_path(id: @nh_project.id),
-                :class => "btn-text cbrain bg-cbrain-wash",
-                :method => :post,
-                :target => "_blank"
-            %>
+<%= javascript_include_tag 'jquery'    %>
+<%= javascript_include_tag 'userfiles' %>
+
+<% 
+  upload_size_limit = RemoteResource.current_resource.meta[:upload_size_limit]
```
prioux, line 31: 1) these things are not always defined
2) if the admin edits the remote resource but leaves the fields blanks, they will probably get the empty string as a value, which is TRUE; so remember to use `.presence` to turn balnk strings into nils

BrainPortal/app/views/nh_projects/new_file.html.erb
```diff
@@ -24,56 +24,89 @@
...
-          <div class="label ">
-            For more advanced upload options:
-            <% cbrain_text= ("<p class='d-flex'>CBRAIN File Manager <span class='icon-cb'>#{nh_icon_cb_external}</span></p>").html_safe %>
-            <%= link_to cbrain_text,
-                switch_groups_path(id: @nh_project.id),
-                :class => "btn-text cbrain bg-cbrain-wash",
-                :method => :post,
-                :target => "_blank"
-            %>
+<%= javascript_include_tag 'jquery'    %>
```
prioux, line 27: I think including the CBRAIN JS could create interference... ? Are there errors in the console logs?

BrainPortal/app/views/nh_projects/new_file.html.erb
natacha-beck: done

BrainPortal/app/views/nh_projects/new_file.html.erb
natacha-beck: It was my main issue when working on it, but I did not want to create special code for something that should do the same think. 
For example I do not want to cut and paste the Regex somewhere else, so I decided to go with the `userfiles.js` I worked with my js console open and get on each error one by one (all my if statement). I also test it again on CBRAIN side userfiles index page to see if everything was ok. 
And everything seems to work fine but I definitively would like a second look.

BrainPortal/app/views/nh_projects/new_file.html.erb
natacha-beck: @prioux what I did not like on my side is the fact that I include jquery. In my mind this one should already be present but  it seems to not be the case.

BrainPortal/app/views/nh_projects/new_file.html.erb
natacha-beck: @prioux what I did not like on my side is the fact that I include jquery. In my mind this one should already be present but  it seems to not be the case.

prioux:
Here's an overall suggestion. This PR, as it is, requires patching a lot of CBRAIN JS code.

Would it be better to create a separate neurhub.js file with ONLY the minimum amount of JS code necessary to implement the feature, and without modifying anything on the cbrain side?

Even if this requires copying a few lines of JS code, this might be a better overall solution. Can it be done without jQuery at all?

PR #983, opened 2020-08-10
File Upload NH

natacha-beck:


natacha-beck:
@prioux and @candicecz this PR is ready for review and will close #982 and #977

BrainPortal/app/views/nh_projects/new_file.html.erb
```diff
@@ -24,56 +24,104 @@
...
+<%
+  upload_size_limit = RemoteResource.current_resource.meta[:upload_size_limit].presence
+  site_url_prefix   = RemoteResource.current_resource.site_url_prefix.presence
+  large_upload_url  = RemoteResource.current_resource.meta[:large_upload_url].presence
+%>
+
+<div id="userfiles_menus_and_filelist">
+  <div id="userfiles_dialogs" >
+
+    <div id="upload-dialog"
```
prioux, line 38: Are these ids and classes (e.g. `dlg-dialog`) used or are they left-over from the CBRAIN code where they were copied from?

BrainPortal/app/views/nh_projects/new_file.html.erb
```diff
@@ -24,56 +24,95 @@
...
-          <div class="label ">
-            For more advanced upload options:
-            <% cbrain_text= ("<p class='d-flex'>CBRAIN File Manager <span class='icon-cb'>#{nh_icon_cb_external}</span></p>").html_safe %>
-            <%= link_to cbrain_text,
-                switch_groups_path(id: @nh_project.id),
-                :class => "btn-text cbrain bg-cbrain-wash",
-                :method => :post,
-                :target => "_blank"
-            %>
+<%= javascript_include_tag 'neurohub' %>
```
prioux, line 27: Why do we load neurohub.js ONLY in this page? SHouldn't it be loaded as part of the overall layout?

If not, if it's because it only contains code related to upload, why call it 'neurohub.js' instead of 'neurohub_upload.js' or something like that?

PR #990, opened 2020-08-12
Invite notification fixes #959

### LINKED ISSUE #595: Signups "comment" field is string instead of text

### LINKED ISSUE #959: NeuroHub Alpha user feedback - technical issue - user invitation and notification
Received via direct email message sent to Diana:

a) Invite users: easy, I invited Diana, invitation is pending. I would like to receive a notification when the person ‘accept’ my invitation.

MontrealSergiy:
This adds notification for invitation acceptance using elements of cbrain messaging system.

### DIFF
--- a/BrainPortal/app/controllers/invitations_controller.rb
+++ b/BrainPortal/app/controllers/invitations_controller.rb
@@ -105,6 +105,13 @@ class InvitationsController < ApplicationController
     @invitation.save
 
     flash[:notice] = "You have been added to project #{@group.name}."
+
+    Message.send_message(@invitation.sender,
+                         :message_type   => 'notice',
+                         :header         => "Invitation Accepted",
+                         :variable_text  => "User #{current_user.login} accepted your invitation and joined project #{@invitation.group.name}"
+    )
+
     respond_to do |format|
       format.html { redirect_to groups_path }
       format.xml  { head :ok }
--- a/BrainPortal/app/controllers/invitations_controller.rb
+++ b/BrainPortal/app/controllers/invitations_controller.rb
@@ -109,8 +109,8 @@ class InvitationsController < ApplicationController
     Message.send_message(@invitation.sender,
                          :message_type   => 'notice',
                          :header         => "Invitation Accepted",
-                         :description    => "A user joined project #{@nh_invitation.group.name}",
-                         :variable_text  => "#{current_user.login} accepted your invitation and joined project #{@invitation.group.name}"
+                         :description    => "A user joined project #{@group.name}",
+                         :variable_text  => "#{current_user.login} accepted your invitation and joined project #{@group.name}"
     )
 
     respond_to do |format|
--- a/BrainPortal/app/controllers/invitations_controller.rb
+++ b/BrainPortal/app/controllers/invitations_controller.rb
@@ -109,7 +109,8 @@ class InvitationsController < ApplicationController
     Message.send_message(@invitation.sender,
                          :message_type   => 'notice',
                          :header         => "Invitation Accepted",
-                         :variable_text  => "User #{current_user.login} accepted your invitation and joined project #{@invitation.group.name}"
+                         :description    => "A user joined project #{@nh_invitation.group.name}",
+                         :variable_text  => "#{current_user.login} accepted your invitation and joined project #{@invitation.group.name}"
     )
 
     respond_to do |format|
--- a/BrainPortal/app/controllers/neurohub_application_controller.rb
+++ b/BrainPortal/app/controllers/neurohub_application_controller.rb
@@ -87,31 +87,6 @@ class NeurohubApplicationController < ApplicationController
   # Messaging System Filters (presently only invite acceptance)
   ########################################################################
 
-  # Find new messages to be displayed at the top of the page.
-  def prepare_messages #:nodoc:
-    return unless current_user
-    return if     current_user.all_licenses_signed.blank?
-    return if     request.format.blank? || request.xhr?
-    return unless request.format.to_sym == :html
-
-    @display_messages = []
-
-    unread_messages = current_user.messages.where( :read => false, :header => 'Invitation Accepted' ).order( "last_sent DESC" )
-
-    unread_messages.each do |mess|
-      if mess.expiry.blank? || mess.expiry > Time.now
-        if mess.critical? || mess.display?
-          @display_messages << mess
-          unless mess.critical?
-            mess.update_attributes(:display  => false)
-          end
-        end
-      else
-        mess.update_attributes(:read  => true)
-      end
-    end
-  end
-
   # Find the number of new invitations to be displayed at the top of the page.
   def prepare_invites
     return unless current_user
@@ -119,6 +94,14 @@ class NeurohubApplicationController < ApplicationController
     nh_new_invites        = Invitation.where(user_id: current_user.id, active: true, read: false).all || [];
     @nh_invites_count     = nh_invites.count
     @nh_new_invites_count = nh_new_invites.count
+  end
+
+  # this patches prepare_message method
+  # to display 'Invitation Accepted' entitled messages. They will be show separately from Invitations
+  # and only kind of message to be shown for now.
+
+  def unread_messages_to_display
+    current_user.messages.where( :read => false, :header => 'Invitation Accepted' ).order( "last_sent DESC" )
   end
 
   # Check if password need to be reset.
--- a/BrainPortal/app/controllers/neurohub_application_controller.rb
+++ b/BrainPortal/app/controllers/neurohub_application_controller.rb
@@ -95,7 +95,7 @@ class NeurohubApplicationController < ApplicationController
     @nh_new_invites_count = nh_new_invites.count
   end
 
-  # below a method is overwrited to allow to render only invitation confirmation, filtered by header (for now)
+  # below a method is overwriten to render only invitation confirmation, filtered by header (for now)
   
   def unread_messages_to_display #:nodoc:
     current_user.messages.where( :read => false, :header => 'Invitation Accepted' ).order( "last_sent DESC" )
--- a/BrainPortal/app/controllers/neurohub_application_controller.rb
+++ b/BrainPortal/app/controllers/neurohub_application_controller.rb
@@ -31,6 +31,7 @@ class NeurohubApplicationController < ApplicationController
 
   before_action :switch_to_neurohub_layout
   before_action :prepare_invites
+  before_action :prepare_messages
 
   # This before_action callback sets a instance
   # variable @_NeuroHubLayout_ which is used by
@@ -80,6 +81,36 @@ class NeurohubApplicationController < ApplicationController
       format.json  { render :json => {:message => message}, :status => :service_unavailable }
     end
 
+  end
+
+  ########################################################################
+  # Messaging System Filters (presently only invite acceptance)
+  ########################################################################
+
+  # Find new messages to be displayed at the top of the page.
+  def prepare_messages #:nodoc:
+    return unless current_user
+    return if     current_user.all_licenses_signed.blank?
+    return if     request.format.blank? || request.xhr?
+    return unless request.format.to_sym == :html || params[:controller] == 'messages'
+
+    @display_messages = []
+
+    unread_messages = current_user.messages.where( :read => false, :header => 'Invitation Accepted' ).order( "last_sent DESC" )
+    @unread_message_count = unread_messages.count
+
+    unread_messages.each do |mess|
+      if mess.expiry.blank? || mess.expiry > Time.now
+        if mess.critical? || mess.display?
+          @display_messages << mess
+          unless mess.critical?
+            mess.update_attributes(:display  => false)
+          end
+        end
+      else
+        mess.update_attributes(:read  => true)
+      end
+    end
   end
 
   # Find the number of new invitations to be displayed at the top of the page.
--- a/BrainPortal/app/controllers/neurohub_application_controller.rb
+++ b/BrainPortal/app/controllers/neurohub_application_controller.rb
@@ -92,12 +92,11 @@ class NeurohubApplicationController < ApplicationController
     return unless current_user
     return if     current_user.all_licenses_signed.blank?
     return if     request.format.blank? || request.xhr?
-    return unless request.format.to_sym == :html || params[:controller] == 'messages'
+    return unless request.format.to_sym == :html
 
     @display_messages = []
 
     unread_messages = current_user.messages.where( :read => false, :header => 'Invitation Accepted' ).order( "last_sent DESC" )
-    @unread_message_count = unread_messages.count
 
     unread_messages.each do |mess|
       if mess.expiry.blank? || mess.expiry > Time.now
--- a/BrainPortal/app/controllers/neurohub_application_controller.rb
+++ b/BrainPortal/app/controllers/neurohub_application_controller.rb
@@ -96,11 +96,9 @@ class NeurohubApplicationController < ApplicationController
     @nh_new_invites_count = nh_new_invites.count
   end
 
-  # this patches prepare_message method
-  # to display 'Invitation Accepted' entitled messages. They will be show separately from Invitations
-  # and only kind of message to be shown for now.
-
-  def unread_messages_to_display
+  # below a method is overwrited to allow to render only invitation confirmation, filtered by header (for now)
+  
+  def unread_messages_to_display #:nodoc:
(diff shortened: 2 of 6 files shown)


BrainPortal/app/views/layouts/_neurohub_flash_message.html.erb
```diff
@@ -35,3 +35,16 @@
       <%= (h(msg.to_s.strip).gsub(/\n/, '<BR/>')).html_safe %>
     </div>
   <%- end -%>
+
+  <div class="card">
```
prioux, line 39: Shouldn't the div be inside the `if` block ? Otherwise we get a div with nothing inside.

BrainPortal/app/views/nh_invitations/_message_details.html.erb
```diff
@@ -0,0 +1,52 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.  
+#
+-%>
+<% messtype = mess.message_type # disguise confirmation for a flash notice until better styling added
+   prefix = "flash"
+   messclass = "#{prefix}_#{messtype}"
+%>
+
+<% unless mess.description.blank? %>
```
prioux, line 29: Change the `unless something.blank?` to `if something.present?` it looks much nicer and avoids double negatives.

BrainPortal/app/controllers/invitations_controller.rb
```diff
@@ -105,6 +105,14 @@ def update #:nodoc:
     @invitation.save
 
     flash[:notice] = "You have been added to project #{@group.name}."
+
+    Message.send_message(@invitation.sender,
+                         :message_type   => 'notice',
+                         :header         => "Invitation Accepted",
+                         :description    => "A user joined project #{@nh_invitation.group.name}",
```
natacha-beck, line 112: I confirm it make it crash.

BrainPortal/app/views/layouts/_neurohub_flash_message.html.erb
natacha-beck: it should be inside the `unless` block.

natacha-beck: + same comment as @prioux made later:
> Change the unless something.blank? to if something.present? it looks much nicer and avoids double negatives.

BrainPortal/app/views/layouts/_neurohub_flash_message.html.erb
```diff
@@ -35,3 +35,16 @@
...
+  <%- unless @display_messages.blank? %>
+
+    <%- for mess in @display_messages -%>
+      <p>
+      <div class="nh-invitations card-item">
+      <%= render(:partial  => "nh_invitations/message_details", :locals  => { :mess  => mess } ) rescue "Oh oh. your message was not rendered. Talk to devs!" %>
```
natacha-beck, line 45: You do not need a rescue on a render. Rails will handle that.

In meantime we have the same code in `./messages/_message_display.html.erb` ==> ligne 33.

MontrealSergiy: well, much code is similiar to similiar cbrain features. Do you say I should 'unrescue' old code too?

BrainPortal/app/controllers/invitations_controller.rb
```diff
@@ -105,6 +105,14 @@ def update #:nodoc:
     @invitation.save
 
     flash[:notice] = "You have been added to project #{@group.name}."
+
```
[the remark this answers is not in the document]
MontrealSergiy: I guess @group will be best or at the least the shortest

BrainPortal/app/views/layouts/_neurohub_flash_message.html.erb
natacha-beck: Can you give me example in CBRAIN ? I did not see them.

BrainPortal/app/controllers/neurohub_application_controller.rb
```diff
@@ -83,6 +84,36 @@ def check_if_locked
...
+  ########################################################################
+  # Messaging System Filters (presently only invite acceptance)
+  ########################################################################
+
+  # Find new messages to be displayed at the top of the page.
+  def prepare_messages #:nodoc:
+    return unless current_user
+    return if     current_user.all_licenses_signed.blank?
+    return if     request.format.blank? || request.xhr?
+    return unless request.format.to_sym == :html || params[:controller] == 'messages'
```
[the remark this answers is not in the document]
MontrealSergiy: well... I think putting && it will break the functionality (mind `unless`). As a compromise I can remove or comment out  the second part of the condition since, at the moment, there is no messages controller in NeuroHub part at the moment.

BrainPortal/app/views/nh_invitations/_message_details.html.erb
MontrealSergiy: ok Not sure is being blank something negative, `if` is shorter and as I know many devs consider `unless` bad style unless it removes a double negation.

BrainPortal/app/views/layouts/_neurohub_flash_message.html.erb
MontrealSergiy: ./messages/_message_display.html.erb ==> ligne 33.

natacha-beck: Yep, I already mention it. Seems to be the 2 only place with this strange rescue. Please remove at both place.

MontrealSergiy: besides the templates, much of message processing in neurohub_controller.rb is copy of cbrain_controller.

BrainPortal/app/controllers/neurohub_application_controller.rb
```diff
@@ -83,6 +84,36 @@ def check_if_locked
...
+
+    @display_messages = []
+
+    unread_messages = current_user.messages.where( :read => false, :header => 'Invitation Accepted' ).order( "last_sent DESC" )
+    @unread_message_count = unread_messages.count
```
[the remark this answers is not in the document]
MontrealSergiy: no reason. The possible future use is to put number of the unread messages over 'message' item of the navigation bar/top menu like Candice done. I guess can be deleted both from CBRAIN and NeuroHub controllers.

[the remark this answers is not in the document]
prioux: Don't do that. Just keep the code clean. If someone in the future needs an instance variable, they'll add it then.

BrainPortal/app/views/layouts/_neurohub_flash_message.html.erb
MontrealSergiy: It does not handle well errors on my pc image

prioux: Try in the Rails `production` environment

MontrealSergiy: in production mode during malfunction the message seems silently hidden, which mean the user will do not know that his message got lost. Do you still want remove that rescues ? @natacha-beck

prioux: Just use the same `rescue` that was there on the CBRAIN side, then.

prioux: I don't want to waste too much time about this anymore.

MontrealSergiy: As a compromise I removed the rescue form neurohub as users might understand that it is experimental project and confirmations could be lost

MontrealSergiy:
@natacha-beck actually you are right the rails handles templates error and produces an error message saying admins are notified of the issue. Regretfully for some reason when that happens neurohub redirects user to cbrains group page which seems strange.

Screenshot 17 08 2020 22_47_19

BrainPortal/app/controllers/neurohub_application_controller.rb
```diff
@@ -31,6 +31,7 @@ class NeurohubApplicationController < ApplicationController
 
   before_action :switch_to_neurohub_layout
   before_action :prepare_invites
+  before_action :prepare_messages
```
prioux, line 34: Now you're asking the rails framework to call `prepare_messages` twice for every request, because the callback is already registered in ApplicationController

BrainPortal/app/controllers/neurohub_application_controller.rb
```diff
@@ -83,6 +84,35 @@ def check_if_locked
 
   end
 
+  ########################################################################
+  # Messaging System Filters (presently only invite acceptance)
+  ########################################################################
+
+  # Find new messages to be displayed at the top of the page.
+  def prepare_messages #:nodoc:
```
prioux, line 92: This is a complete duplication of the method of the same name in Application controller, except for a single line. That is not DRY at all.

Please:

1. remove this method here since it's already defined in ApplicationController
2. change the method in ApplicationController such that the one line that differs invokes a method that can be overrided, which you will add to both files:
```ruby
   unread_messages = unread_messages_to_display
```

so that  in ApplicationController:

```ruby
# put this method just after prepare_messages
def unread_messages_to_display 
   current_user.messages.where( :read => false ).order( "last_sent DESC" )
end
```

and in NeurohubApplicationController we ONLY have this now:

```ruby
def unread_messages_to_display
  current_user.messages.where( :read => false, :header => 'Invitation Accepted' ).order( "last_sent DESC" )
end
```

BrainPortal/app/views/nh_invitations/_message_details.html.erb
```diff
@@ -0,0 +1,52 @@
+
```
prioux, line 1: This file seems like aalso a almost complete duplication of `messages/_message_details.html.erb`, can we instead use THAT and not duplicate so much code?

BrainPortal/app/controllers/neurohub_application_controller.rb
prioux: I checked and it's not called twice, Rails is intelligent enough not to invoke a method twice if you register it several time. But the point is still valid, it's already registered in ApplicationController.

BrainPortal/app/controllers/neurohub_application_controller.rb
```diff
@@ -83,6 +84,35 @@ def check_if_locked
...
+    return unless current_user
+    return if     current_user.all_licenses_signed.blank?
+    return if     request.format.blank? || request.xhr?
+    return unless request.format.to_sym == :html
+
+    @display_messages = []
+
+    unread_messages = current_user.messages.where( :read => false, :header => 'Invitation Accepted' ).order( "last_sent DESC" )
```
prioux, line 100: This query is brittle by the way. If someone in the future change the header of the message that is being sent (see your code in InvitationControlelr and NhInvitationController), then this piece of code will stop displaying the message.

BrainPortal/app/views/nh_invitations/_message_details.html.erb
```diff
@@ -0,0 +1,52 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.  
+#
+-%>
+<% messtype = mess.message_type
+   prefix = "flash"  # disguise an invite confirmation message for a flash notice until proper styling defined
+   messclass = "#{prefix}_#{messtype}"
+%>
+
+<% if mess.description.present? %>
+  <p class=<%= messclass %>>
```
prioux, line 30: By the way, the values of the HTML attributes should be quoted. They were quoted in the original `messages/_message_details.html.erb`, why did you remove the quotes in your copy?

BrainPortal/app/views/nh_invitations/_message_details.html.erb
MontrealSergiy: Surely, it is easy to adjust that file. That what I done in an early version of the feature. I opted to allow Neurohub have it own template that can evolve independtly, but totally can drop that

BrainPortal/app/views/nh_invitations/_message_details.html.erb
MontrealSergiy: I tried a bit improve over existing code.

MontrealSergiy: Note If I add quotes to this version I will get double quotes

BrainPortal/app/views/nh_invitations/_message_details.html.erb
prioux: OK, I agree with this sentiment. Let's have a separate file for neurohub, like you did.

prioux: You convinced me.

BrainPortal/app/controllers/neurohub_application_controller.rb
MontrealSergiy: Should I then add `message_type=confirmation` flag along with relevant styles? 

Showing all the cbrain messages seems a bit risky. Or as a half measure may be add a unittest?

prioux: Just use the same query but encspsulate it in a method, like I explained above, so that you only have one 'prepare_message' callback in ApplicationController.

BrainPortal/app/controllers/nh_invitations_controller.rb
```diff
@@ -132,6 +132,13 @@ def update #:nodoc:
     @nh_invitation.save
 
     flash[:notice] = "You have been added to project #{@nh_project.name}."
+
+    Message.send_message(@nh_invitation.sender,
```
prioux, line 136: By the way, here, why not use @nh_project.name instead of @nh_imvitation.group.name ?

MontrealSergiy: ok that's easy. I experimented a bit to make cut and paste from cbrain which we often use for neurohub smoother ( you do not have to remember that groups became nh_projects, just add/remove nh). But can change.

BrainPortal/app/controllers/neurohub_application_controller.rb
```diff
@@ -122,5 +97,13 @@ def prepare_invites
     @nh_new_invites_count = nh_new_invites.count
   end
 
+  # this patches prepare_message method
+  # to display 'Invitation Accepted' entitled messages. They will be show separately from Invitations
+  # and only kind of message to be shown for now.
+
```
prioux, line 103: You must not leave a blank line between the method description and the method, otherwise the documetnation code produces a method with no description

prioux: I'd also suggest adding the #:nodoc: keyword because that method is not really a method that CBRAIN developer will want to invoke at other places.

prioux: Same with the copy in ApplicationController

BrainPortal/app/controllers/neurohub_application_controller.rb
prioux: Please remove it

BrainPortal/app/views/nh_invitations/_message_details.html.erb
prioux: This is not fixed yet

BrainPortal/app/controllers/neurohub_application_controller.rb
MontrealSergiy: added

BrainPortal/app/views/nh_invitations/_message_details.html.erb
MontrealSergiy: added

BrainPortal/app/controllers/neurohub_application_controller.rb
MontrealSergiy: nodoc added

MontrealSergiy:
I think I addressed the remaining comments. I kept a query method description separated because, as you noticed, it is fragile but put the nodoc thing.

BrainPortal/app/controllers/neurohub_application_controller.rb
```diff
@@ -91,6 +95,12 @@ def prepare_invites
     @nh_new_invites_count = nh_new_invites.count
   end
 
+  # below a method is overwriten to render only invitation confirmation, filtered by header (for now)
+
```
prioux, line 99: Don't leave a blank line!

PR #995, opened 2020-08-13
CBRAIN users can generate API token

prioux:
They get a new page with a new token ready
to be used for API requests.

Right now the new page is only on the CBRAIN side, I'll add a similar on to the NeuroHub side too.

Possible improvement: adding a button to send the token to the clipboard. Requires Javascript however.


prioux:
By the way, do not wait for further features (e.g. page in NeuroHub, please review right away so users can start using the feature.

prioux:
@xlecours Thanks! Just an hour ago I was saying to the CBRAIN team that maybe asking you to review this was not necessary!

Issue #957, opened 2020-06-29
NeuroHub Alpha user feedback - technical issue - user registration, login and UI flow 


bryancaron:
Received via direct email message sent to Diana:

a) I tried to log-in with my Orcid ID
I was correctly re-directed in Orcid environment – I typed my credentials then I got this message
"No NeuroHub user matches your ORCID ID. Create a NeuroHub account, or add your ORCID ID to your account."
 
I was not sure what to do, I did not see any ‘ADD your ORCID ID to your account’ so I decided to create a NeuroHub account.
Note: The ‘request an account’ option should be in the log-in section

b)In the form to request an account, I was wondering what kind of ‘nickname’ should I use, just my name, something funny, or more professional? I was not clear if you recommend to have only one letter + numbers or if it was meaning that its mandatory to have at least one letter. It is necessary to have that ‘preferred login name’ ?

PREFERRED 'LOGIN' NAME
(one letter + alphanums)
 
c) For current position, I would harmonize
PhD Candidate
Master degree Candidate
Undergraduate student
And I would add:
Research professional
Health professional
Physician

d) After requesting an account; it took about 1:30 h to get the email to verify the email address… and I received it twice.
Once I got the ‘welcome to CBRAIN’ email, it felt weird, as a user, to be re-directed to CBRAIN.. and ‘access CBRAIN’… terms of use of CBRAIN…. etc, even if it’s says powered by CBRAIN at the bottom. I think this will need to be changed for NeuroHub…?!

e) I followed the steps and I was naturally connecting and navigating in CBRAIN. I had to close the window and go back to Neurohub platform and log-in with the credentials freshly obtained through CBRAIN. It felt wrong.
NeuroHub use my CBRAIN username (automatically given) instead of the ‘preferred login name’ mentioned in my ‘NeuroHub request form’

prioux:
Yep we need to document for users:
1- user gets neurohub account
2- user connects with login password
3- user ADDs ORICID ID
4- at that point user can, in the future, login in with orcid

prioux:
for b) we the login name is a suggestion but we ignore it if it's too unlike our conventions

prioux:
For c) that list of titles was agreed upon last year with Shawn

prioux:
for d) email are sent immediately ; the welcome email is created by CBRAIN, we can create a template for neurohub with a neurohub specific text

bryancaron:
Skip c).  Nothing to change.
d) already addressed, to be confirmed.

[...]

PR #996, opened 2020-08-17
Added note about ORCID in login page

prioux:


prioux:
This will close #957 a)

prioux:
@candicecz This is just a single line of HTML code, just check the appearance and if you have suggestions, just tell me how to change it and I will make the modifs. I used the same field_note CSS class as the other note in the login page.

prioux:
Note: I already merged this by mistake on my work computer and pushed it.

Issue #992, opened 2020-08-12
Add information in user new page + allow selection of preferred DP


natacha-beck:
When creating a new account in CBRAIN we do not have information about the provenance of the request, was is send by CBRAIN or NH (even if we have the information in index signup page).

In the meantime, we can add a select box allowing the admin to select a `pref_data_provider_id` (note: this information is store in the metadata of the user.

prioux:
I'm adding the NeuroHub tag because it will help provide a default DP for new NeuroHub users.

prioux:
Implemented in #997

Issue #958, opened 2020-06-29
NeuroHub Alpha user feedback - technical issue - File upload and Storage location selection


bryancaron:
Received via direct email message sent to Diana:

a) Upload file:The name of the file I was trying to upload had invalid characters (it was: File 1 2020-05-25). So I had to rename it (File_1) and try again. 

b) I don’t understand this: SELECT STORAGE FOR FILE CONTENT
The storage configuration selected here will control who has access to the content of the file. Each storage configuration is associated with a particular project, and the members of that project are the ones who can access the file's content.

I choose NeuroHubStore randomly, but someone will have to explain to me.

bryancaron:
a) no spaces allowed in file names, among other restrictions. documentation and training issue.

b) Default will be NeuroHubStore
- but needs to be set at account creation time
- right now not easily to differentiate between users applying though CBRAIN versus NeuroHub instance to get context (cumbersome!)

- can add an attribute where the form was filled from
- easy to implement 

- done after the code merge - few lines of code - estimate < 1 day after code merge

prioux:
Let's make sure all users gets a default data provider configured at account creation, and depends on neurohub user vs cbrain user

prioux:
This was fixed by #979, we simply need to be careful when creating new accounts to remember to assign the proper default DP.

PR #997, opened 2020-08-18
We can specify default DP for new users

prioux:
Also we show in the new user creation form
if the signup was performed on CBRAIN or NeuroHUb


prioux:
This will implement #992

prioux:
This will also help for #958 issue b)

Issue #999, opened 2020-08-19
Add "new token" page in NeuroHub interface.


prioux:
The CBRAIN interface has "generate a new token" page. We should have the counterpart on the NeuroHub page.

prioux:
Implemented in #1000

PR #1000, opened 2020-08-20
New Token page for NH users.

prioux:
Also includes a few other fixes:

* We log the user agent for the first API request
* Session cookies are sent with SameSite: Strict
* The switcher button is less likely to switch to bad routes on the other side


prioux:
Will fix #999

prioux:
@MontrealSergiy You're right and I removed it. It's not a the Rails level that we can handle this sort of problem anyway. We'll get a contract with cloudflare if we're ever big. But that's not likely.

MontrealSergiy:
webpage works. I think it is not so difficult to limit users clicks properly by checking timestamps but not big deal to me

MontrealSergiy:
image

PR #1007, opened 2020-09-08
[FIX] Disable refresh if needed.

### LINKED ISSUE #937: ResourceUsage data filtering seem to ignore the "To" date
For the resource usage index page, we have the capability to filter by date ranges; it seems the "to" date is ignored.

natacha-beck:
#937 seems to be not reproducible, but if we forgot to change the radio button, no filter will be applied. This PR will unable the `Refresh table` button according to the radio button value. 

Closed #937.

### DIFF
--- a/BrainPortal/app/views/resource_usage/_resource_usage_table.html.erb
+++ b/BrainPortal/app/views/resource_usage/_resource_usage_table.html.erb
@@ -158,7 +158,7 @@
         $(document).ready(function() {
           $('input[name="date_range[date_attribute]"]').on('click change', function(e) {
             if ($('input[name="date_range[date_attribute]"]:checked').val() === "") {
-              $('input[id="refresh_data_table"]').prop( "disabled", true )
+              $('#refresh_data_table').prop( "disabled", true )
             } else {
               $('#refresh_data_table').prop( "disabled", false )
             };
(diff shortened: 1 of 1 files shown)


prioux:
"unable" ? Do you mean "enable" or "disable" ?

natacha-beck:
The `Refresh table` button should now be unable only if the user have NOT selected the `Do not filter` radio button.

BrainPortal/app/views/resource_usage/_resource_usage_table.html.erb
```diff
@@ -152,7 +152,20 @@
...
       <br>
 
-      <%= submit_tag 'Refresh Table' %>
+      <%= submit_tag 'Refresh Table', :id => "refresh_data_table", :disabled => true %>
+
+      <script type="text/javascript">
+        $(document).ready(function() {
+          $('input[name="date_range[date_attribute]"]').on('click change', function(e) {
+            if ($('input[name="date_range[date_attribute]"]:checked').val() === "") {
+              $('input[id="refresh_data_table"]').prop( "disabled", true )
```
prioux, line 161: You already gave an ID to the submit tag, so why do you use a complex jQuery selector like that instead of simply `$('#refresh_data_table')`  ?

PR #1011, opened 2020-09-15
Allow creator to delete project via NH

natacha-beck:


BrainPortal/app/controllers/nh_projects_controller.rb
```diff
@@ -52,6 +52,25 @@ def create #:nodoc:
...
+
+    if ! current_user.has_role?(:admin_user)
+      if current_user.id != @nh_project.creator_id
+        cb_error "Cannot destroy this project: you are not its creator.", :redirect => nh_project_path(@nh_project)
+      end
+    end
+    
+    @nh_project.destroy
+
+    respond_to do |format|
```
prioux, line 66: Please don't add support for other request types until we need them

BrainPortal/app/controllers/nh_projects_controller.rb
```diff
@@ -52,6 +52,25 @@ def create #:nodoc:
...
+  def destroy #:nodoc:
+    @nh_project = current_user.modifiable_groups.find(params[:id])
+
+    if ! current_user.has_role?(:admin_user)
+      if current_user.id != @nh_project.creator_id
+        cb_error "Cannot destroy this project: you are not its creator.", :redirect => nh_project_path(@nh_project)
+      end
+    end
+    
+    @nh_project.destroy
```
prioux, line 64: If the project cannot be destroyed, it would be nice if the user got the reasons, in the flash maybe.

MontrealSergiy:
Creator or maintainer?

natacha-beck:
Creator only it was discussed in the issue #1011

prioux:
I think Sergiy was talking about your message, where you use the word 'creator'.

MontrealSergiy:
It is not like I care who creator or maintainer. Just want to make sure that we are requester is aware of the choice and wording is consistent with functionality.

BrainPortal/app/views/nh_projects/show.html.erb
```diff
@@ -23,6 +23,9 @@
 -%>
 
 
+<script src="/assets/jquery.js"></script>
```
prioux, line 26: This won't work in production, assets are precompiled

prioux: Why do we even need to include this here anyway?

prioux (changes requested):
The jQuery includes are wrong

Issue #1016, opened 2020-09-22
Add support for launching Singularity tasks with overlays


prioux:
Right now the CBRAIN system has support for data providers where the file content is store in SquashFS files; the data is extracted on demand by launching a singularity container which will run the necessary rsync/sftp/cp commands. This means CBRAIN can properly connect to such a data provider, and extract the information into the local cache for a Portal or a Bourreau. When launching a task, the `sync_to_cache()` operation on in the input file will result in the data being copied to the cache before the task is started, and the task will access the data from that cache. That's all part of the standard operating model.

It's not very efficient if the SquashFS data provider contains tons of information that need to be processed, though: it means we get to copy out the data files before launching a task. It's particularly redundant if the task itself is meant to run within a singularity container, which has the ability to mount directly a SquashFS filesystem.

We need to find a way to launch tasks such that:

1) if input files are on a SquashFS data provider and
2) the SquashFS files are on the same execution server and accessible by the CBRAIN account running the Bourreau and
3) the task's processing code is run through singularity

then instead of synchronizing the input files, we add the DP's SquashFS overlays to the singularity environment.

Now, here's an important point: within the singularity container, the paths to the data files will be different. We need a way for the `make_available()` methods in ClusterTask to be aware of the future execution context and create symbolic links that are not pointing to the cache (as is the case right now), but are pointing to the paths inside the SquashFS mountpoints.

I think this can all be accomplished with the following improvements to the CBRAIN framework:

1) a tool config will get new attributes that can be set by the admin to describe which squashfs files to add as overlay. There are several ways, not sure which one to choose
  a) a text box with a list of explicit path "/a/b/data1.sqs /a/b/data2.sqs" or file pattern "/a/b/data*.sqs"
  b) a list of DataProvider IDs; the tool config would query them and get from them the path to the squashfs files (by building DP's remote_dir + get_squashfs_basenames; suggestion: add a method in SingSquashDataProvider class)

2) a boutiques descriptor would get new custom cbrain control entries, and the Boutiques integrator would modify its behavior when creating the task's code:
```json
"custom": {
  "cbrain:no-sync-input": [ "id_of_input_file_1", "id_of_input_file_2"]
}
```

3) the "make_available()" method in cluster task would need somehow to be aware that the input is not to be synchronized, and that the symbolic link to be created
is to be made using the internal SquashFS path; I am not sure how to provide the information to trigger this alternate behavior. Also, we must make sure that the boutiques
integrator invokes it properly.

[...]

PR #1019, opened 2020-09-23
ToolConfigs can have Singularity overlays to mount

prioux:
The admin can specify singularity overlays to mount.
The overlays can be specific paths, or refer to
all the overlays assocuated with a SingSquashfsDataProvider.

The make_available() method will detect if a file is
located on such an overlay and create its symlink to
point direcly there without synchronizing.


prioux:
Will fix #1016

prioux:
OK, well, I've read the code diffs yet again (like I've done it a dozen times already, for real) and I'll merge this.

Issue #1028, opened 2020-09-29
NeuroHub file upload problem - cannot assign existing Storage 


bryancaron:
I have an issue when testing some aspects of the NeuroHub portal.  Specifically file upload.  I have my test project created but when attempting to upload a file through the GUI it gives an error  message that I have to assign storage first.  Doing a + Add Storage then prompts me to create a new storage configuration (private storage).

![Screen Shot 2020-09-28 at 9 11 23 PM](https://user-images.githubusercontent.com/65622686/94503335-393e2280-01d4-11eb-93a4-11330f02106a.png)

However, when I’m in CBRAIN I go to the same Project and select Upload it immediately proceeds and allows me to select the file and does the upload.

![Screen Shot 2020-09-28 at 9 17 18 PM](https://user-images.githubusercontent.com/65622686/94503359-44914e00-01d4-11eb-9964-5a43fc2ee2e2.png)

What is the difference happening in behind that is preventing the upload from working within NeuroHub?
I tried this since I noticed the documentation on the NeuroHub page didn’t have the upload step, and I wanted to exercise it, as something to add to the documentation and to further have updated Tuesday as a confirmed functional action. 

Thanks.

MontrealSergiy:
not necessary bug, could be misconfiguration issue. Very recently experimental NeuroHub server on port 5000 was deactivated yet setting for "neurohub service" storage `.meta[:neurohub_service_dp_ids]` probably were not copied to the main portal server

prioux:
Yes, it was just a configuration issue. The list of default DPs for neurohub users was empty.

prioux:
By the way, Sergiy's hypothesis was exactly right. The real fix was in the console:

```ruby
cb=BrainPortal.find(5)
nh=BrainPortal.find(58)
cb.meta[:neurohub_service_dp_ids] = nh.meta[:neurohub_service_dp_ids]
```

bryancaron:
Thank you.

PR #1029, opened 2020-09-29
[FIX] User metadata hash for Neurohub DP

natacha-beck:
This should fix this issue #1028. Since is it a hotfix I put the base as the master branch.


MontrealSergiy:
@natacha-beck will it help users who log in using cbrain credentials? Maybe not a bug but configuration issue not necessary bug, could be misconfiguration issue. Very recently experimental NeuroHub server on port 5000 was deactivated yet setting for "neurohub service" storage .meta[:neurohub_service_dp_ids] probably were not copied to the main portal server

prioux:
Rejected. The code was perfectly correct

PR #1035, opened 2020-10-01
neurohub messaging resolves #986

### LINKED ISSUE #986: add full messaging system to Neurohub
Create system and user to user messages for neurohub

Depending on feedback from the team add
- message to support
- cross visibility of cbrain-neurohub messages

@prioux @natacha-beck -

MontrealSergiy:
user to user messaging

### DIFF
--- a/Bourreau/spec/boutiques/boutiques_tester_spec.rb
+++ b/Bourreau/spec/boutiques/boutiques_tester_spec.rb
@@ -63,8 +63,9 @@ describe "Bourreau Boutiques Tests" do
       @boutiquesTask = SchemaTaskGenerator.generate(schema, descriptor)
       @boutiquesTask.integrate if File.exists?(descriptor)
       # Create a new instance of the generated task class
-      @task          = CbrainTask::BoutiquesTest.new
-      @task.params   = {}
+      @task             = CbrainTask::BoutiquesTest.new
+      @task.tool_config = ToolConfig.new
+      @task.params      = {}
       # Assign it a bourreau
       resource = RemoteResource.current_resource
       @task.bourreau_id = resource.id
--- a/BrainPortal/app/controllers/tool_configs_controller.rb
+++ b/BrainPortal/app/controllers/tool_configs_controller.rb
@@ -280,7 +280,7 @@ class ToolConfigsController < ApplicationController
     params.require(:tool_config).permit(
       :version_name, :description, :tool_id, :bourreau_id, :env_array, :script_prologue, :script_epilogue,
       :group_id, :ncpus, :container_image_userfile_id, :containerhub_image_name, :container_index_location,
-      :container_engine, :extra_qsub_args,
+      :container_engine, :extra_qsub_args, :singularity_overlays_specs,
       # The configuration of a tool in a VM managed by a
       # ScirCloud Bourreau is defined by the following
       # parameters which specify the disk image where the
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -427,8 +427,22 @@ class ClusterTask < CbrainTask
     base_dir = start_dir || self.full_cluster_workdir
 
     # Fetch and sync the requested userfile
-    userfile      = Userfile.find(userfile) unless userfile.is_a?(Userfile)
-    userfile.sync_to_cache
+    userfile = Userfile.find(userfile) unless userfile.is_a?(Userfile)
+
+    # Detect if we're trying to access a userfile content that is
+    # containerized in a singularity overlay.
+    is_local = nil
+    if self.tool_config
+           .data_providers_with_overlays
+           .map(&:id)
+           .include? userfile.data_provider_id
+      userfile_path = Pathname.new(userfile.provider_full_path) # path inside container
+      is_local      = true
+      self.addlog("Input file '#{userfile.name}' is on a local Singularity overlay")
+    else
+      userfile.sync_to_cache
+      userfile_path  = Pathname.new(userfile.cache_full_path)
+    end
 
     # Compute the final absolute path to the target file symlink
     file_path     = Pathname.new(file_path.to_s)
@@ -438,14 +452,19 @@ class ClusterTask < CbrainTask
     # Pathname objects for the userfile and bourreau directories
     workdir_path   = Pathname.new(self.cluster_shared_dir)
     dp_cache_path  = Pathname.new(RemoteResource.current_resource.dp_cache_dir)
-    userfile_path  = Pathname.new(userfile.cache_full_path)
+
+    # Pathname for the target value of the symlink with subpath
     userfile_path += Pathname.new(userfile_sub_path) if userfile_sub_path.to_s.present?
 
     # Try to figure out if the DataProvider of userfile is local, or remote.
     # SmartDataProviders supply the real implementation under a real_provider method;
     # other providers don't have that.
-    userfile_real_dp = userfile.data_provider.real_provider rescue userfile.data_provider
-    is_local = userfile_real_dp.is_a?(LocalDataProvider)
+    if is_local.nil?
+      userfile_real_dp = userfile.data_provider.real_provider rescue userfile.data_provider
+      is_local = userfile_real_dp.is_a?(LocalDataProvider)
+      self.addlog("Input file '#{userfile.name}' is on a local provider")  if   is_local
+      self.addlog("Input file '#{userfile.name}' is on a remote provider") if ! is_local
+    end
 
     # Files on local data providers are symlinked directly
     if is_local
@@ -2311,23 +2330,39 @@ docker_image_name=#{full_image_name.bash_escape}
   # is the raw scientific bash script.
   def singularity_commands(command_script)
 
-    # Values we substitute in our script:
-    # 1) The path to the task's work directory
-    task_workdir        = self.full_cluster_workdir
-
-    # 2) The root of the DataProvider cache
-    cache_dir          = self.bourreau.dp_cache_dir
-
-    # 3) The root of the shared area for all CBRAIN tasks
-    gridshare_dir      = self.bourreau.cms_shared_dir
-
-    # 4) Basename of the singularity wrapper script
+    # Basename of the singularity wrapper script
     singularity_wrapper_basename = ".singularity.#{self.run_id}.sh"
 
-    # 5) More -B (bind mounts) for all the local data providers.
+    # Values we substitute in our script:
+    # Numbers in (paren) correspond to the comment
+    # block in the script, well below.
+
+    # (1) The root of the shared area for all CBRAIN tasks
+    gridshare_dir = self.bourreau.cms_shared_dir
+
+    # (2) The root of the DataProvider cache
+    cache_dir     = self.bourreau.dp_cache_dir
+
+    # (5) The path to the task's work directory
+    task_workdir  = self.full_cluster_workdir
+
+    # (3) More -B (bind mounts) for all the local data providers.
     # This will be a string "-B path1 -B path2 -B path3" etc.
     esc_local_dp_mountpoints = local_dp_storage_paths.inject("") do |sing_opts,path|
       "#{sing_opts} -B #{path.bash_escape}"
+    end
+
+    # (4) Overlays defined in the ToolConfig
+    # Some of them might be patterns (e.g. /a/b/data*.squashfs) that need to
+    # be resolved locally.
+    # This will be a string "--overlay=path:ro --overlay=path:ro" etc.
+    overlay_paths = self.tool_config.singularity_overlays_full_paths.map do |path|
+      paths = Dir.glob(path) # checks on the local file system
+      cb_error "Can't find any overlays matching '#{path}'" if paths.blank?
+      paths
+    end.flatten
+    overlay_mounts = overlay_paths.inject("") do |sing_opts,path|
+      "#{sing_opts} --overlay=#{path.bash_escape}:ro"
     end
 
     # Set singularity command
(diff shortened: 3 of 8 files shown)


BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+    @message && @message.destroy
+
+    redirect_to :action => :index
+  end
+
+  private
+
+  def message_params
+    params.require(:message).permit(:header, :description, :variable_text, :message_type, :read, :user_id, :expiry, :last_sent, :critical, :display, :send_email, :group_id, :sender_id)
```
prioux, line 123: Messages between users don't have to support all these attributes... some of them are system-only

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  # POST /messages
+  # POST /messages.xml
+  def create #:nodoc:
+    @message              = Message.new(message_params)
+    @message.message_type = :communication
+    @message.sender_id    = current_user.id
+    [:header, :description, :variable_text].each do |section|
+      if @message.send(section).include? '['
+        txt = 'Square brackets are not supported. Please use curly or round brackets only.'
```
prioux, line 58: Please put your error messages in validation code in the model; if you don't want to interfere with the CBRAIN code, at least just put this all in  a method that you invoke here. That way you can validate in update() too

BrainPortal/app/models/message.rb
```diff
@@ -36,6 +36,10 @@ class Message < ApplicationRecord
 
   attr_accessor :send_email
 
+  has_many                :resource_usage
+  before_save             :track_resource_usage_update
```
prioux, line 40: I'd use before_create, because otherwise we are going to get a count every time someone modifies a message.

BrainPortal/app/models/message.rb
```diff
@@ -329,4 +333,26 @@ def self.parse_markup(string) #:nodoc:
     arr.join.html_safe
   end
 
+  def track_resource_usage_update
+    if self.message_type == "communication"
```
prioux, line 337: Keep code structure simpler. Unindent like this:

```ruby
def method
   return true if condition_where_nothing_happens
   stuff
   stuff
end
```

BrainPortal/app/views/nh_messages/new.html.erb
```diff
@@ -0,0 +1,62 @@
...
+      <fieldset>
+        <%= f.label :variable_text, 'Message' %>
+        <%= f.text_area :variable_text, :rows => 5, :cols => 80 %>
+      </fieldset>
+      <div class="btn-section">
+        <%= f.submit "Send", :class=>"btn-solid primary" %>
+      </div>
+    </p>
+  </div>
+<% end %>
```
prioux, line 60: Indentation is all messed up; missing newline on last line

prioux (changes requested):
See comments

BrainPortal/app/controllers/nh_messages_controller.rb
MontrealSergiy: I will double check, maybe allow use square brackets, that was a kludge to disable markup abuse by regular users. In past I have situation when management would like to bend rules for few exceptional cases. For example, user has a strange email or nick name but if he is important client or an executive, rules can be bended.

natacha-beck:
On the `nh_messages` index page I get the following: 
Screen Shot 2020-10-02 at 11 34 19 AM

See the `from #<NormalUser:0x00007f87e73f0a40>`. I send a message to myself.

MontrealSergiy:
@natacha-beck I was testing with ruby 2.6 setup so it was rendering the name ... should test with old ruby I guess . I will fix that

prioux:
It's not related to Ruby. The template is wrong. It's at line 62 of index.html.erb : 

```ruby
            <p class="card-item"> from  <%= message.sender || "unknown" %> </p>
```

prioux:
In the CBRAIn messaging system we have more robust code, because sender can be nil:

```ruby
<%= mess.sender.try(:login) || "System" %>
```

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -120,6 +112,6 @@ def destroy #:nodoc:
   private
 
   def message_params
-    params.require(:message).permit(:header, :description, :variable_text, :message_type, :read, :user_id, :expiry, :last_sent, :critical, :display, :send_email, :group_id, :sender_id)
+    params.require(:message).permit(:header, :description, :variable_textent, :group_id)
```
natacha-beck, line 115: Did you join 2 attributes? `variable_textent`

MontrealSergiy: a typo

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    if @message.header.blank?
+      @message.errors.add(:header, "cannot be left blank.")
+    end
+
+    @group_id = params[:group_id]
+    if @group_id.blank?
+      @message.errors.add(:base, "You need to specify the project whose members will receive this message.")
+    elsif @message.errors.empty?
+      group = current_user.assignable_groups.find(@group_id)
```
natacha-beck, line 64: When I try to send a message from a standard user to another standard user, it fails at this line: 

> Couldn't find Group with 'id'=13 [WHERE `groups`.`id` IN (3, 7, 20, 23, 29, 36, 55, 56, 83)]

even if I was able to select this user from the list when sending the message

BrainPortal/app/models/message.rb
MontrealSergiy: changed. Not sure is it simple (cyclometric complexity, more difficult to add new branches, the least interesting case on top of the method description) yet shorter

BrainPortal/app/views/nh_messages/new.html.erb
MontrealSergiy: I tried, but might need hint what to change. I cannot guess what you template indentation expectation.

BrainPortal/app/controllers/nh_messages_controller.rb
MontrealSergiy: corrected

BrainPortal/app/models/message.rb
MontrealSergiy: ok

BrainPortal/app/controllers/nh_messages_controller.rb
MontrealSergiy: ok

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+    @messages_count  = @messages.count
+    @read_count      = @messages.where(:user_id => current_user.id, :read => true).count
+    @unread_count    = @messages.where(:user_id => current_user.id, :read => false).count
+    @page, @per_page = pagination_check(@messages, :nh_messages)
+    @pagy, @messages = pagy(@messages, :items => @per_page)
+  end
+
+  def new #:nodoc:
+    @message        = Message.new # blank object for new() form.
+    @message.header = "A personal message from #{current_user.full_name || current_user.login}"
```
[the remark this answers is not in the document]
MontrealSergiy: ok

MontrealSergiy:
> In the CBRAIn messaging system we have more robust code, because sender can be nil:
> 
> ```ruby
> <%= mess.sender.try(:login) || "System" %>
> ```

ok copied verbatim

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+    @read_count      = @messages.where(:user_id => current_user.id, :read => true).count
+    @unread_count    = @messages.where(:user_id => current_user.id, :read => false).count
+    @page, @per_page = pagination_check(@messages, :nh_messages)
+    @pagy, @messages = pagy(@messages, :items => @per_page)
+  end
+
+  def new #:nodoc:
+    @message        =   Message.new # blank object for new() form.
+    @message.header =   "A personal message from #{current_user.full_name.presence || current_user.login}"
+    @recepients     =   find_nh_contacts(current_user)
```
prioux, line 44: "recipient" is misspelled everywhere

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+    @recepients     |= find_nh_messages(current_user) && current_user.assignable_groups
```
prioux, line 45: What is the reason for this |= assignment to @recipients? You did not document find_nh_contacts so I do not even know what that method is supposed to return, and so why it needs to be modified here

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  # POST /messages
+  # POST /messages.xml
+  def create #:nodoc:
+    @message              = Message.new(message_params)
+    @message.message_type = :communication
+    @message.sender_id    = current_user.id
+
+    @recepients =   find_nh_contacts(current_user)
+    @recepients |= find_nh_projects(current_user) && current_user.assignable_groups || contacts
```
prioux, line 56: Somme comment as above

prioux: Also, `contacts` is not defined anywhere and it raises an exception

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    if @message.header.blank?
```
prioux, line 58: Put all your validations in a method in the model, not as a series of if else in the controller. Then just invoke the method here

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+      @message.errors.add(:header, "cannot be left blank.")
+    end
+
+    @group_id = params[:group_id]
+    if @group_id.blank?
+      @message.errors.add(:destination_id, "You need to specify the project whose members will receive this message.")
```
prioux, line 64: this error is added in the `if` part of an if elsif structure; the elsif itself is a test for if `errors` is empty (next line) so that means the errir message here will not prevent the test that follows on the next line. Simplify the structure

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+    @message.sender_id    = current_user.id
+
+    @recepients =   find_nh_contacts(current_user)
+    @recepients |= find_nh_projects(current_user) && current_user.assignable_groups || contacts
+
+    if @message.header.blank?
+      @message.errors.add(:header, "cannot be left blank.")
+    end
+
+    @group_id = params[:group_id]
```
prioux, line 62: You use params[:group_id] without any sort of validation so that means a hacker coudl send messages to anything on the system, even things that it doesn't have access.

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+    if @group_id.blank?
+      @message.errors.add(:destination_id, "You need to specify the project whose members will receive this message.")
+    elsif @message.errors.empty?
+      if @recepients.any? { |x| x.id == @group_id }
+        @message.send_me_to(Group.find(id))
```
prioux, line 67: When tring to send a message from admin to admin, the code never even entered this statement, so no message was ever sent.

prioux: The `id` method raises an exception.

Extract your relevant pieces of information into variables early in the method's code

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+
+        @message.errors.add(:destination_id, "Invalid message destination.")
+      end
+    end
+    prepare_messages
+    if @message.errors.empty?
```
prioux, line 73: indentation messed up

BrainPortal/app/controllers/nh_messages_controller.rb
prioux: Also consider giving a new name to this param. `group_id` is way too generic, call it `destination_group_id` or something so that we can distinguish that param from the the other param we normally use to specify the project.

BrainPortal/app/models/message.rb
```diff
@@ -329,4 +350,30 @@ def self.parse_markup(string) #:nodoc:
     arr.join.html_safe
   end
 
+  # before_destroy callback
+  def track_resource_usage_create
```
prioux, line 354: use `return true if cond`

BrainPortal/app/models/message.rb
```diff
@@ -329,4 +350,30 @@ def self.parse_markup(string) #:nodoc:
...
+    if self.message_type != "communication"  # only user to user communication is counted so far
+      return true
+    end
+    CountResourceUsageForUserMessage.create(
+      :value              => 1,
+      :user_id            => self.sender_id,
+      :group_id           => self.group_id
```
prioux, line 361: This doesn't do what you think. The 'group_id' attribute in a message is only used for the invitation system. It is not associated with the sender or receiver.

BrainPortal/lib/neurohub_helpers.rb
```diff
@@ -145,4 +145,16 @@ def neurohub_search(token, limit=20, user=current_user)
...
   end
 
+  # filter NeuroHub messages (no invites)
+  def find_nh_messages(user=current_user)
+    Message.where(:user_id        => user.available_users.map(&:id),
+                  :type           => nil  #  filter out invites
+                  ).order( "last_sent DESC" )
+  end
+
+  def find_nh_contacts(user)
```
prioux, line 155: This is totally not efficient; please don't reconstruction a bunch of objects with all their attributes just to extract their ids, given that rails already provide methods to get the association IDs or pluck individual attribute without such reconstruction.

prioux: Also document what this method does. I am not even sure what it does by looking at the code. And why in the controller code every time it is invoked, some adjustments are always made to the results

prioux: Consider running your statements in the console pieces by pieces to see how inefficient they are.

prioux (changes requested):
Serious issues. Doesn't work. Code is buggy. Restructuring needed.

PR #1039, opened 2020-10-09
neurohub messaging  #986

MontrealSergiy:
note I heavily reworked code for neurohub user messaging. I summited it for review in this new pull request. Please share you comments


prioux:
I'm getting layout problems with the messages. See screenshot:

Screen Shot 2020-10-21 at 13 19 38

prioux:
The messaging form contains a textare for the 'variable_text' attribute; that is not necessary. That text space is mostly used by system messages that append notifications to existing messages (a concept that doesn't exist in mormal mail). The user interface for users should only ask for a subject and a body (attributes: header and description, respectively).

Screen Shot 2020-10-21 at 17 02 01

prioux:
I created a brand new user that doesn't share anything with anyone, and the 'new message' button doesn't work at all. I suspect the icon for 'You currently have no messages' is interfering, because it's overlapping the rest of the interface including in places it's not supposed to go.

Screen Shot 2020-10-21 at 17 12 43

prioux:
As an admin, I wasn't able to see at all the new user I created (see previous comment) at all as a possible recipient. I don't understand why.

prioux:
So I added manually a message for my user 'noshare' with the console:

```ruby
Message.send_message(User.find(103), :header => 'Sub', :description => "Hello\r\nBye\r\n", :sender_id => 1)
```

Now as user 'noshare' I can see the message in the interface,a nd I can click the 'New Message' button, except the interface shows a select box for destinations that is totally empty. There shoudl be at least one user there: 'noshare' itself.

prioux:
I have created the method to generate the list of groups that a user can write messages to (people or projects). Please add this method and use it to 1) generate the form for new messages 2) validate that the destination group_id is correct when the POST is received. Remove all the methods in `neurohub_helpers.rb` that you've added in this PR, they are now obsolete.

```ruby
  def find_nh_message_recipients(user = current_user)
    # Projects that the user can assign to resources
    projs = find_nh_projects(user)
    projs = ensure_assignable_nh_projects(user, projs)
    # List of users in these projects
    user_ids    = (projs.to_a.map(&:user_ids).flatten + [user.id]).uniq
    user_logins = User.where(:id => user_ids).pluck(:login)
    # Same list as a set of UserGroup
    user_projs  = UserGroup.where(:name => user_logins)
    # The possible destinations are described by a set of Groups
    projs.to_a + user_projs.to_a
  end
```

prioux:
For `find_nh_messages(user = current_user)` is there any reason why not simply replace this by

```
def find_nh_messages(user = current_user)
   Message.where(:user_id => user.id)
end
```

That will extract all the messages that is directly assigned to the current user, and nothing else, which is good enough even for an admin.

prioux:
Still serious layout problems. See screenshot. Can you please test this on your side with a selection of common system messages?

Screen Shot 2020-10-27 at 13 57 24

prioux:
When the header line is long it gets truncated with '...' and we can never see the rest.

Screen Shot 2020-10-27 at 14 17 45

BrainPortal/app/models/message.rb
```diff
@@ -329,4 +354,24 @@ def self.parse_markup(string) #:nodoc:
...
+  def track_resource_usage_create
+    return true if self.message_type != "communication"  # only user to user communication is counted so far
+    CountResourceUsageForUserMessage.create(
+      :value              => 1,
+      :user_id            => self.sender_id
+    )
+    true
+  end
+
+  # before_destroy callback
```
prioux, line 367: We don't need to track destroy events. please remove this callback

BrainPortal/app/models/message.rb
```diff
@@ -329,4 +354,24 @@ def self.parse_markup(string) #:nodoc:
     arr.join.html_safe
   end
 
+  # before_destroy callback
+  def track_resource_usage_create
+    return true if self.message_type != "communication"  # only user to user communication is counted so far
+    CountResourceUsageForUserMessage.create(
+      :value              => 1,
```
prioux, line 361: why not put something more useful in the value field? Like the size of the message? (sum of header, description and var_text)

BrainPortal/app/controllers/nh_messages_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+    @message && @message.destroy
+
+    redirect_to :action => :index
+  end
+
+  private
+
+  def message_params
+    params.require(:message).permit(:header, :description, :variable_text)
```
prioux, line 93: remove variable_text if user can't provide it anyway in the form

BrainPortal/app/views/nh_messages/index.html.erb
```diff
@@ -0,0 +1,92 @@
...
+                <p class="card-item"><em> <%= header %> </em></p>
+              <% end %>
+              <% if message.read %>
+                <p class="card-item"> Read </p>
+              <% else %>
+                <p class="card-item"> Not read </p>
+              <% end %>
+              <p class="card-item"> from  <%= message.sender.try(:login) || "system" %> </p>
+              <div class="card-item">
+                <%= button_to 'Delete', nh_message_path(message.id), :method => :delete, :class => "btn-solid secondary wash btn" %>
```
prioux, line 67: The delete button doesn't ask for any confirmation message at all

BrainPortal/app/views/nh_messages/new.html.erb
```diff
@@ -0,0 +1,62 @@
...
+                            :groups   => @recipients,
+                            :selector => @destination_group_id, # :group_id is not an attribute of Message.
+                          },
+                          {:include_blank => true}
+            )
+         %>
+      </fieldset>
+      <fieldset>
+        <%= f.label :header %>
+        <%= f.text_field :header, :size => 92 %>
```
prioux, line 48: put a limit on the size of the content too, I get an exception if I put too much text in the header or body

BrainPortal/config/routes.rb
```diff
@@ -312,6 +312,8 @@
         post :upload_file
       end
     end
+    resources :nh_messages,        :except => [ :edit, :show ] do
```
prioux, line 315: Why do you put an empty block? You don't need this. Look at other examples in the file

prioux (changes requested):
Bugs and fixes.

prioux:
I've tested the PR quite a bit with a number of fake and less fake users on my dev machine, and if these last 8 issues are fixed I will agree to merge it.

BrainPortal/app/models/message.rb
MontrealSergiy: Why not? Do we have a way to see how Delete button being used.

BrainPortal/app/models/message.rb
MontrealSergiy: Agree, the count is the goal, but sizes could be of potential interest for reporting and ux studies. I guess we still can count message numbers. But are any downsides like any issues with it if adding message count to official report?

prioux:
All issues have been addressed.

PR #1040, opened 2020-10-13
[ADD] Launch_task div on Userfiles index when params tool_id is set

natacha-beck:


natacha-beck:
Was done to allow the following behavior on CONP: 
- https://github.com/CONP-PCNO/conp-portal/issues/347

MontrealSergiy:
Confirm the above message indeed appears. I would say 'first select appropriate input( or output) files (or collection) and launch ... ' - user does not select arbitrary files but those relevant to the tool.

MontrealSergiy (changes requested):
a space between message and the tool name is missing 
image

BrainPortal/public/javascripts/userfiles.js
```diff
@@ -401,6 +401,72 @@ $(function() {
...
+          var nb_selected_files = parseInt($('.psel-count').text())
+          var have_selection    = nb_selected_files > 0
+
+          var file_status_text  = have_selection ? "Launch with " + nb_selected_files + " file(s)" : "No files selected"
+
+          $('.launch_bar').remove();
+
+          $('#menu_bar').after(
+            $('<div class="launch_bar">')
+              .append($('<span class="info">Select some files to launch' + tool_name + '</span>'))
```
prioux, line 438: See Sergiy's comment

BrainPortal/public/javascripts/userfiles.js
```diff
@@ -401,6 +401,72 @@ $(function() {
 
     /* Show/Hide dynamic actions/menu elements according to current selection */
     (function () {
+      function launch_task() {
+        var parameters    = window.location.search.split(/\?|&/);
```
prioux, line 405: There must exist query parameter APIs or helpers functions, at least in jQuery ? Also, given the page is rendered as a consequence of a GET, instead of duplicating all this code from userfiles/launch_task.js, why not simply trigger the behavior with one synthetic event?

prioux (changes requested):
Simplify this, please

BrainPortal/public/javascripts/userfiles.js
```diff
@@ -401,6 +444,42 @@ $(function() {
...
+            var name                = tool.childNodes[0].data;
+                name_by_tool_id[id] = name;
+          }
+          var tool_name   = name_by_tool_id[tool_id]
+
+          if (tool_name === undefined) {
+            continue;
+          }
+
+          /* Do we have some files selected to launch the task on? */
```
prioux, line 471: Why do you do this IN THE LOOP

The loop should be just about finding the tool_id, and as soon as it's found, to `break` out of.

natacha-beck:
Replace by https://github.com/aces/cbrain/pull/1047.

PR #1042, opened 2020-10-15
Empty invite fix #1038

### LINKED ISSUE #1038: Not selecting any users when sending an invitation raises an exception
In the CBRAIN invite form when not selecting any user and clicking the invite button, on the controller side we get `params[:user_id]` to be nil, so the map() there crashes.

MontrealSergiy:


BrainPortal/app/controllers/invitations_controller.rb
```diff
@@ -48,18 +48,21 @@ def new #:nodoc:
...
+    elsif rejected_ids.present?
       flash_message = "\n#{User.find(rejected_ids).map(&:login).join(", ")} already invited."
     end
 
     @users = User.find((user_ids - already_sent_to) & current_user.visible_users.map(&:id))
 
     unless @group.can_be_edited_by?(current_user) && @users.present?
       flash[:error]  = "Could not send the requested invitations."
-      flash[:error] += flash_message if rejected_ids.present?
+      flash[:error] += flash_message if defined? flash_message
```
prioux, line 65: Don't use `defined?` it will return true even if the variable isn't set to a true value. Why not simply `if flash_message.present?`

prioux (changes requested):

BrainPortal/app/controllers/invitations_controller.rb
MontrealSergiy: do not see problem if set to non-true value. But if to confusing or looks hackish can refactor with present, no problem.

prioux: There IS a problem. It's a bug. defined?() returns the status of a variable as a string. A string is true. Try it yourself:

```
shell> cat x.rb 
if false
  mess = 'xyz'
elsif nil
  mess = 'xyz2'
end
puts "doing this" if defined? mess
shell> ruby x.rb 
doing this
```

prioux: The condition for appending the flash_message is not whether or not it's defined, it's whether or not it's present. These are different. nil is defined, but not present. If you take this line just by itself with no other context you shoudl be able to see it can raise an exception:

```ruby
      flash[:error] += flash_message if defined? flash_message
```

when flash_message is nil, because you're going to try do string concatenation with nil.

prioux (changes requested):
Bug

BrainPortal/app/controllers/invitations_controller.rb
MontrealSergiy: ok got it now

MontrealSergiy: BTW there is some dead code inside the method (the last branching) that I can get rid of

prioux: Yes remove the dead code. Good catch.

MontrealSergiy: fixed

PR #1047, opened 2020-10-26
Launch task div

natacha-beck:


BrainPortal/public/javascripts/userfiles.js
```diff
@@ -401,6 +439,45 @@ $(function() {
...
+        var prepare_tool_id = undefined
+        for (var i = 0; i < parameters.length; i++) {
+          var [name, id]     = parameters[i].split(/=/);
+          var userfile_checkboxes = $("input[name='file_ids[]']");
+
+          if (name != "prepare_tool_id") {
+            continue;
+          }
+
+          prepare_tool_id = id;
```
prioux, line 455: Once you found the tool_id, break out of the loop instead of continuing to scan all the other params

BrainPortal/public/javascripts/userfiles.js
```diff
@@ -401,6 +439,45 @@ $(function() {
...
+        }
+
+        if (!prepare_tool_id) {
+          return;
+        }
+
+        var name_by_tool_id = {};
```
prioux, line 462: Lines 462-468: why do you create alookup table of ALL `{ dds : names }` just to use it once for one particular ID that you already know? Just do like the loop above: as soon as you found the ID, set the name in a variable. No need to build `name_by_tool_id` at all.

prioux (changes requested):
Cleanup

Issue #1049, opened 2020-11-06
Boutiques taks failed when optional output is missing. 


natacha-beck:
For example, FSL bet failed when run without the `-o` option. Running without `-o` will not produce the overlay output. 
Then the task will fail on CPU.

prioux:
Fixed by #1050

PR #1050, opened 2020-11-06
[FIX] for optional output file in Boutiques

natacha-beck:
Boutiques task fail on CPU when optional output is missing (see #1049). 

If everything is OK with this PR, it should be merge on Master (hot fix) and deploy on Bourreau.

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
@@ -494,7 +494,7 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
     # Helper for checking whether outputs are optional
 %   optionalFiles = outputs.select { |o| o['optional'] }
     isOptional = lambda do |f|
-      [<%= optionalFiles.map{ |x| ":#{x['id'].to_sym}" }.join(",") %>].include?(f)
+      [<%= optionalFiles.map{ |x| ":'"+x['id']+"'" }.join(",") %>].include?( f.to_sym )
     end
 
     # Save (and register) all generated files to the results data provider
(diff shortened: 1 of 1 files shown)


BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -494,7 +494,7 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
     # Helper for checking whether outputs are optional
 %   optionalFiles = outputs.select { |o| o['optional'] }
     isOptional = lambda do |f|
-      [<%= optionalFiles.map{ |x| ":'"+x['id']+"'" }.join(",") %>].include?(f)
+      [<%= optionalFiles.map{ |x| ":#{x['id'].to_sym}" }.join(",") %>].include?(f)
```
prioux, line 497: The original code properly generated symbols with single quotes around the symbol name to allow for IDs that contains non-alpha characters, your new code doesn't do that.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -505,8 +505,8 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
         # Print a warning if the file is not present
         # Note that, since the output may be optional, the process may still be successful
         unless path.present? && File.exists?(path)
-          self.addlog("Unable to find optional output file: #{path}") if isOptional.(param)
-          succeeded = false if ! isOptional.(param)
+          self.addlog("Unable to find optional output file: #{path}") if isOptional.(param.to_sym)
```
prioux, line 508: Instead of requiring the user of the lambda to invoke `to_sym` on the arguments each and every time (like you did here), why don't have the lambda invoke `to_sym` itself? The lambda is it's own domain, so it know what its internal types are (the array of symbols) and it know what transformations to do. You, as a user of hte lambda, should not have to know that.

prioux (changes requested):

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
prioux: Not only that, but your code is buggy, it basically does a thing-to-symbol-to-string-to-symbol transformation.

PR #1059, opened 2020-11-23
Tool config data immutability flag resolves #1054

### LINKED ISSUE #1054: Allow the admin to indicate that task can modify its input in a ToolConfig
Right now the CBRAIN task launching interface will refuse to launch a task when:

1. the user has selected a file that they don't own
2. the file's `group_writable` attribute is set to `read-only`
3. the task doesn't say explicitly that it doesn't modify its inputs

The behavior of point 3 is the default. A task can inform the framework that it doesn't modify its inputs by having either:

* `task_properties :readonly_input_files` in `portal/taskname.rb` (standard integration)
*  `"custom": { "cbrain:readonly-input-files": true }` (Boutiques integration)

Given that often integrators forget to set these properties, the admin should have the ability to override the behavior directly in a ToolConfig (or maybe directly at the Tool level?). This would be provided by a simply checkbox in the ToolConfig form, associated with a new boolean attribute:

```
  [ ] Does not require write-access to inputs
```

MontrealSergiy:


BrainPortal/app/views/tool_configs/_form_fields.html.erb
```diff
@@ -67,6 +67,14 @@
         </div>
       <% end %>
 
+      <% t.edit_cell(:inputs_readonly, :header => "Read only input files", :show_width => 2) do |f| %>
```
prioux, line 70: Within the web page, let's not use the 'inputs_readonly' phrasing, it's super ugly. Let's show this as 'Does not modify its inputs files'

BrainPortal/app/views/tool_configs/_form_fields.html.erb
```diff
@@ -67,6 +67,14 @@
...
+        <%= f.check_box :inputs_readonly %>
+        <div class="wide_field_explanation">
+          Are you sure that input files are not mutated by any task? It will allow execution with data
```
prioux, line 73: Next text:

```
Check this if the tool is known <strong>not to modify its input files</strong>. This will allow a user to launch the tool on files that are not marked as group-writable in the file manager.
```

BrainPortal/db/migrate/20201123133913_add_inputs_readonly_to_tool_config.rb
```diff
@@ -0,0 +1,5 @@
+class AddInputsReadonlyToToolConfig < ActiveRecord::Migration[5.0]
```
prioux, line 1: Add license header. Check other migration for examples.

prioux (changes requested):
Some minor changes.

prioux:
I just realized that the Boutiques integrator is not checking that new flag.

Please adjust `BrainPortal/liv/cbrain_task_generators/templates/portal.rb` so that it looks up that values. Look for `readonly_input_files` in the code:

```
./portal.rb.erb:103:      :readonly_input_files => <%=
./portal.rb.erb:120:    @_file_access ||= (self.class.properties[:readonly_input_files].present? ? :read : :write)
```

There is a complexity there. At line 103 the template sets a constant value in the CLASS based on what is found in the descriptor. We can't adjust that, because at the class level we have no way to find out the configuration at run time that is in the tool config.

But at line 120 is where the run-time code looks it up, so basically we need the code to check the property in in the tool config there. It will be tru if it's true EITHER in the descriptor, or in the tool config.

prioux (changes requested):
One more change requested, see discussion panel.

PR #1070, opened 2020-12-21
Stop assigns to ro projects

MontrealSergiy:


MontrealSergiy:
close due to travis message, will resubmit once fixed

Issue #1072, opened 2021-01-04
The persistent selection mechnism doesn't respect the "own files vs all available files" state


prioux:
In the file manager, users can toggle between "see only my files" vs "see all available files".

If the "select all files" link is pressed, and the user is in "see only my files" mode, the files selected will still be all the available files.

See screenshot, which has all the numbers (a filter is applied on the DP, and there are 131530 files on that DP, though only 125 belong to the user):

![Screen Shot 2021-01-04 at 14 35 56](https://user-images.githubusercontent.com/777588/103572449-9532da00-4e9a-11eb-8a50-39502e376e35.png)

Discovered on Safari because in Chrome and Firefox issue #1071 is still active.

MontrealSergiy:
Not sure it is related one thing I hear the quotes for storage changed https://chromium.googlesource.com/chromium/src/+blame/refs/heads/master/storage/browser/quota/quota_settings.cc

MontrealSergiy:
@natacha-beck Just curious do you observe the same bug you described in Safari, if so what version?

Issue #1071, opened 2021-01-04
The persistent selection mechanism is broken


prioux:
The link 'select all' when you have several pages of userfiles no longer selects all the files on all pages. In fact it seems to not select anything for real.

Tested on both Firefox and Chrome. I suspect something changed in the browser APIs. The firefox console logs errors about access to the persistent selection database. On the server side I verified and the controller code properly returns a list of all IDs when a GET to `/userfiles?only_ids=1` is performed by the JS code in the browser.

prioux:
It works perfectly fine in Safari by the way.

natacha-beck:
Works fine on chrome and firefox, on my side on Ubuntu even in incognito mode.

natacha-beck:
Never mind it seems to work but it doesn´t.

MontrealSergiy:
Last year indeed there were few changes in chrome/firefox local storage api, I hear the sync and quotes https://chromium.googlesource.com/chromium/src/+blame/refs/heads/master/storage/browser/quota/quota_settings.cc there are of course some minor changes, syncing, yet they lead to data loss only at concurrent conflicting requests I think https://blog.mozilla.org/addons/2020/07/09/changes-to-storage-sync-in-firefox-79/

MontrealSergiy:
The bug shows on my Safari btw

![image](https://user-images.githubusercontent.com/11949081/105368695-cbc56000-5bcf-11eb-88df-4a941db0713e.png)

MontrealSergiy:
@prioux while I observe problems, especially on Chrome, I am wondering what exactly you meant 'it seems to not select anything for real'. On my Chrome 'select all' seem select all files, at the least count them, though I cannot count all _my_ files .

prioux:
The screenshot above is a symptom of the other issue, #1072 

This issue was about not being able to select all files on all pages, and I can't reproduce it. So I will close this. All further discussions about the persistent selection mechanism should go to issue #1072 or the associated PR, #1073

PR #1073, opened 2021-01-06
[FIX] api_request when request came from jQuery

natacha-beck:


natacha-beck:
Could fix the #1071

MontrealSergiy:
it seems fix #1072 at the least for chrome yet not fully fixes #1071 (it allows one to chose his or all files by clicking 'select all' even though when you switch from all/my files or back the counter is not updated) 

image

prioux:
@natacha-beck Is the CSV issue mentioned by @MontrealSergiy fixed ?

natacha-beck:
I need to double check, I do not remember getting this issue when I tested
it.

MontrealSergiy:
Actually is CSV is addressed by the PR, but not CBCSV (or probably any other operation).

natacha-beck:
@MontrealSergiy I am looking at it.

natacha-beck:
The CBCSV works too. From my point of view the only missing part is the update of the count, I am still on it.

natacha-beck:
Let do a recap: 

For issue #1072, the bug is the following: 

> If the "select all files" link is pressed, and the user is in "see only my files" mode, the files selected will still be all the available files.
> See screenshot, which has all the numbers (a filter is applied on the DP, and there are 131530 files on that DP, though only 125 belong to the user):

So when the user is in "see only my files" mode, when the user clicks on 'select all files' it selected all the files available to the user (the one owned by the user + the one available to the user). **Normally it should not be the case with my PR**.

I feel like one of your test @MontrealSergiy  was to be in "Show all files" mode, then click on "select all", then switch to "Show only my file". In this case, I do not expect the selection of files to change (after all you just ask to change the view). For example, the persistent selection file is not updated every time you update a filter in the index page. 


For the following point: 

> Curiously, on the above snapshot, even though it says '5 files selected' link for csv generation will download list of 3 files but for other operations, such building cbrain-csv picks will contain all the files. 

Indeed but it was like that in CBRAIN or it should have been like that. The `Export to CSV` functionality was here before the `Create a CBCSV file` both functions have a different purpose:  

- **Export to CSV**: will create a list based on what is available on the index page. So if you have 3 files it will create a list with 3 entries.
- **Create a CBCSV file**: will create a list based on what is available in your persistent selection (that can differ from what you have in your index).

MontrealSergiy:
ok then,

prioux:
This PR seems to work fine in every cases I tried

prioux:
Except: the 'export to csv' function now exports all pages, in the past it exported only the current page. Not sure if that's what we want.

natacha-beck:
Indeed it is a modification I have done I can revert it if you want. Was
not sure about what we expected form the `Export to CSV`.

prioux:
So the export to CSV behavior is caused by this, right?

https://github.com/aces/cbrain/pull/1073/files#diff-f57dc1e6518dd2853ec0f3c228ba9bbb8cc7aa9e57faaa74a226ff958cbc6fc4R132

In that case, if you revert that single line change, I will merge.

prioux:
As a matter of clarification for @MontrealSergiy, the list of selected files persistently shoudl not really change if you change the mode: what is selected is what is selected, and that allows users to create a larger list by selecting things during multiple rounds of filtering the view and clicking 'select all'.

PR #1079, opened 2021-01-19
New GitHub actions CI

prioux:
This PR replaces the old Travis CI hooks with a GitHub Actions workflow.

Please have a look at the files inside `.github/workflows`.

Reviewers: **Do not merge this PR!** I will merge it myself using a `squash` merge. While developing this I generated 48 commits (!) and they don't need to be part of the CBRAIN codebase.

The entire process of developing this GitHub actions configuration is instructive: if you want, consider looking at each of my commits in order (and maybe even looking at the result of the workflow too). You will see how I built and tested each feature and the mistakes I made and how I fixed them.


prioux:
Ah it's too bad, it seems this PR will not itself generate a CI test because the dev branch doesn't have the workflow configured, yet.

natacha-beck (approved):

Issue #1055, opened 2020-11-16
Creating a CbrainFileList sometimes assign it to a project that the user should not be able to write to.


prioux:
The code simply uses:

```ruby
current_project.id
```

that's too general.

Also: check all instances of such code? (including `current_project.try(:id)`)

prioux:
This is a consequence of NeuroHub, by the way, That bug wasn't possible before.

MontrealSergiy:
@prioux also new task and file registration try to assign to the current project. Fix within this issue?

prioux:
Yes, good thinking.

Try to make sure the code use the proper `_ids` methods we have defined (assignable_group_ids etc) if possible/relevant.

PR #1084, opened 2021-01-25
no default targeting to readonly active aka current groups #1055

MontrealSergiy:


BrainPortal/lib/session_helpers.rb
```diff
@@ -52,6 +52,11 @@ def current_project
     @current_project
   end
 
+  # return currently active project id if user can assign to it. Serves as default destination project for many operation
+  def current_assignable_project_id
+    return current_project.try(:id) && current_user.assignable_group_ids.include?(current_project.id)
```
prioux, line 57: This returns `nil` or `true` ... not an ID.

Was this even tested?

prioux:
Other general comment:

Given that everywhere that the new method `current_assignable_project_id` is invoked, it's in a statement like

```
current_assignable_project_id || current_user.own_group.id
```

why not make the method do itself the lookup to `current_user_own_group.id` when no other ID works?

MontrealSergiy:
sure, correcting

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -89,7 +89,7 @@ def new #:nodoc:
   def create #:nodoc:
     @provider = DataProvider.sti_new(data_provider_params)
     @provider.user_id  ||= current_user.id # disabled field in form DOES NOT send value!
-    @provider.group_id ||= (( current_project && current_project.id ) || current_user.own_group.id)
+    @provider.group_id ||default_target_group_id
```
prioux, line 92: This is wrong; this used to be an assign statement `||=` and you changed it to a NOOP `||`

BrainPortal/app/controllers/bourreaux_controller.rb
```diff
@@ -88,7 +88,7 @@ def info #:nodoc:
   end
 
   def new #:nodoc:
-    bourreau_group_id = ( current_project && current_project.id ) || current_user.own_group.id
+    bourreau_group_id = default_target_group_id
```
prioux, line 91: Why has the method's name changed to something that is even LESS descriptive than what it was before? `current_assignable_group_id` was perfectly fine.

BrainPortal/lib/session_helpers.rb
```diff
@@ -52,6 +52,15 @@ def current_project
     @current_project
   end
 
+  # return currently active project id if user can assign to it or own group id. Serves as default destination project for many operation
+  def default_target_group_id
+    if current_user.assignable_group.include? current_project
```
prioux, line 57: This crashes. There is no method `assignable_group` . This means you have not even tried your code at all.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -76,7 +76,7 @@ def show  #:nodoc:
   end
 
   def new #:nodoc:
-    provider_group_id = ( current_project && current_project.id ) || current_user.own_group.id
+    provider_group_id = default_target_group_id || current_user.own_group.id
```
prioux, line 79: Why leave the `|| current_user.own_group.id` here given the logic is in the method `default_target_group_id` ?

prioux (changes requested):
Unacceptable crashing code. See comments.

MontrealSergiy:
I corrected bugs.

BrainPortal/lib/session_helpers.rb
MontrealSergiy: I was sure I fixed that typo, sorry

BrainPortal/app/controllers/bourreaux_controller.rb
MontrealSergiy: If it was better I revert. My line of thought was as follows: `default_project = current_project || current_user.own_group`  is already in codebase, and I thought it would be more descriptive, not in sense how method work but what goal it achieves. Also it is no always current, but sometimes just an assignable branch.  Having a different name might one day be a step to enhance logic behind defaulting target.

BrainPortal/app/controllers/data_providers_controller.rb
MontrealSergiy: corrected

MontrealSergiy:
@natacha-beck do I read you right, whatever the is public project (assignable or not), whatever is user (project owner, member etc) cbcsv file go to own project instead of the current? Do you want same rules for file copying, creating collections, tasks, data providers etc, or just cbcsv

MontrealSergiy:
@prioux I just realized we better fix `default_project = current_project || current_user.own_group` as well. It should not take too much time.

natacha-beck:
I just need the cbcsv file to be fixed for now.

prioux:
@natacha-beck Are you saying that even after all these changes, the cbcsv files are still created in the wrong project?

MontrealSergiy:
now I also committed a refactoring for tag (in extra option). It is more decorative and does not seem change much the form behavior, where the first assignable workgroup is used as default group selected. @prioux @natacha-beck

MontrealSergiy:
> I just need the cbcsv file to be fixed for now.

too late,  I protected public folders for other objects too.

natacha-beck:
Just tested it, it works when I am connected as a regular user. If I am in a public project and I create a CBCSV the project of the file will be the `user project`. 

BUT when I am connected as an admin in a public project and I create a CBCSV the project of the file will be the `public project`.

MontrealSergiy:
> Just tested it, it works when I am connected as a regular user. If I am in a public project and I create a CBCSV the project of the file will be the `user project`.
> 
> BUT when I am connected as an admin in a public project and I create a CBCSV the project of the file will be the `public project`.

Yes. Do you think that admins cbcsv should go to user's own project or it's fine as is?

natacha-beck:
Yes it should go the user project even if it is an admin. I do not want the public project to be populated with CBCSV file.

prioux:
Given that when someone created a cbcsv, the user is redirected to the show page of the cbcsv file, we should add a flash message that says at the top "This CbrainFileList file was created in the project 'XYZ'; please double-check below that this is where you want it".

natacha-beck:
Good point.

prioux:
In fact there is already a message:

"Created file list named '#{file_list.name}'."

Let's add the project name in the message.

MontrealSergiy:
> Yes it should go the user project even if it is an admin. I do not want the public project to be populated with CBCSV file.

so CBCSV should _never_ go to a public project.

I already have a branch which does just that, yet additionally protect public projects from any other new files or associations like copying, creating collections, tasks, data providers.

Would you like that or prefer for now concentrate on protecting public project from cbcsv only?

@natacha-beck

natacha-beck:
I would like to have one or the other for tomorrow evening. So as you
wish/can.

MontrealSergiy:
> I would like to have one or the other for tomorrow evening. So as you wish/can.


I can do either before that.

MontrealSergiy:
@natacha-beck I understand no mater what original request was, now you prefer that same rules are both for non-assignable and assignable public projects. Correct?

natacha-beck:
Yes, for me a CBCSV file should not go in a PUBLIC project even if it is
assignable, at least not by default.

MontrealSergiy:
@natacha-beck @prioux Notice is improved, admins blocked from adding cbcsv. Please review the change.

If you are curious there is MontrealSergiy:no_ro_public_assign that further prohibits attaching other things to public projects.

BrainPortal/app/helpers/select_box_helper.rb
```diff
@@ -613,5 +613,14 @@ def uploadable_data_providers(user=current_user)
     return uploadable_data_providers
   end
 
+  # returns currently active project if user can assign to it or own group id. Serves as default destination project
```
prioux, line 616: I don't know why this method is in this file instead of `session_helper.rb`; move it there

Also, make one invoke the other so that you do not duplcaite code; e.g. the _group method has the logic and the _group_id method invokes the group method and then calls `.id`

BrainPortal/lib/session_helpers.rb
```diff
@@ -52,6 +52,15 @@ def current_project
     @current_project
   end
 
+  # return currently active project id if user can assign to it or own group id. Serves as default destination project for many operation
```
prioux, line 55: Use a nice uppercase letter to start your comment block? Split it to make it shorter a little?

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -1333,16 +1333,17 @@ def export_file_list
...
       :data_provider_id => dest_dp_id,
-    )
+      )
 
     # Save it and set its content.
     if file_list.save
       csv_text = CbrainFileList.create_csv_file_from_userfiles(userfiles)
       file_list.cache_writehandle { |fh| fh.write(csv_text) }
-      flash[:notice] = "Created file list named '#{file_list.name}'."
+      flash[:notice] = "Created file list named '#{file_list.name}' in '#{file_list.group.name}' project."\
```
prioux: Why the backlash string continuation? It's very ugly. We never use this in CBRAIN

prioux (changes requested):
Seems functional; requesting some minor adjustments.

BrainPortal/app/helpers/select_box_helper.rb
MontrealSergiy: Somehow it did not work from template, I can try again but
@prioux  what about dropping the whole tag template commit? That code ('default_project') is dead and affect nothing.

prioux: I don't understand what you are talking about. What tag template? What dead code? Can you please explain?

MontrealSergiy: the method `current_assignable_group` is used only to set variable `default_project`, the rest of code use `current_assignable_group_id`. While that variable is nominally used in the template `dialogs.html.erb` , however it does affect the behavior ( tag is defaulted to the first assignable workgroup ). So I suggested to remove the changes related to that template and method. That is revert commit

https://github.com/MontrealSergiy/cbrain/commit/6873f053818cad6a3a9ca89ffa9947e3be48a27a

prioux: Sure, revert that commit if you want.

prioux: I still don't know which 'dead code' you were referring to... I don't see any dead code.

MontrealSergiy: You are right, then, my bad. And I am doing as you suggested. 

But probably `current_assignable_group_id` can be replaced with `current_assignable_group.id`.

Alternatively commit still can be dropped because non-assignable projects seem to be filtered somewhere else and tag is not as offensive as a new file.

BrainPortal/lib/session_helpers.rb
MontrealSergiy: corrected

BrainPortal/app/helpers/select_box_helper.rb
prioux: Yes, it might be worth it to have only the method `current_assignable_group` and then do `current_assignable_group.id` when we need the ID.

You can make that change if you want, everywhere you invoke the _id method right now (just make sure not to forget any place)

BrainPortal/app/controllers/userfiles_controller.rb
MontrealSergiy: ok

BrainPortal/app/helpers/select_box_helper.rb
MontrealSergiy: done

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -1333,16 +1333,19 @@ def export_file_list
     # Create the new file list
     file_list = CbrainFileList.new(
       :user_id          => current_user.id,
-      :group_id         => current_project.try(:id) || current_user.own_group.id,
+      :group_id         => if current_project.try(:public) then current_user.own_group.id else current_assignable_group.id end,
```
prioux, line 1336: Please don't use a if then else, it's really ugly. if then else are more appropriate for statements, not for expressions, even though they work as expressions.

```ruby
:group_id         => (current_project.try(:public) ? current_user.own_group.id : current_assignable_group.id),
```

BrainPortal/spec/controllers/userfiles_controller_spec.rb
```diff
@@ -676,6 +678,26 @@ class << file; attr_reader :tempfile; end
 
 
 
+    describe "export_file_list" do
+      before(:each) do
+        session[:session_id] = 'session_id'
+        allow(controller).to      receive(:current_project).and_return(1010)
```
prioux, line 684: What is this `1010` that the method is retuning? That's not a group....

prioux (changes requested):
Minor mods.

BrainPortal/app/controllers/userfiles_controller.rb
MontrealSergiy: ok

BrainPortal/spec/controllers/userfiles_controller_spec.rb
MontrealSergiy: replaced with a group

PR #1090, opened 2021-02-11
DataProviders can have files under 'browse_paths'

prioux:
DataProvider subclasses can tell the CBRAIN framework
that they have a 'browse_path' capability. In that case
when a files is registered, a relative path 'a/b/c' is
added in the userfile's attribute :browse_path.

The DataProvider 'browse' page feature has been extended
to give users a minimal interface to navigate the
remote tree structure.

Programming notes:

From this commit on, all DataProvider classes defined
in plugins must change their definition for the method
impl_provider_list_all():

Before: impl_provider_list_all(as_user)
Now:    impl_provider_list_all(as_user, browse_path)


prioux:
To test this PR:

----
**1)** make sure you also have the code checked out for the DataladDataProvider class; this is currently in a PR in the repo aces/cbrain-plugins-neuro:

https://github.com/aces/cbrain-plugins-neuro/pull/174

(If you have cbrain-plugins-loris installed, update it too. The changes were pushed directly to that repo)

----
**2)** I found a datalad dataset out there that has has a tree structure suitable for testing many features. Configure a new data provider on your test CBRAIN and set:

a) the type as `DataladDataProvider`
b) the datalad repository URL as `https://datasets.datalad.org/naturalistic-data/Sherlock/preprocessed`
c) the datalad relative path as `fmriprep`

This means that within the 'preprocessed' dataset (which is a top-level datalad dataset), we fix the component 'fmriprep' and
we start browsing UNDER (below, inside) that level.

----
**3)** Check the the DP is alive, the go browse it. I suggest you keep alongside in a separate tab the following URL:

http://datasets.datalad.org/?dir=/naturalistic-data/Sherlock/preprocessed/fmriprep

This will allow you to examine the size of the directories and files BEFORE you register them

----
**4)** check with the interface and the console how the contents of the stuff you register is fetched properly. Reminders in the IRB console:

```
irb> Userfile.ctoday
irb> DataProvider.ctoday
irb> fff 1234 # find a userfile in @f or DP in @d
irb> tv @f # see all the attributes
irb> @f.sync_to_cache
```

The datalad dataprovider caches the entire datalad directory structure in a userfile in the scratch space of CBRAIN; you can find it with

```
irb> d=DataladDataProvider.last
irb> cf = d.send :userfile_for_datalad_cache
irb> puts cf.cache_full_path  # then go examine the content there; don't touch anything!
```

The DP browse page will absolutely allow you to register a file inside the directory structure of something else that is registered too; I think it works OK but let me know if you encounter any special problems.

prioux:
Another comment: while writing this I experimented with many other datalad datasets out there, but it turns out a great many of them contain corrupted data (e.g. files that are registered in datalad but the source has disappeared). It's quite frustrating. If you try with other data providers, keep this in mind when the sync_to_cache fails. It might not be a bug in CBRAIN.

Another note: some datalad datasets out there really require the very latest version of datalad and git-annex; everything fails otherwise. It's the case of for the CONP datasets.

prioux:
Note: in development mode, since classes are loaded dynamically, you might have to run these shell commands to make sure the DataladDataProvider class is loaded so you can create the first one:

```
cd BrainPortal/config/initializers
( echo 'puts_red "Loading Datalad class"' ; echo 'DataladDataProvider.nil?'  ) > zzz_dl_tmp.rb
```

Otherwise, the 'create data provider' page will not show the type.

MontrealSergiy:
I guess this requires datalad to be installed. (Browsing yields _Cannot get list of files. Maybe the remote directory doesn't exist or is locked?_  But exception is Exception: _CbrainError: Could not run datalad install_)

natacha-beck:
@MontrealSergiy yes you need to have Datalad on your system.

MontrealSergiy:
Yup with datalad installed seem work. Non-root user recommended process is via conda

natacha-beck:
Seems ok on my side, I reviewed the code and test it.

natacha-beck:
Since the DP is read-only, the registration panel should not allow moving the data (DataLad)

natacha-beck:
For DataLad DP the when registering a file if I ask CBRAIN to COPY the file on another DP, it is still register under the DataLad DP.

natacha-beck:
I have the same behaiour with the Multi DP.

prioux:
I'll check all that.

prioux:
OK well I'm moving forward with this PR now.

PR #1091, opened 2021-02-12
Tag

MontrealSergiy:
tag/rev info refactored with git tag describe

### DIFF
--- a/BrainPortal/lib/cbrain_file_revision.rb
+++ b/BrainPortal/lib/cbrain_file_revision.rb
@@ -418,8 +418,10 @@ class CbrainFileRevision
 
     #Dir.chdir(what_root.to_s) do
       out = `git describe --tags --long`
-      git_tag = out.rpartition('-')[0] if out.present?
-
+      if out.present?
+        git_tag, _, _ = out.rpartition('-')
+        git_tag.delete_suffix('-0')
+      end
   end
 
   protected
--- a/BrainPortal/lib/cbrain_file_revision.rb
+++ b/BrainPortal/lib/cbrain_file_revision.rb
@@ -407,34 +407,9 @@ class CbrainFileRevision
     rails_root = Pathname.new(Rails.root)
     what_root  = what == 'CBRAIN' ? rails_root.parent : Pathname.new(CBRAIN::Plugins_Dir) + what
 
-    # Unfortunately, in multithreaded versions of this app (puma server maybe?)
-    # we get into problems with chdirs being active in multiple threads.
-    # So the Dir.chdir() block below is commented out and all bash statements
-    # are prefixed with a "cd" command. It's ugly but eh.
     cd_bash    = "cd #{what_root.to_s.bash_escape} ; " # Prefix for all bash commands below
 
-    #Dir.chdir(what_root.to_s) do
-      return DEFAULT_TAG if `#{cd_bash} git rev-parse --show-toplevel 2>/dev/null`.strip != what_root.to_s
-      tags_set = `#{cd_bash} git tag -l`.split.shuffle # initial list: all tags we can find
-      git_tag = tags_set.shift unless tags_set.empty? # extract one as a starting point
-      seen[git_tag] = true
-      while tags_set.size > 0
-        tags_set = `#{cd_bash} git tag --contains #{git_tag.bash_escape}`.split.shuffle.reject { |v| seen[v] }
-        git_tag = tags_set.shift unless tags_set.empty? # new first
-        seen[git_tag] = true
-      end
-      if git_tag
-        num_new_commits = `#{cd_bash} git rev-list '#{git_tag}..HEAD'`.split.size
-      elsif DEFAULT_TAG # fallback for plugins
-        num_new_commits = `#{cd_bash} git rev-list HEAD`.split.size
-        git_tag = DEFAULT_TAG
-      end
-      if git_tag
-        git_tag += "-#{num_new_commits}" if num_new_commits > 0
-      end
-    #end
-
-    git_tag # not cached, so it's really live
+    git_tag = `#{cd_bash} git describe --tags --long | sed 's/\\(.*\\)-.*/\\1/'`# not cached, so it's really live
 
   end
 
--- a/BrainPortal/lib/cbrain_file_revision.rb
+++ b/BrainPortal/lib/cbrain_file_revision.rb
@@ -408,8 +408,12 @@ class CbrainFileRevision
     what_root  = what == 'CBRAIN' ? rails_root.parent : Pathname.new(CBRAIN::Plugins_Dir) + what
 
     cd_bash    = "cd #{what_root.to_s.bash_escape} ; " # Prefix for all bash commands below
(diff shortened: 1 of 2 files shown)


MontrealSergiy:
found bug

prioux:
I was about to comment.

I will still do it in the code section here even if it's closed.

BrainPortal/lib/cbrain_file_revision.rb
```diff
@@ -409,32 +409,16 @@ def self.cbrain_head_tag(what = 'CBRAIN') #:nodoc:
...
-        git_tag += "-#{num_new_commits}" if num_new_commits > 0
-      end
-    #end
+    cd_bash = "cd #{what_root.to_s.bash_escape} ; " # Prefix for all bash commands below
+
+    return DEFAULT_TAG if `#{cd_bash} git rev-parse --show-toplevel 2>/dev/null`.strip != what_root.to_s
 
-    git_tag # not cached, so it's really live
+    #Dir.chdir(what_root.to_s) do
+      out = `git describe --tags --long`
```
prioux, line 420: 1) Indentation.

2) you forgot to do the `cd` command so this code no longer extracts the revision numbers for the plugins.

prioux: 2) many of the variables defined above no longer are useful (seen, what_root etc), so clean up

prioux: 4) Don't leave the commented-out Dir.chdir statement, it's useless now.

BrainPortal/script/show_cbrain_rev
```diff
@@ -34,25 +34,10 @@ end
...
-        num_new_commits = `git rev-list '#{git_tag}..HEAD'`.split.size
-      elsif default_tag # fallback for plugins, when -z used
-        num_new_commits = `git rev-list HEAD`.split.size
-        git_tag = default_tag
-      end
-      if git_tag
-        git_tag += "-#{num_new_commits}" if num_new_commits > 0
+      out = `git describe --tags --long`
+      if out
+        git_tag, _, _ = out.rpartition('-')
```
prioux, line 40: why do you use this strange statement when in the other code you used `out.rpartition('-')[0]` ?

prioux: to be clean, redirect stderr in the git command

MontrealSergiy: ok for '-'. that is a common pattern with python, so I used with it.

MontrealSergiy:
@prioux Are we supporting ruby 2.4 ? (in 2.5 offers one convenient sting method)

BrainPortal/script/show_cbrain_rev
MontrealSergiy: sorry not follow sderr to std out ? or null?

BrainPortal/lib/cbrain_file_revision.rb
MontrealSergiy: `what_root` is used for cd (which you asked to put back)

BrainPortal/script/show_cbrain_rev
prioux: To /dev/null . Redirect the command's stderr to /dev/null so we don't pollute the output with spurious error messages. **Just like everywhere else in this file!**

MontrealSergiy: adding command to that file (not sure for another)

BrainPortal/lib/cbrain_file_revision.rb
```diff
@@ -403,39 +403,18 @@ def self.cbrain_head_tag(what = 'CBRAIN') #:nodoc:
...
-      if git_tag
-        git_tag += "-#{num_new_commits}" if num_new_commits > 0
-      end
-    #end
-
-    git_tag # not cached, so it's really live
+    return DEFAULT_TAG if `#{cd_bash} git rev-parse --show-toplevel 2>/dev/null`.strip != what_root.to_s
 
+    out = `#{cd_bash} git describe --tags --long 2>/dev/null`
+    if out.present?
```
prioux, line 414: I won't ask you to change it but just so you know, you could have done all four lines of your `if` block with a single line instead

```
out.gsub!(/-0$/,"")
```

You don't even have to verify it ends with -0 or not: if the substitute fails, it leaves the string alone.

BrainPortal/script/show_cbrain_rev
```diff
@@ -34,26 +34,11 @@ end
...
-      if git_tag
-        num_new_commits = `git rev-list '#{git_tag}..HEAD'`.split.size
-      elsif default_tag # fallback for plugins, when -z used
-        num_new_commits = `git rev-list HEAD`.split.size
-        git_tag = default_tag
-      end
-      if git_tag
-        git_tag += "-#{num_new_commits}" if num_new_commits > 0
-        puts git_tag
+      out = `git describe --tags --long`
```
prioux, line 38: Same here.. Al of this could be

```
  `git describe --tags --long`
  .strip
  .sub(/(-0)?-\w+$/,"")
```

MontrealSergiy: thanks, but can replace if you prefer. I know that regex is powerful and concise, yet I believe tat re is a bit condense for most people, me included, so we have more trouble with deciphering regexes than with string operations involving partition/delete_suffix/split and alike, so tried ditch re.

prioux: That's a terrible argument. This regex is very simple. If you think it's still too complex then you write it with the extended syntax that allows you to embed comments:

```
  `git describe --tags --long`
  .strip
  .sub(
    /
      (-0)?           # optional dash zero, like "-0"
      -\w+            # a dash and alphanums like "-b027af3"
     $                  # anchor at end of string
    /x ,
    ""                  # replace with nothing
 )
```

To me that is overkill, and the straight version ` .sub( /(-0)?-\w+$/,  "" ) ` is totally readable.

MontrealSergiy: well, IMHO. stackoverflow is filled with regex hacks question. I shortened the old version a bit, hope better now

prioux:
Your code still doesn't work properly. I said I wanted no change in behavior.

I'm running your script in a plugin with no tags I a get a STDERR message.

Here's what my original code did:

```
> cd /BrainPortal/cbrain_plugins/cbrain-plugins-bids-fmriprep # any plugins with no tags
> ../../script/show_cbrain_rev -z
0.1.0-14
```

Here's what your code does:

```
> ../../script/show_cbrain_rev -z
fatal: No names found, cannot describe anything.

```

(one error line is on stderr, and one blank line on stdout)

prioux (changes requested):
Still buggy. See my last comment.

Also remember to redirect stderr in the mail CBRAIN code too.

MontrealSergiy:
sorry, I think you meant to redirect stderr in standalone secondary script?

MontrealSergiy:
done

PR #1095, opened 2021-02-15
New hide_filtered() method

prioux:
In old Rails there used to be a method called 'hide_filtered" for hash objects; we had a temporary replacement but this PR replaces it properly with a call to a class in ActionDispatch.

You can test this on the command-line. To check the exception reports, try adding `raise 'oh no'` in some of your controllers, especially those handling parameters with passwords.


natacha-beck:
Tested, it works fine.

natacha-beck (approved):

PR #1099, opened 2021-03-01
refactor BrowseProviderFileCaching fixes #1094

### LINKED ISSUE #1094: Replace the class BrowseProviderFileCaching with Rails.cache
The class BrowseProviderFileCaching just fetches and stores browsing data in a file in /tmp ; replace the caching part with calls to Rails.cache instead.

Make sure the cache key is tied to all three of : 1) which user is browsing, 2) which data provider is being browsed, and 3) under which path is the browsing happening.

MontrealSergiy:
Refactored provider browsing cache to use default Rails cache. 

People who use in memory caching (memcache etc) might need re-adjust setting if using large providers (e.g. 2 M for 10000 files). Currently non-issue for us as our portal uses disk for caching.


prioux:
It all looks good but you removed the description of the method `self.cache_file`

PR #1102, opened 2021-03-29
New ORCID integration.

prioux:
Users no longer enter their ORCID iD directly.


prioux:
@natacha-beck I added you as a reviewer but basically you already reviewed this, so no need to do it again. This PR is just so that this code now get officially integrated into master.

natacha-beck:
Works for me.

PR #1104, opened 2021-03-30
Added new SingBindmountDataProvider class

prioux:
This is a new DP class. To test this you need the following:

1. access to a computer, let's call it 'h' which can be local or remote
2. that computer must have Singularity 3.7 or better
3. in some new directory of your choice install:
  a) a file "sing_squashfs.simg" : a singualrity container file that contains at least "rsync"
  b) a ext3 or squashfs filesystem-in-a-file; this can be named whatever you want
  Note: I can provide you with examples for these two files.
4. configure the new DP in your CBRAIN, it will appear under the category 'Singularity Bind Mount'
  a) enter the remote host name 'h'
  b) enter a remote user, port, and make sure the SSH key of your CBRAIN is installed (just like a normal SSH DP)
  c) for the remote_dir attribute (shown as 'physical data localtion' in the form, enter the full path of the ext3 or squashfs file (yes, it's not a directory)
  d) for the 'Containerized Data Path' attribute, enter any path for something inside your filesystem-in-a-file, but prefixed with "/cbrain/" (even if there is no such directory in the filesystem). E.g. if you want to access stuff that is inside "/data/collection", you enter "/cbrain/data/collection" instead. If you want to access the root, enter just "/cbrain".

### DIFF
--- a/BrainPortal/app/models/sing_bindmount_data_provider.rb
+++ b/BrainPortal/app/models/sing_bindmount_data_provider.rb
@@ -1,0 +1,412 @@
+
+#
+# CBRAIN Project
+#
+# Copyright (C) 2008-2019
+# The Royal Institution for the Advancement of Learning
+# McGill University
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# This class implements a Data Provider which fetches
+# files out of a single filesystem file. The
+# implementation requires Singularity 3.2 or better to
+# be installed on the host, as well as a singularity
+# container image that contains the basic Linux
+# commands and the 'rsync' command too.
+#
+# For the moment the DP is read-only. Allowing
+# write access would require a mechanism in CBRAIN
+# to prevent more than one singularity process to
+# be accessing the mounted file at any one time.
+#
+# Bind mounts are performed using a command such as:
+#
+#   singularity shell -B filesystem.img:/my/mount/point:image-src=/inside/dir,ro sing_squashfs.simg
+#
+# where:
+# 1- sing_squashfs.simg is a singularity image file,
+# 2- filesystem.img is either a ext3 or squashfs filesystem,
+# 3- /inside/dir is a path inside that filesystem,
+# 4- /my/mount/point is a path where the data will be mounted under
+# and which doesn't need to exist in sing_squashfs.simg
+#
+# This class hardcode the values for /my/mount/point to be "/cbrain"
+# and /inside/dir to be "/". The value of the data provider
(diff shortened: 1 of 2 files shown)


prioux:
I have added two files in the CBRAIN service for helping you test the DP. You can find them in the project 'CBRAIN-Developers' (you are a member of that project).

* sing_squashfs.simg
* top-data-rw.ext3

This last file contains this simple data structure, listed with "." being "/" :

```
     2    4 drwxr-xr-x   6 root     root         4096 Mar 30 10:54 .
    19    8 -rwxrwxr-x   1 1000     1000         5404 Mar 29 17:12 ./withoverlay
    12    4 drwxrwxrwt   2 1000     1000         4096 Mar 30 10:56 ./rwdir
    20    0 -rw-r--r--   1 prioux   prioux          0 Mar 30 10:56 ./rwdir/abcd
    17    4 drwxrwxr-x   2 1000     1000         4096 Mar 29 17:12 ./top2
    18    4 -rw-rw-r--   1 1000     1000            6 Mar 29 17:12 ./top2/file.txt
    11   16 drwx------   2 root     root        16384 Mar 30 10:54 ./lost+found
    13    4 drwxrwxr-x   3 1000     1000         4096 Mar 29 17:12 ./top1
    15    4 drwxrwxr-x   2 1000     1000         4096 Mar 29 17:12 ./top1/subdir1
    16   36 -rwxr-xr-x   1 1000     1000        35064 Mar 29 17:12 ./top1/subdir1/cat
    14  136 -rwxr-xr-x   1 1000     1000       133792 Mar 29 17:12 ./top1/ls
```

(uids and gids will be different depending on where you open it up).

You can inspect the ext3 with this command:

```
singularity shell -B top-data-rw.ext3:/cbrain:image-src=/,ro sing_squashfs.simg
```

will will mount the / of the ext3 under /cbrain, like the DP does.

MontrealSergiy:
In the view mode I do not see containerized path, only full path to singularity file

MontrealSergiy:
[cbrain_SingbindDP.pdf](https://github.com/aces/cbrain/files/6273606/cbrain_SingbindDP.pdf)

prioux:
Please don't post PDFs here, just add screenshots

prioux:
You are right, it's a bug in the view code. In the meantime if you need to change that value, use the console.

MontrealSergiy:
Screenshot is below image

prioux:
Added a commit with the fix

MontrealSergiy:
now container "path" is shown

MontrealSergiy:
I was not able to browse the data provider - no files are shown. I though used a VM (vagrant box) rather than real remote pc.

prioux:
Well, what do you get when you type

```ruby
fis=@d.provider_list_all;1
```

in the console? fis should be an array of FileInfo objects. If you get an exception, it might tell you something about an error in the configuration.

MontrealSergiy:
I tried today a slightly different vm's setup (with a 'private' network, on my home pc) and the feature worked well. Maybe I will try latter again it tell if I see any problems. I understand it is readonly.

For old setup I got no error message, but when tried a ssh dp to same machines , it generated an exception, so, most likely the problem with key or other setup details, not the feature under test. Ideally,  some detailed warnings about connection or auth issues are desirable, but I guess an admin will figure anyway if something not working. 
image

prioux:
OK, good, thank you @MontrealSergiy . If you could just approve the PR now ?

MontrealSergiy:
the feature works, approved

MontrealSergiy (approved):
feature works

MontrealSergiy:
Some explanations how to configure could be added to Data Provider Type Information

PR #1110, opened 2021-04-26
keep select 'chosenified' after ajax fixes #1098

### LINKED ISSUE #1098: The selection boxes in the DP browse page are no longer chosen-ified on page 2+
When switching pages in the DP browse page, the selection boxes no longer are chosen-ified.

Followup to #1087 and #1089

MontrealSergiy:


prioux:
Seems to work.

PR #1111, opened 2021-04-26
Cbcsv could have different extensions fixes #1063

### LINKED ISSUE #1063: The Boutiques integrator requires CbrainFileList files to have a .cbcsv extension
See subject. If a CbrainFileList is not named with a `.cbcsv` extension, it will be treated as a normal file for the purposes of launching an array of task (so, only one file, the cbrain file list itself, is assigned to the input).

This is a bit strict, maybe we can relax this.

MontrealSergiy:


BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
```diff
@@ -530,7 +530,7 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
     return [] if files.nil? || files.length == 0
     files.select { |f| ! self.params[f].nil? && ! file_lists.include?(f) } # Prevent problems with file-type inputs with list=true
          .map    { |f| [f, Userfile.find_accessible_by_user(self.params[f], self.user, :access_requested => file_access)] }
-         .select { |f| f[1].name.end_with?('.cbcsv') || f[1].suggested_file_type == CbrainFileList }
+         .select { |f| f[1].name.end_with?('.cbcsv') || f[1].is_a?(CbrainFileList) }
```
natacha-beck, line 533: Why removing the `f[1].suggested_file_type == CbrainFileList` since it is a || you just have to add an extra condition.

MontrealSergiy: It is just a bonus refactoring. The f[1].suggested_file_type == CbrainFileList is same as f[1].name.end_with?('.cbcsv'), why keep both here?

MontrealSergiy: I'd prever to drop that, but if say ,there are plans to change or add extensions to cbrainfilelist, we may keep `suggested_file_type == CbrainFileList`,

prioux: Sergiy is right. The only times that

```ruby
f[1].suggested_file_type == CbrainFileList
```

will be true is when the file's name ends with '.cbcsv', so the two conditions that were there were the same.

We want this PR to work with these three scenarios:

1) a file that has a .cbcsv extension and is a CbrainFileList
2) a file that has some other extension (.csv, .tab, .txt) but still has type CbrainFileList
3) a file that has a .cbcsv extension yet has a different type (e.g. CsvFile)

The original code only worked for case 1 and 3

Sergiy's code here works for case 1, 2 and 3, which means it's correct.

Does that make sense Natacha?

prioux (approved):
Is it the only part of the code where we need to change this? There is an awful lot of places with CbrainFileList checks in this file.

MontrealSergiy:
Not sure. I do not see which instance of type check can be safely deleted. The template seems a bit convoluted and repetitive, but that's what I would of expect from code-generating code.

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
natacha-beck: I agree with the equality of both conditions right now. 

But if we look  at the `suggested_file_type` definition (https://github.com/aces/cbrain/blob/master/BrainPortal/app/models/userfile.rb#L258), it call the `file_name_pattern`. So if at some point we modify the `file_name_patern` method in CbrainFileList the code will start to not detect some files, that could be a CbrainFileList. 

So we should remove the `f[1].name.end_with?('.cbcsv')` no ? @prioux.

prioux: No Natacha. If we remove that part we can no longer detect case #3 in my list

MontrealSergiy: @natacha-beck , imho keeping in mind that just keeping `f[1].suggested_file_type == CbrainFileList` might suffuce for .cbcvs extention detection. But to make possible the hypotetical  future changes to cbcvs extension, yet there are few others .cbcvs extension checks within the code.

natacha-beck: True. So I came back to my first comment: 

> Why removing the f[1].suggested_file_type == CbrainFileList since it is a || you just have to add an extra condition.

we then can change the `file_name_patern` without impacting the code.

prioux: After more discussion with Natacha, the new alternative I suggest is now:

```ruby
  .select { |f| (f[1].suggested_file_type || Object) <= CbrainFileList }
```

This means we check for cases 1, 2 and 3 properly as long a the CbrainFileList model detects the '.cbcsv' extension (which it does right now) and will also support future extensions too, if we extend the model, or if we subclass it.

Do you both agree it makes sense?

prioux: I used `Object` as a default when `suggested_file_type` returns nil so that we fail the comparison `<=` with CbrainFileList.

MontrealSergiy: I did not get why case 2 is passing.

prioux: Oh, you're right, case #2 doesn't work.

OK new solution:

```ruby
.select { |f| f[1].is_a?(CbrainFileList) || (f[1].suggested_file_type || Object) <= CbrainFileList }
```

MontrealSergiy: Anyway, it seem to work. Well, I do not mind then. Should I replace other hard-coded extension changed in that template too?

MontrealSergiy: it's ```cbcsvs = files.map { |f| Userfile.find_accessible_by_user( f, self.user, :access_requested => file_access ) }.select { |f| f.name.end_with?('.cbcsv') || f.is_a?(CbrainFileList) }```

Do you think we can move `cbcsv = self.cbcsv_files` before the conditions?

MontrealSergiy: Still have a questions `SingleFile <= CbrainFileList` evaluates to `true`? Maybe you meant reverse of?

natacha-beck: If you try in a console `SingleFile <= CbrainFileList` evaluate to false. 

```ruby
NaN main :009 > p SingleFile <= CbrainFileList 
false
 => false 
```
 So @prioux  solution works, and cover future extension.

MontrealSergiy: ok. BTW it that can be simplified a bit

```.map    { |f| Userfile.find_accessible_by_user(self.params[f], self.user, :access_requested => file_access)] }
         .select { |f| f.is_a?(CbrainFileList) || (f.suggested_file_type || Object) <= CbrainFileList }
```

MontrealSergiy: Also, FYI a form validation rule blocks files with right extension but different form ('case 2').  image

MontrealSergiy: but after change, if we associate more extensions to File List, this message should appear for that new messages too

PR #1118, opened 2021-05-17
avoid exception nil.inputs_readonly fix

MontrealSergiy:
I noted exceptions in production when method is checked on nil. This should fix it.


natacha-beck (approved):
Seems ok to me.

PR #1120, opened 2021-05-20
Added list of all tools and datasets.

prioux:
Note: the sysadmin must configure an AccessProfile
named 'Restricted Datasets' to show dataset groups.
No user need to be in that profile.


prioux:
@natacha-beck @MontrealSergiy You dont; really have to test this; I'd like you to just read the code.

I will merge this very very soon so I can test it in the dev server.

Issue #1076, opened 2021-01-12
Extract folder from file collection


natacha-beck:
Actually, we can extract a file from a file collection but we cannot extract a subfolder from a file collection. I am wondering to know why we have this situation. 
It can be pretty useful to extract a subfolder.

MontrealSergiy:
Is it closed?

Issue #229, opened 2015-09-28
Selecting CBRAIN output files (Redmine 7925)


prioux:
(From Najma, summer 2015)

This is related to the objective of making MINC and other softwares interoperable.

Scenario: CIVET performs superior to FSL tools in brain extraction function. I like to combine FSL's functional suite but take advantage of CIVET outputs at the same time. For example, FSL fails on brain-masking miserably and I would like to use one set of files in the CIVET output "mask" directory to fix that problem.

Desired solution: I like to have an option in CBRAIN that allows me to use files that are in the CBRAIN output directory structure as inputs to other tasks.
Or simpler, I like to take advantage of a CBRAIN filter that allows me to create a "collection" from a particular file in a CIVET sub_directory.

Any thoughts on this?

prioux:
We have several experimental code frameworks that approach such problems, but none of them are ready (or are just at the planning stage). Yes, it would be a useful feature.

1- generic data reformatter: select a bunch of files or file collections, enter a rewrite rule, and create new files repackaged differently, suitable (Issue #40)

2- register files that are part of other files (Issue #168)

3- proxy objects (experimental code by Rémi, still in private branch, too unstable)

natacha-beck:
Similar to what Carolina want: 

> 
> CIVET customization
> a. Running mincbet to alter parameters of the brain mask required for brain extraction.
> Below is the mincbet command to implement to create a new brain mask, with recommended values:
> 
> mincbet <final/Prefix_ID_t1_final.mnc> mask -n -m -h 1.15 -f 0.50
>  
> The above command should be applied to the final t1 image (in the "final" CIVET output directory).
> The values for flags "-h" and "-f" can be altered by the user, to something like -h 1.10, and/or -f 0.6.
> We will have to keep in mind a way to keep the "old" mask in /mask/Prefix_ID_brain_mask.mnc, before replacing it with this newly created mask.

crocodoyle:
Is this issue still relevant? There is now a way to extract files from FileCollections, and a recent PR of mine will make it easier to download files from inside a FileCollection

PR #1131, opened 2021-06-11
Added SimpleFileExtractor tool

prioux:
This PR adds a new standard CbrainTask, the `SimpleFileExtractor`.

You select a bunch of FileCollections, and provide patterns such as `*/*.txt` or `*/thickness/*30mm*.dat` etc, and it will create a new FileCollection with all these extracted files in a single level.

The program verifies that the input FileCollections are on a local DP.

It will also provide a bunch of statistics and warning messages when collisions occur, or patterns don't match anything, etc.

I took great care of validating as much as possible the globbing code so that no files are extracted outside of their proper collections.

### DIFF
--- a/BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/simple_file_extractor/bourreau/simple_file_extractor.rb
+++ b/BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/simple_file_extractor/bourreau/simple_file_extractor.rb
@@ -1,0 +1,185 @@
+
+#
+# CBRAIN Project
+#
+# Copyright (C) 2008-2021
+# The Royal Institution for the Advancement of Learning
+# McGill University
+#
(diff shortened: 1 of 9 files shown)


prioux:
Possible enhancements:

1. allow the admin to extract files that are no local
2. allow a normal user to extract files that are not local if the total size is < some threshold?
3. add flags that tell the task to fail immediately if any subset of the warning conditions happen (e.g. stop if there is a collision, or stop if a pattern doesn't match anything
4. create non-flat outputs, where when a file is extracted, a subdirectory creation pattern can be provided for it
5. extract subfolders (see #1076)

See also #229

MontrealSergiy:
1. Not sure is it really extractor, it is more of filter: a true extractor will extract files themselves.
2. Not sure how the Prime Minister is involved in that.
3. Should I test it?

prioux:
What are you saying Sergiy? It *IS* a true extractor, it creates a new FileCollection with the extracted files!

Issue #1136, opened 2021-06-22
Boutiques all-or-none support


MontrealSergiy:
CBRAIN already supports 'one-is-required' and mutex boutiques group flag, but 'all-or-none', which seems to be useful feature, is allowed, yet not fully supported. It seems that adding that one not so difficult, so we have all the group flags supported.

One use case for all-or-none is to define a 3d vector or coordinate. While one can try use list feature and setup, representing a vector as 3 params with all-or-none constraint is more straightforward and user has to do less clicking, resulting in more comfortable UX. For example see https://github.com/aces/cbrain-plugins-neuro/blob/master/cbrain_task_descriptors/deform_sim.json

MontrealSergiy:
resolved

PR #1137, opened 2021-06-22
let's add 'all-or-none' boutiques group parameter validation? #1136

MontrealSergiy:
How about adding 'all-or-one' boutiques group parameter validation?

(see the linked issue #1136 for expanded motivation )

### DIFF
--- a/BrainPortal/app/models/boutiques_portal_task.rb
+++ b/BrainPortal/app/models/boutiques_portal_task.rb
@@ -203,6 +203,7 @@ class BoutiquesPortalTask < PortalTask
     descriptor.groups.each do |group|
       check_mutex_group(group)         if group.mutually_exclusive
       check_oneisrequired_group(group) if group.one_is_required
+      check_allornone_group(group) if group.all_or_none
     end
 
     # ------------------------------------------------
@@ -589,6 +590,13 @@ class BoutiquesPortalTask < PortalTask
     are_set = members.select { |inputid| ! isInactive(descriptor.input_by_id inputid) }
     return if are_set.size > 0
     params_errors.add(group.name, " need at least one parameter set")
+  end
+
+  def check_allornone_group(group, descriptor = self.descriptor_for_after_form)
+    members = group.members
+    are_set = members.select { |inputid| ! isInactive(descriptor.input_by_id inputid) }
+    return if [0, members.size].include? are_set.size
+    params_errors.add(group.name, " need either all the parameter set or neither")
   end
 
   # MAYBE IN COMMON
--- a/BrainPortal/app/models/boutiques_portal_task.rb
+++ b/BrainPortal/app/models/boutiques_portal_task.rb
@@ -594,8 +594,8 @@ class BoutiquesPortalTask < PortalTask
 
   def check_allornone_group(group, descriptor = self.descriptor_for_after_form)
     members = group.members
-    are_set = members.select { |inputid| ! isInactive(descriptor.input_by_id inputid) }
-    return if [0, members.size].include? are_set.size
+    num_unset = members.count { |inputid| isInactive(descriptor.input_by_id inputid) }
+    return if num_unset == 0 || num_unset == members.size # all, or none
     params_errors.add(group.name, " need either all the parameter set or neither")
   end
 
--- a/BrainPortal/app/models/boutiques_portal_task.rb
+++ b/BrainPortal/app/models/boutiques_portal_task.rb
@@ -203,7 +203,7 @@ class BoutiquesPortalTask < PortalTask
     descriptor.groups.each do |group|
       check_mutex_group(group)         if group.mutually_exclusive
       check_oneisrequired_group(group) if group.one_is_required
-      check_allornone_group(group) if group.all_or_none
+      check_allornone_group(group)     if group.all_or_none
     end
 
     # ------------------------------------------------
--- a/BrainPortal/app/models/boutiques_portal_task.rb
+++ b/BrainPortal/app/models/boutiques_portal_task.rb
@@ -596,7 +596,7 @@ class BoutiquesPortalTask < PortalTask
     members = group.members
     num_unset = members.count { |inputid| isInactive(descriptor.input_by_id inputid) }
     return if num_unset == 0 || num_unset == members.size # all, or none
-    params_errors.add(group.name, " need either all the parameter set or neither")
+    params_errors.add(group.name, " need either all the parameters set or neither")
   end
 
   # MAYBE IN COMMON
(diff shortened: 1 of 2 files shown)


prioux:
I haven't had time to review this Sergiy but this seems interesting. It's likely this PR will suddenly have conflicts once I merge #1145 in a few minutes, but we can fix that later. In the meatime, can you post here two snippets of boutiques code that test the before and after of what this PR does? Maybe using a modified version of the BoutiquesDemo descriptor?

MontrealSergiy:
it is already in deformation tool boutiques (presently it does nothing but after change should do validation)

natacha-beck (approved):

natacha-beck:
Fine for me I tested this PR with the DeformSim tool.

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
```diff
@@ -412,7 +415,15 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
       params_errors.add(grpName[group], errMsg) unless hasOneActive.(group)
     end
 % end
-%#
+% if hasAllNone
+    # Lambda for checking all-or-none (either none or all are active)
+    hasAllNone = lambda { |gid| gidToMbrs[gid].select{ |m| ! isInactive.(m.to_sym) }.count == 0 ||
+                           gidToMbrs[gid].select{ |m| isInactive.(m.to_sym) }.count == 0}
```
prioux, line 421: I'm not going to request a change but this complex test could be made way simpler.

For instance:

Using `.all?` instead of `.select { code }.count == 0`

```ruby
asAllNone = lambda { |gid| gidToMbrs[gid].all? { |m|  ! isInactive.(m.to_sym) } ||
                                            gidToMbrs[gid].all?{ |m|  isInactive.(m.to_sym) }
```

Or you could just make sure all the `isInactive()` call return the same thing:

```
hasAllNone = lambda { |gid| gidToMbrs[gid].map { |m| isInactive.(m.to_sym) }.uniq.count == 1 }
```

prioux (changes requested):
I would like this PR to be extended to implement the same logic in the new integrator.

The code changes should be easy to make, because the code is already pretty similar.

The file to be modified is `app/models/boutiques_portal_task.rb` ; find the method `check_oneisrequired_group` as an example of how it was restructured.

BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb
MontrealSergiy: These are cool, especially the second one. Perhaps too cool, as requires some thinking to understand.

MontrealSergiy:
image
seems work. Note I had to rebase

MontrealSergiy:
I would probably add 'Coordinates parameter group' to avoid potential confusion with similarly named parameters, but best keep uniform verbiage

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -203,6 +203,7 @@ def after_form
     descriptor.groups.each do |group|
       check_mutex_group(group)         if group.mutually_exclusive
       check_oneisrequired_group(group) if group.one_is_required
+      check_allornone_group(group) if group.all_or_none
```
prioux, line 206: Align the `if` please

MontrealSergiy:
> I would like this PR to be extended to implement the same logic in the new integrator.
> 
> The code changes should be easy to make, because the code is already pretty similar.
> 
> The file to be modified is `app/models/boutiques_portal_task.rb` ; find the method `check_oneisrequired_group` as an example of how it was restructured.

done

BrainPortal/app/models/boutiques_portal_task.rb
MontrealSergiy: done

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -591,6 +592,13 @@ def check_oneisrequired_group(group, descriptor = self.descriptor_for_after_form
     params_errors.add(group.name, " need at least one parameter set")
   end
 
+  def check_allornone_group(group, descriptor = self.descriptor_for_after_form)
+    members = group.members
+    are_set = members.select { |inputid| ! isInactive(descriptor.input_by_id inputid) }
```
prioux, line 597: Personally I'd have written:

```ruby
num_unset = members.count { |inputid| isInactive(descriptor.input_by_id inputid) }
return if num_unset == 0 || num_unset == members.size # all, or none
```

That's because arrays have a .count() method that can take a true-false block

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -203,6 +203,7 @@ def after_form
     descriptor.groups.each do |group|
       check_mutex_group(group)         if group.mutually_exclusive
       check_oneisrequired_group(group) if group.one_is_required
+      check_allornone_group(group)     if group.all_or_none
```
prioux, line 206: Align the `if` please

BrainPortal/app/models/boutiques_portal_task.rb
MontrealSergiy: ok, I thought of something like that, even though opted to keep checkers uniform

prioux:
I'll merge this tomorrow morning, it's ready

BrainPortal/app/models/boutiques_portal_task.rb
MontrealSergiy: done

MontrealSergiy:
Well I am still not sure of ```Coordinates need either all the parameter set or neither``` vs ```Coordinates group need either all the parameters set or neither```. Can group name be singular? Like Coordinate? then verb-noun agreement is broken in  ```Coordinate need either all the parameter set or neither``

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -591,6 +592,13 @@ def check_oneisrequired_group(group, descriptor = self.descriptor_for_after_form
     params_errors.add(group.name, " need at least one parameter set")
   end
 
+  def check_allornone_group(group, descriptor = self.descriptor_for_after_form)
+    members = group.members
+    num_unset = members.count { |inputid| isInactive(descriptor.input_by_id inputid) }
+    return if num_unset == 0 || num_unset == members.size # all, or none
+    params_errors.add(group.name, " need either all the parameters set or neither")
```
prioux, line 599: I will fix the message after I merge this

PR #1138, opened 2021-06-24
Added support for Login With Globus

prioux:
This is difficult to test. First you need a developer endpoint at Globus.
Then you need to enter the endpoint's ID and secret using the Rails Console:

```ruby
r=BrainPortal.first # or whatever
r.meta[:globus_client_id] = "e25539dd-5348-4711-ae4b-ec2735be709c" # fake values here
r.meta[:globus_client_secret] = "Pv2Hh6l7/OI/xns7Y8vsTBVPpl37dCrY/MoVJq+XvTs="
```

At least maybe read the code very carefully....?

### DIFF
--- a/BrainPortal/Gemfile
+++ b/BrainPortal/Gemfile
@@ -45,6 +45,7 @@ gem "pbkdf2-ruby"
 gem "json-schema"
 gem "aws-sdk-s3", '~> 1'
 gem 'zenodo_client', '>= 1.0.0', :git => 'https://github.com/aces/zenodo-client-gem'
+gem 'jwt'
 
 # User interface gems
 gem "will_paginate"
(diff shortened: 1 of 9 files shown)


MontrealSergiy:
Besides id and secret your globus check asks for site_url_prefix. Otherwise you code consider globus login non-configured. What is it and how to setup that?  

def globus_auth_configured?
    myself   = RemoteResource.current_resource
    site_uri = myself.site_url_prefix.presence

prioux:
It's the scheme+host part of the URL for your sitel in dev mode it's typically set to:

```
http://localhost:3000
```

MontrealSergiy:
Thanks, with that I was able to login locally

MontrealSergiy (approved):
I was able to login with my primary (@mcgill.ca) and associated (@mcin.ca) globus identities.

1. There were no linked issues, so I am surprised it goes to CBRAIN rather than Neurohub,
the flow. 
2. Considering how emails are handled in CBRAIN I think we should better follow same process as with Orcid, also it may be better have the uniform even if complicated process for all the oauth2 logins
3. It is strange that admin is supposed use console to set secrets, should go via config file/env variables, in the way other secrets are set
4. Would be nice to put a warning if Globus auth server is down (or maybe not if it is a stable enough)

PR #1140, opened 2021-06-28
Introduce Docker

dlq:


dlq:
docker-compose build
docker-compose up
docker-compose run brainportal rake db:create

MontrealSergiy:
Same failure with Ubuntu 16 laptop

dlq:
Hmm.  Do you have any outputs that I can see?


Darcy

MontrealSergiy:
> Hmm. Do you have any outputs that I can see? Darcy

Do you see the stack trace in the above comment https://github.com/aces/cbrain/pull/1140#issuecomment-869994201

dlq:
Yes, now I do.  Thanks!

prioux:
This doesn't work. There are no places where the configuration for the servers are made:

1. the config_portal.rb (or config_bourreau.rb) is not properly poluated with a name
2. the database.yml is never configured

I also don't understand what the PR is supposed to be about anyway. Provide a developement environement for a developer? A production CBRAIN portal? What?

If the first, how is the developer supposed to be able to edit CBRAIN's code base and see what effect they have on the server?

prioux:
No activity and no fixes, closing.

PR #1145, opened 2021-07-27
Boutiques: File inputs with list=true support CbrainFileList

prioux:
When a Boutiques descriptor contains an input of type `File` with `"list=true"` like this:

```json
        {
            "name":              "Required multi input 1",
            "id":                "minput1",
            "description":       "Any file or directory in CBRAIN, or even a CbrainFileList",
            "type":              "File",
            "optional":          false,
            "list":              true,
            "value-key":         "[MINPUT1]"
        }
```

then the user can provide any normal CBRAIN file or a  CbrainFileList; in the case of a CbrainFileList, the files specified in it will be synchronized and substituted on the command-line for the `value-key`.

This PR contains a few other Boutiques-related bug fixes and improvements.

This PR also replaces completely the code in PR #1141


MontrealSergiy:
fails with error 

[2021-08-02 13:39:52 UTC] make_cluster_workdir() Trying to create workdir '/data1/user-data/sboroday/dp/BorWork/00/00/10/admin-MultiBoutiquesDemo-T10'.
[2021-08-02 13:39:53 UTC] make_available() Input file 'masks.png' is on a local provider
[2021-08-02 13:39:53 UTC] setup_and_submit_job() Exception raised while setting up: CbrainError: No scratch content available for userfile 15 on resource LocalBourreau

( run with ruby 2.4 on ace-vh-2. Maybe test environment setup issue, but old BoutiquesDemo works file)

prioux:
@MontrealSergiy That's a configuration error, yes. It means you DP cache and/or your ScratchDataProvider isn't configured properly on your LocalBourreau

MontrealSergiy:
Yup you are right, works now

image

MontrealSergiy:
One minor issue is that if file is not declared as cbrain list file, it would be handled as non-list even if has .cbcsv extension. I guess this means some issues with running Tractoflow with loris. Not sure how is it important

MontrealSergiy (approved):
Works as intended. Does not support, however, file list only buy extension, should be declared as Cbrain File List , which complicates loris integration. Not sure though how important is it.

PR #1148, opened 2021-08-24
Individual download

### LINKED ISSUE #1133: Made files downloadable from within a file collection
With the new viewer capability, we have lost the capability of CBRAIN to allow users to download files within a file collection. 

CBRAIN should allow the user to download individual file from a file Collection. 

Similar to what was done in: 
https://github.com/aces/cbrain/pull/526/commits/f0199cc5560d0263c7295f5b8afcecce94780f3b

Maybe CBRAIN can allow the user to download several files from a FileCollection in this case it should download as an archive.

natacha-beck:
Close #1133


BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/file_collection/views/_file_collection_top_table.html.erb
```diff
@@ -37,8 +37,7 @@
       <%= select_all_checkbox "collection_checkbox" %>
```
prioux, line 37: There's a bug here, it was there before this PR : the TH is not inserted in the table if the file is not synchronized, but all rows must have the same number of columns.

prioux: Fix this unrelated bug please.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/file_collection/views/_plain_file_list_row.html.erb
```diff
@@ -60,6 +60,14 @@
     </div>
 </td>
 
+<td>
+  <% if file.symbolic_type == :regular && file.size > 0 && file.size <  UserfilesController::MAX_DOWNLOAD_MEGABYTES.megabytes %>
```
prioux, line 64: why the `file.size > 0` test?

prioux (changes requested):
Minor changes requested

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/file_collection/views/_plain_file_list_row.html.erb
natacha-beck: To avoid download of empty files.

prioux: Obviously. My mistake.

PR #1151, opened 2021-09-13
breaking lines in boutiques param descriptors resolves #1150

### LINKED ISSUE #1150: Boutiques parameter descriptors with linebreaks
Presently the line breaks in boutiques parameter descriptors are ignored, but users ask for line break support. This will help to better structure long parameters.

MontrealSergiy:
Darius would like to better format long parameter descriptors in PhysIO, so asked to add support for line breaks

https://github.com/dariusvalevicius/PhysIO_CBRAIN/blob/master/cbrain_task_descriptors/tapasphysio.json

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
+++ b/BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
@@ -21,6 +21,8 @@
 #
 
 require 'fileutils'
+require 'erb'
+include ERB::Util
 
 # This module handles generation of CbrainTasks from schema and descriptor
 # files. The generated code can be added right away to CBRAIN's available tasks
@@ -30,6 +32,8 @@ require 'fileutils'
 module SchemaTaskGenerator
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  include ERB::Util
 
   # Directory where descriptor schemas are located
   SCHEMA_DIR = "#{Rails.root.to_s}/lib/cbrain_task_generators/schemas"
@@ -314,6 +318,9 @@ module SchemaTaskGenerator
   # can be provided to be used as the source for the CBRAIN Revision_info constant
   # that will be created inside the generated task class.
   def self.generate(schema, descriptorInput, strict_validation = true, file_for_revision_info = nil)
+
+    include ERB::Util
+
     descriptor = self.expand_json(descriptorInput)
     name       = self.classify(descriptor['name'])
     schema     = Schema.new(schema) unless schema.is_a?(Schema)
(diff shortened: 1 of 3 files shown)


prioux:
Why are you including erb and ERB::Util at all these places?!?? These were never necessary before.

MontrealSergiy:
To avoid error messages for undefined ` h```  method , namely when running rake

Should I put ERB::Util somewhere else or what?

> Why are you including erb and ERB::Util at all these places?!?? These were never necessary before.

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -301,7 +301,7 @@
 <%%# Generate a parameter's description block for +desc+ %>
 <%% description = lambda do |desc, isGroupMember = false| %>
   <span class="tsk-prm-desc">
-    <%%= desc %><%%= "<br><br>".html_safe if isGroupMember %>
+    <%%= desc.split("\n").map { |s| h(s) }.join("<br>").html_safe if desc %><<%%= "<br><br>".html_safe if isGroupMember %>
```
prioux, line 304: You have a syntax error here, you have one too many `<` (search for `<<`)

BrainPortal/lib/cbrain_task_generators/schema_task_generator.rb
```diff
@@ -21,6 +21,8 @@
 #
 
 require 'fileutils'
+require 'erb'
```
prioux, line 24: All modifications to `schema_task_generator.rb` are unnecessary

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -450,7 +450,9 @@
         <%= name %> <%% groupCheckbox.('<%= id %>',<%= defaultChecked %>) %>
       </h3>
       <h5 style="font-size : 9.5pt; font-weight: normal; padding-top: 2px;" class="grp-desc">
-        <%= groupDescription %>
+        <% groupDescription.split('\n') do |paragraph|%>
```
prioux, line 453: That code change does not split lines in the rendered page, there are no new `<br>` introduced

prioux (changes requested):
Please fix all the problems

prioux:
I tried with a plain CBRAIN that introduced ONLY the change at line 304 of `BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb` for a field that contains a single "\n" in the description. it works if we fix the syntax error:

Before:
Screen Shot 2021-09-13 at 15 27 53
After:
Screen Shot 2021-09-13 at 15 28 47

There was absolutely no need for any of the changes in the schema task generator.

MontrealSergiy:
ok, will do

MontrealSergiy:
Yes, with one line change, no import is needed. I guess it is better than nothing

PR #1152, opened 2021-09-20
boutiques parameter line breaks

MontrealSergiy:
This is to allow boutiques parameters descriptions include line breaks (Darius request).

I was not able to reopen previous https://github.com/aces/cbrain/pull/1151, so I opened this new pr

see 1151 for previous discussions. This is lightweight version compared to the previous, only tackles the params in task view but does not need any imports. 

PS. Personally, I think displaying parameters in verbatim mode (`<pre>`) would look nicer, though I know that not everybody likes fixed fonts. Perhaps, a subset of html/bbcode/markdown may allow for more eye candy.

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
@@ -438,7 +438,7 @@
     # Get group parameters
     id = group['id']
     name = group['name']
-    groupDescription = group['description']
+    groupDescription = group['description'].presence || ''
     groupParams = gIdToPrms[id]
     mbrs = gIdToMbrs[id]
     defaultChecked = group['one-is-required'] || required.any? { |p| mbrs.include? p["id"] }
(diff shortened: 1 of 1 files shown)


prioux:
What about doing the same thing for the group description too?

prioux (changes requested):
Do the same for groups

MontrealSergiy:
Doing 'same' for group description requires import which you do not like or some workaround. The request was just for parameter descriptors, group descriptors are usually short and less structured.  May be let's concentrate then on parameter descriptors?

prioux:
Wouldn't that work?

```diff
diff --git a/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb b/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
index 742311b7f..67b3f3548 100644
--- a/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
@@ -450,7 +450,9 @@
         <%= name %> <%% groupCheckbox.('<%= id %>',<%= defaultChecked %>) %>
       </h3>
       <h5 style="font-size : 9.5pt; font-weight: normal; padding-top: 2px;" class="grp-desc">
-        <%= groupDescription %>
+        <% (groupDescription.presence || "").split("\n").each do |line| %>
+          <%= line %><br>
+        <% end %>
       </h5>
     </div><%%= "<br>".html_safe %>
     <%# Write group parameters (required ones first) %>
```

MontrealSergiy:
Yup sure, I did not suggested it due to not using escape method. Note this is adding an extra unneeded line break. E.g. if even if there is no '\n' one line break is added. To avoid extra linebreak, something like

`       <h5 style="font-size : 9.5pt; font-weight: normal; padding-top: 2px;" class="grp-desc">
-         <%= groupDescription %>
+        <%= lines = (groupDescription.presence || "").split("\n") # map linebreaks into html
+        <% lines.shift       
+        <% lines.each do |line| %>
+           <br><%= line %>
+        <% end %>
       </h5>`

well can added whichever difference is marginal

prioux:
We're in the context of an `<H5>` anyway, so '<P>' are probably not allowed or recommended.

If putting the `<BR>` tag before the lines improves the esthetics, do it that way.

MontrealSergiy:
I do not see any noticeable change in visuals (Chrome), my suggestion was just to avoid unnecessary tags. Accepted your version

PR #1153, opened 2021-09-21
New Boutiques Integrator

prioux:
To test this codebase, I suggest you compare the old boutiques form with the new one in this way:

1. check out this branch
2. run the migration
3. pick a boutiques descriptor of your choice (any of them) and copy it to the new subdirectory `BrainPortal/cbrain_plugins/installed-plugins/boutiques_descriptors` (this subdirectory is created initially empty)
4. edit that descriptor so that you change the `"name"` attribute
5. when you boot your portal, the descriptor will be read and a new Tool will automatically configured to match that name
6. when you boot your bourreau, the descriptor will also be read and will create also a ToolConfig for it, which you will need to launch the job

I typically test the form by not even starting the bourreau but making sure it is marked as 'online' in the DB, and when I click "launch task" I make sure to leave the "data provider for results" selection box empty so that the form is returned to me. This allows an infinite number of rounds of testing the form elements.


MontrealSergiy:
a few tools, like PhysIO do not support full path to input files or folders, which seems to cause an issue with `bosh exec launch`. This seems to be an obscure boutiques issue

prioux:
@MontrealSergiy Can you give me an example of what you say? As much info as possible. Because the integrator, just like the previous one, never provides full path for any files arguments. It always run boutiques commands with plain basenames (which happen to be symlinks).

MontrealSergiy:
then there is no problem
> @MontrealSergiy Can you give me an example of what you say? As much info as possible. Because the integrator, just like the previous one, never provides full path for any files arguments. It always run boutiques commands with plain basenames (which happen to be symlinks).

MontrealSergiy:
what could be ```CBRAIN initial check failed: a090_ensure_boutiques_descriptors_are_loaded
uninitialized constant BoutiquesBootIntegrator::BoutiquesTask```

when I put demo descriptor into the said folder and launch the portal?

```{
    "name": "BoutiquesDemoNEW",
    "tool-version": "5.1.2",
    "schema-version": "0.5",
    "author": "Pierre Rioux",
    "description": "A demo CBRAIN task using a Boutiques descriptor. This program runs the UNIX 'du' command on a file or directory.",
    "descriptor-url": "https://github.com/aces/cbrain/blob/master/BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task_descriptors/boutiques_demo.json",
    "command-line": "du [ALL] [HUMAN] [FOLLOW_LINKS] [CBRAIN_INPUT] | tee [DU_OUTPUT_NAME]",
    "inputs": [
        {
            "name": "Input file or directory",
            "id": "my_input",
            "description": "Any file or directory in CBRAIN",
            "type": "File",
            "optional": false,
            "list": false,
            "value-key": "[CBRAIN_INPUT]"
        },
        {
            "name": "Name of report",
            "id": "my_output_name",
            "description": "The name of the text report containing the output of the 'du' command",
            "type": "String",
            "optional": false,
            "list": false,
            "value-key": "[DU_OUTPUT_NAME]"
        },
        {
            "name": "All files",
            "id": "option_a",
            "description": "Whether or not to provide a breakdown of each and every file",
            "type": "Flag",
            "optional": true,
            "command-line-flag": "-a",
            "value-key": "[ALL]"
        },
        {
            "name": "Human readable values",
            "id": "option_h",
            "description": "If true, units will be shown in readable human values, otherwise in blocks",
            "type": "Flag",
            "optional": true,
            "command-line-flag": "-h",
            "value-key": "[HUMAN]"
        },
        {
            "name": "Follow initial symlinks",
            "id": "option_upper_h",
            "description": "If true, initial symlinks are followed. Needed for file collections in CBRAIN",
            "type": "Flag",
            "optional": true,
            "default-value": true,
            "command-line-flag": "-H",
            "value-key": "[FOLLOW_LINKS]"
        }
    ],
    "output-files": [
        {
            "name": "Output report",
            "id": "du_report_out",
            "description": "The output of the 'du' command",
            "optional": false,
            "list": false,
            "path-template": "[DU_OUTPUT_NAME]"
        }
    ],
    "groups": [
        {
            "id": "files",
            "name": "Files",
            "members": [
                "my_output_name",
                "my_input"
            ]
        },
        {
            "id": "options",
            "name": "Options",
            "members": [
                "option_a",
                "option_h",
                "option_upper_h"
            ]
        }
    ],
    "tags": {
        "domain": [
            "boutiques",
            "testing",
            "cbrain",
            "platform",
            "internal"
        ]
    },
    "suggested-resources": {
        "cpu-cores": 1,
        "ram": 1,
        "walltime-estimate": 60
    },
    "custom": {
        "cbrain:readonly-input-files": true
    }
}
```

MontrealSergiy:
trace [
[logb.log](https://github.com/aces/cbrain/files/7213880/logb.log)
](url)

prioux:
@MontrealSergiy You forgot to run 'rake cbrain:plugins:install:plugins' ; this code has new plugin classes that need to be installed, in particular BoutiquesTask (located in `BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task` )

prioux:
I adjusted step 2 in the PR description

MontrealSergiy:
BTW I created a draft scripty to mass rename boutiques
https://gist.github.com/MontrealSergiy/d69ac087953a5981984f278ef349e8a4

MontrealSergiy:
getting weird `This task requires 1 mandatory file(s) and 0 optional file(s)
Input folder (required)` error when launching my task on a folder

prioux:
It's not weird @MontrealSergiy the new integrator validates the number of files you select vs the number of files the task wants. You descriptor wants just one file, and you have more than one selected when you click the 'launch' button.

MontrealSergiy:
```
<html><body>
<!--StartFragment--><header>
  <h1>
    NameError in
    Tasks#new
  </h1>
</header>

<div id="container">
  <p>
    Showing <i>/data1/user-data/sboroday/cbrain/BrainPortal/app/views/tasks/cbrain_plugins/installed-plugins/cbrain_task/boutiques_task/views/_form_js.html.erb</i> where line <b>#48</b> raised:
  </p>
  <pre><code>undefined local variable or method `file_input' for #&lt;#&lt;Class:0x007f9ec176ede8&gt;:0x007f9ec31bb200&gt;
Did you mean?  file_inputs</code></pre>

      <div class="source " id="frame-source-0">
      <div class="info">
        Extracted source (around line <strong>#48</strong>):
      </div>
      <div class="data">
        

46 47 48 49 50 51 | file_inputs = @descriptor.file_inputs     if file_inputs.count == 1 && ! file_inputs[0].list       single_file_id = file_input.id     end %>
-- | --


</div></div></div><!--EndFragment-->
</body>
</html>```

**Comment by MontrealSergiy** [comment]
image

**Comment by prioux** [comment]
Yep, that's a bug. Instead of adding a screenshot, please just put a comment in the code where the bug is.

**Comment by MontrealSergiy** [comment]
now Task creates but fails on setup, cannot find bosh

**Comment by prioux** [comment]
Well that has nothing to do with CBRAIN, it's you shell setup that's wrong. Make sure bosh is in the execution PATH. Check your bashrc using a non interactive shell, e.g. `ssh localhost which bosh` and 'ssh localhost type -p bosh` (just to check, assuming your bourreau connects to the current machine, though you should not use 'localhost' as the hostname for the bourreau)

File: BrainPortal/app/models/boutiques_cluster_task.rb [hunk 1]
```diff
@@ -0,0 +1,354 @@
...
+        bosh exec simulate
+          -i #{invoke_json_basename.bash_escape}
+          #{boutiques_json_basename.bash_escape}
+      SIMULATE
+      simulate_com.gsub!("\n"," ")
+      simulout = IO.popen(simulate_com) { |fh| fh.read }
+      simulout.sub(/^Generated.*\n/,"") # header junk from simulate
+      commands = <<-COMMANDS
+        # Main tool command, generated with bosh exec simulate
+        #{simulout.trim}
```
MontrealSergiy, line 153: I have undefined method trim errors

prioux: It's a bug I will fix right now

prioux: Mea culpa. That will teach me about making last minute changes.

MontrealSergiy:
The autogenerated BoutiquesDemo config tasks runs and succesfully terminates with minor differences in internal db representation, such as params[:invoke] . 

The manually added tool config fails with error ```
Coding error: method default_launch_args() for 
BoutiquesTask::BoutiquesDemoNEW raised an exception: NoMethodError: 
undefined method `inputs' for nil:NilClass

invoke_struct = self.descriptor_for_before_form 
      .inputs
      .reject { \|input\|   input.type == 'File'
``

MontrealSergiy:
The new style task do not have description. For old tools the description is added automatically 

image

prioux:
The missing description was a bug, I made a fix just now.

prioux:
The bug about 'undefined method inputs' I cannot investigate, but it seems to be a config issue. This happens when a boutiques task for a tool named "Abc" with a tool config with version_name 'mnop' can't find a boutiques descriptor installed that has both 'Abc' and 'mnop' in it. it's possible it's a problem with the casing you chose (you have 'NEW' in uppercase in the name of the tool, but the generated class and/or name in the decriptor is more like 'New').

MontrealSergiy:
Once again, to make sure we are on the same page. That  'undefined method inputs'  happens when an additional tool config is added via UI. So, yes that is 'mnop' is not in any boutiques descriptor. Occasionally we create boutiques that refer a docker but then create a config with a manually uploaded/registered singularity file. There maybe several tools with manually created tool configs in the portal. That is 'mnop' is not in any boutiques descriptor. If this feature (manually added tool config) is not to be supported, it could be better to block it for new style integrated tools.

> The bug about 'undefined method inputs' I cannot investigate, but it seems to be a config issue. This happens when a boutiques task for a tool named "Abc" with a tool config with version_name 'mnop' can't find a boutiques descriptor installed that has both 'Abc' and 'mnop' in it. it's possible it's a problem with the casing you chose (you have 'NEW' in uppercase in the name of the tool, but the generated class and/or name in the decriptor is more like 'New').

prioux:
No, we are supposed to be able to add manually any tool config we want. But oif the version_name in the tool_config does not match a descriptor already installed, then you need to provide the path to the descriptor manually. That's what there is a field for that in the tool config's form. In the show page an automated match is hsown with "automatic":
Screen Shot 2021-09-24 at 10 27 44

Then when you edit you can override this with ANY tool config, even one for a different tool or a version that doesn't match the tool_config's version_name:

Screen Shot 2021-09-24 at 10 28 11

prioux:
So basically, if you create a new tool_config using the UI, and you give it a version_name that doesn't match any known descriptor, you should provide the path to the descriptor in that box.

prioux:
@MontrealSergiy Thanks for testing this so well, by the way, I appreciate it.

MontrealSergiy:
after editing a task and restart got error 

`Cannot find attribute that match reader method 'outputs'`

MontrealSergiy:
Now starting/Launching  (new type) task fails with an exception

```
    # Seems weird, it is to accommodate rails5 deprecation warning for find (accept only id)
    id = id.id if id.is_a?(Userfile)
    self.accessible_for_user(user, options).find(id)
  end

```

prioux:
I found the bug about 'outputs', will commit in a minute

prioux:
I don't know about the last problem (`Couldn't find Userfile without an ID`). Can you show me what gets posted to the rails server? The `Parameters:` line in the logs after the `Started POST` ?

MontrealSergiy:
```
Started POST "/tasks" for ::1 at 2021-09-24 12:57:04 -0400
  ActiveRecord::SchemaMigration Load (0.4ms)  SELECT `schema_migrations`.* FROM `schema_migrations`
Processing by TasksController#create as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"[FILTERED]", "tool_id"=>"38", "cbrain_task"=>{"params"=>{"interface_userfile_ids"=>["60"], "invoke"=>{"my_output_name"=>"fr", "my_input"=>"60", "option_a"=>"0", "option_h"=>"0", "option_upper_h"=>"1"}}, "tool_config_id"=>"37", "results_data_provider_id"=>"2", "description"=>"", "user_id"=>"1", "group_id"=>"2"}, "load_preset_id"=>"", "save_preset_id"=>"", "save_preset_name"=>"", "commit"=>"Start BoutiquesDemoNEW"}
```

> I don't know about the last problem (`Couldn't find Userfile without an ID`). Can you show me what gets posted to the rails server? The `Parameters:` line in the logs after the `Started POST` ?

prioux:
Thanks. I see nothing wrong there. But I have another hypothesis. Is it possible this is a test of a tool with a newly-created tool config? And it is possible you didn't set the flag 'Does not modify its inputs' in that tool config?

Screen Shot 2021-09-24 at 13 22 47

MontrealSergiy:
I test with autocreated config. I changed file project permission to write/read.

prioux:
Okay. I looked at the code and I found other verifications where already being made well before that point, so now I don't know why it's happening. I'll need more info. Can you give me precise instructions on how to reproduce the problem? Maybe with a minimal descriptor? (Don't post that here, let's keep the github discussions short)

MontrealSergiy:
works now

MontrealSergiy:
Integer validation fails.  To reproduce take a tool that has integer input params, like fslstatsNEW. The ask will fail with error ' There were problems with the following fields:

    P must be an integer

prioux:
At what step does this fail? In the portal form? During execution?

natacha-beck:
You have the same issue in the new integrator: https://github.com/aces/cbrain/issues/1154

MontrealSergiy:
> At what step does this fail? In the portal form? During execution?

on form after click on start

image

prioux:
The integer verif bug is fixed too. Thanks.

BrainPortal/app/models/boutiques_cluster_task.rb
```diff
@@ -0,0 +1,354 @@
...
+
+        # Add provenance logs
+        all_file_input_ids = descriptor.file_inputs.map do |input|
+          invoke_params[input.id]
+        end.compact.uniq
+        parents = Userfile.where(:id => all_file_input_ids).to_a
+        self.addlog_to_userfiles_these_created_these(parents, [outfile]) if parents.present?
+
+        # If there is only one input file, we move the output under it
+        outfile.move_to_child_of(parents[0]) if parents.size == 1
```
natacha-beck, line 266: If we have multiple files but only one is mandatory, can we move the output under it too ?

BrainPortal/app/models/boutiques_cluster_task.rb
```diff
@@ -0,0 +1,354 @@
...
+  def exit_status_filename
+    ".qsub.exit.#{self.name}.#{self.run_id}"
+  end
+
+  # Returns either :simulate or :launch .
+  #
+  # In the mode 'simulate', at the moment of creating
+  # the tool's script in cluster_commands(), the
+  # output of 'bosh exec simulate' will be substituted in
+  # the script ot generate the tool's command.
```
natacha-beck, line 317: Typo `ot` -> `to`

BrainPortal/app/models/boutiques_cluster_task.rb
prioux: It's a good idea. I'll do it.

BrainPortal/app/models/tool_config.rb
```diff
@@ -465,4 +465,89 @@ def to_carmin #:nodoc:
     }
   end
 
+  ##################################################################
```
natacha-beck, line 468: Some missing `#:nodoc:` in this file.

prioux: Yes, I will do a full review of all the #:nodoc: of all new methods soon (some of them need to actually be documented)

natacha-beck:
When try to launch multiple tasks with a cbcscv file:

```
 Showing /home/users/nbeck/cbrain/BrainPortal/app/views/tasks/cbrain_plugins/installed-plugins/cbrain_task/boutiques_task/views/_form_js.html.erb where line #39 raised:

undefined local variable or method `user' for #<#<Class:0x0000000007176860>:0x00007f32717136a8>
Did you mean?  super

 Extracted source (around line #39): 

      f.sync_to_cache # silly and blocking
      ids = f.ordered_raw_ids.reject { |i| i.blank? || i == 0 }
      names = Userfile.find_all_accessible_by_user(user, :access_requested => :read).where( 'userfiles.id' => ids).pluck(:name)
      [ f.name, names ]
    end.to_h
%>

```





```


https://github.com/aces/cbrain/pull/1153/files#diff-f5c61e7a45a4d8eedb970868a07e57940aceacbf5b6e29dad09886b8140a93efR39

prioux:
@natacha-beck It's a bug. Will fix now.

MontrealSergiy:
New issue. BoutiquesDemoNEW task terminates succesfully, yet it actually fails with error

```Bourreau Exception: ActiveRecord::SubclassNotFound The single-table inheritance mechanism failed to locate the subclass: 'BoutiquesTask::BoutiquesDemoNEW'. This error is raised because the column 'type' is reserved for storing the class in case of inheritance. Please rename this column if you didn't intend it to be used for storing the inheritance class or overwrite CbrainTask.inheritance_column to use another column for that information```

even if adding path to a boutiques manually resolve issue, the succesful status is inappropriate  
image

prioux:
That's caused by your bourreau server not having been restarted. When code is modified the Rails platform tries to reload all the ruby classes. It first removes all the classes that are already defined in memory and then reloads the files. But that doesn't reload the classes that are created dynamically at boot time.  Just make sure to reboot your server if you modify any files (including checking out a new branch).

MontrealSergiy:
Ok then. I thought I restarted it but after restarts seems work ok.

natacha-beck (approved):

Issue #1127, opened 2021-06-08
Multiple tasks are launched if many files are selected.


MontrealSergiy:
This one may be a feature. If several files in file explorer are selected (but only one is used), several tasks, one for each file is launched. 

It might be actually a useful feature, as relieves user from creating CBRAIN File lists, but a bit unexpected. IMHO a warning might be appropriate.

natacha-beck:
Seems more like a bug to me and not a feature.As you said we have the CBCSV file to do that.

natacha-beck:
Close by #1165

prioux:
Closed by #1165

PR #1165, opened 2021-10-27
Fix view in boutique integrator #1127

natacha-beck:


prioux:
I'd like the same changes applied to the new integrator. Sorry! :-)

prioux (changes requested):
See comment

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -94,6 +94,12 @@
   </label>
 <%% end %>
 
+<%% paragraph = lambda do | text | %>
+  <span class="tsk-prm-par">
+    <%%= text   %>
+  <span class="tsk-prm-par">
```
prioux, line 100: I think that's a bug here, did you mean `</span>` instead?

natacha-beck:
Yes.

natacha-beck:
No need to get the same logic within the new integrator, you only support launching multiple tasks with a cbcsv file.

prioux:
Can you double check that the message in the section "launching multiple tasks" in

BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_edit_help.html.erb

is appropriate?

BrainPortal/lib/cbrain_task_generators/templates/task_params.html.erb.erb
```diff
@@ -420,7 +423,7 @@
     <%- end -%>
         )
   <%- end -%>
-  <%- if param['description'] -%>
```
prioux, line 423: This will remove the descriptions of all other inputs if the decriptor is a single_file scenario. Not what you wanted.

prioux (changes requested):

PR #1168, opened 2021-11-05
[ADD] User for automated system

### LINKED ISSUE #1119: Create new subclass of NormalUser for automated systems
To recognize automated systems from live humans, we could create a subclass of NormalUser called 'AutomatedUser' (or something like that).

The class would, at first, implement nothing different from its superclass.

natacha-beck:
Close #1119


prioux:
Two problems.

1) If I query for NormalUser, I'll also get the AutomatedUsers:

```ruby
something = NormalUser.where(:city => 'Montreal')
```

2) We can't create these AutomatedUsers in the user creation form.

prioux:
And yes I know the original issue mentioned that it would be a subclass of NormalUser

PR #1170, opened 2021-11-10
[FIX] tool_config creation for Boutiques with Docker image

### LINKED ISSUE #774: Boutiques/CBRAIN and Singularity/Docker integration.
If a Boutiques descriptor specifies a `container-image` of type `docker`, CBRAIN will not configure a new ToolConfig on a cluster that does not have a `docker` engine defined. 

The process to convert a Docker container as a Singularity container is stable enough to set up a ToolConfig for this task and to define the Container engine as `Singularity` if Singularity is present on the cluster. 

By doing that CBRAIN will be able to create and used Singularity image by building them from a docker reference.

natacha-beck:
Fixes #774

### DIFF
--- a/BrainPortal/app/models/tool_config.rb
+++ b/BrainPortal/app/models/tool_config.rb
@@ -524,7 +524,12 @@ class ToolConfig < ApplicationRecord
     ).first
     return tc if tc
 
-    container_info = descriptor.container_image || {}
+    container_info   = descriptor.container_image || {}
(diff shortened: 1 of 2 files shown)


prioux (changes requested):
This seems ok but we shoudl try to make the same logic in the new integrator.

Start with app/models/tool_config.rb in the method `self.create_from_descriptor`
Also check tool.rb, it has somethign similar.

PR #1172, opened 2021-11-11
[FIX] access profile's log when update a user

### LINKED ISSUE #597: Adding a access profile to a user from the user's show page does not update the access profile's log
From the user 'show' page, when assigning new access profiles or removing them, we're missing logging of the change on the access profile object's log

natacha-beck:
Closed #597


prioux:
Extract code in respond_to block into controller.

Have something like:

```ruby
success = @user.save_with_logging(...)
if success
   housekeeping and log
end

respond_to do
  messsages, redirects etc
end
```


Make the variable last_update obsolete

PR #1177, opened 2021-11-23
Adds the user's private UserGroup in NH

### LINKED ISSUE #1176: Broken link on `latest updates task` in NH
If a user has a task on his `personal project`. On the `Latest update tasks` on the `Dashboard` the link to the project will throw an error saying that the user have no access to the ressource.

prioux:
Will fix #1176

### DIFF
--- a/BrainPortal/lib/neurohub_helpers.rb
+++ b/BrainPortal/lib/neurohub_helpers.rb
@@ -30,8 +30,9 @@ module NeurohubHelpers
   # If +id_or_project_ is already a Group, it will make sure it's
   # a valid one. The WorkGroup is returned. If the validation fails,
   # an exception ActiveRecord::RecordNotFound is raised.
-  def find_nh_project(user, id_or_project, check_license=true)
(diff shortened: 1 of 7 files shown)


BrainPortal/lib/neurohub_helpers.rb
```diff
@@ -47,9 +48,10 @@ def find_nh_project(user, id_or_project, check_license=true)
 
   # For the user +user+, this method will return
   # neurohub projects ('available' groups of class WorkGroup)
-  def find_nh_projects(user)
-    user.listable_groups
-        .where(:type => 'WorkGroup')
+  def find_nh_projects(user, allow_own_group=true)
```
prioux, line 51: I realize now that this modification is not currently necessary.

prioux:
What needs to be checked:

We must make sure that users can't do with the UserGroup some thigns that are normally allowed on WorkGroups. A UserGroup cannot receive invitations, cannot be edited, its permissions cannot be changed, etc etc

PR #1183, opened 2021-12-06
Don't show 'link to globus' when admin browsing a user's account page fixes #1179

### LINKED ISSUE #1179: Don't show 'link to globus' when admin browsing a user's account page
An admin can browse a user's account page, and if the user hasn't linked a globus account, the button 'Link To A Globus Identity' will show up. That button should be hidden. If the admin clicks on it it will make the link with the admin's account anyway.

Change in both CBRAIN and NeuroHub.

MontrealSergiy:
see the attached issue for full description.
Note that the issue presents itself only on CBRAIN part, just in the case the changes done on NeuroHub side too

### DIFF
--- a/BrainPortal/app/views/nh_users/_form.html.erb
+++ b/BrainPortal/app/views/nh_users/_form.html.erb
@@ -115,8 +115,10 @@
       <% glob_prov = @user.meta[:globus_provider_name] %>
       <% if glob_user.blank? %>
         <p class="btn-solid globus">
-          <%= link_to @globus_uri, :class => "" do %>
-            Connect your NeuroHub account to a Globus identity provider
+          <% if @user.id == current_user.id  # show button only on user own page, hide on other users pages %>
+            <%= link_to @globus_uri, :class => "" do %>
+              Connect your NeuroHub account to a Globus identity provider
+            <% end %>
           <% end %>
         </p>
         <p class="field_note">
@@ -132,11 +134,13 @@
       <% end %>
       <% if glob_user.present? %>
         <p class="btn-solid globus bg-warning-bg">
-          <%= link_to nh_unlink_globus_path,
-              :class => "",
-              :method  => :post,
-              :data    => { :confirm => "Are you sure you want to unlink your account from this Globus identity?" } do %>
-              Remove this Globus identity from your NeuroHub account
+          <% if @user.id == current_user.id  # show button only on user own page %>
+            <%= link_to nh_unlink_globus_path,
+                :class => "",
+                :method  => :post,
+                :data    => { :confirm => "Are you sure you want to unlink your account from this Globus identity?" } do %>
+                Remove this Globus identity from your NeuroHub account
+            <% end %>
           <% end %>
         </p>
       <% end %>
--- a/BrainPortal/app/views/nh_users/_form.html.erb
+++ b/BrainPortal/app/views/nh_users/_form.html.erb
@@ -114,13 +114,13 @@
       <% glob_user = @user.meta[:globus_preferred_username] %>
       <% glob_prov = @user.meta[:globus_provider_name] %>
       <% if glob_user.blank? %>
-        <p class="btn-solid globus">
-          <% if @user.id == current_user.id  # show button only on user own page, hide on other users pages %>
+        <% if @user.id == current_user.id  # show button only on user own page, hide on other users pages %>
+          <p class="btn-solid globus">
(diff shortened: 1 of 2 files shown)


BrainPortal/app/views/users/show.html.erb
```diff
@@ -172,13 +172,16 @@
...
                :method => :post,
                :data   => { :confirm => "Are you sure you want to unlink your account with this Globus identity?" }
-             )
-         %>
-       <% else %>
-         (No Globus identity linked to your account)<br>
-         <%= link_to 'Link a Globus identity', @globus_uri, :class => 'button globus_button' %>
-       <% end %>
-    <% end %>
+              )
```
prioux, line 175: I don't understand the view logic here.

If the viewing the page as an admin user, we still get the unlink button, which if pressed will not work properly.

Why not just take the view code as it was originally and JUST put a `if current_user.id == @user.id` around the buttons, and nothing else?

MontrealSergiy: The unlink button was not mentioned in the issue.

prioux: I know it wasn't and it's good your thought about it, but the fix is wrong. Just hide it.

MontrealSergiy: sure, putting condition around each button. Then the Globus section cell itself will remain, albeit potentially empty.

prioux: It won't be totally empty, see above, it would say `Provider name:` and `Provider user:`

prioux: Just look at the original code... if you put the if around each button, it works for both normal users and admin users.

prioux (changes requested):

BrainPortal/app/views/nh_users/_form.html.erb
```diff
@@ -132,11 +134,13 @@
       <% end %>
       <% if glob_user.present? %>
         <p class="btn-solid globus bg-warning-bg">
-          <%= link_to nh_unlink_globus_path,
-              :class => "",
-              :method  => :post,
-              :data    => { :confirm => "Are you sure you want to unlink your account from this Globus identity?" } do %>
-              Remove this Globus identity from your NeuroHub account
+          <% if @user.id == current_user.id  # show button only on user own page %>
```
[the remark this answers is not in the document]
MontrealSergiy: P is moved under if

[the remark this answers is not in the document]
prioux: Where? That's not what I see in the diff. Look at it.

BrainPortal/app/views/nh_users/_form.html.erb
```diff
@@ -115,8 +115,10 @@
       <% glob_prov = @user.meta[:globus_provider_name] %>
       <% if glob_user.blank? %>
         <p class="btn-solid globus">
-          <%= link_to @globus_uri, :class => "" do %>
-            Connect your NeuroHub account to a Globus identity provider
+          <% if @user.id == current_user.id  # show button only on user own page, hide on other users pages %>
```
prioux, line 118: Same problem here

prioux (changes requested):
I just extracted your PR's code to test it out and discovered that an admin user cannot even look at another user's account page in neurohub. This means that there are no changes needed in the file `nh_users/_form.html.erb`. If you didn't mention this, does that mean you made these modifications without even trying them out yourself?

MontrealSergiy:
It might. Yet I mentioned that in the PR (scroll up to read)

> I just extracted your PR's code to test it out and discovered that an admin user cannot even look at another user's account page in neurohub. This means that there are no changes needed in the file `nh_users/_form.html.erb`. If you didn't mention this, does that mean you made these modifications without even trying them out yourself?

BrainPortal/app/views/nh_users/_form.html.erb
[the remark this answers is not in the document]
MontrealSergiy: yes, sorry, failed to push. Pushed now. While the bug does not exist on NeuroHub side, one might add user view capability there one day. That person will likely pay no attention to the need to hide button.

prioux:
This looks good now, I'll make a quick check later this afternoon and merge it.

prioux:
Ok, even though on NeuroHub an admin can't look at a user's account page, I'm ok with the changes.

PR #1184, opened 2021-12-06
Formalize a little the specifications for overlay files in tool config. resolve #1026

### LINKED ISSUE #1026: Formalize a little the specifications for overlay files in tool config.
Following #1016, a tool config now has a text area where the admin can configure overlays. The current syntax supports three specification types:

```
/path/to/file.squashfs
dp:1234
dp:name_of_dp
```

I'd like to make this better looking and more formal.

I suggest we change the spec so that files are described with FILE URIs, and we can also specify overlays directly registered in CBRAIN too. Also, support embedded comments.

```
# A file located on the cluster
file:/path/to/file.squashfs
# All overlays configured for DP 1234
dp:1234
# My special overlay in CBRAIN
userfile:1234
```

We'll need to extract the parser into a more proper stand-alone method (right now we just plsit on whitespace and commas)

MontrealSergiy:
enhancement for tool config singularity overlay

see Formalize a little the specifications for overlay files in tool config. #1026

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -732,6 +732,8 @@ class ClusterTask < CbrainTask
       self.status == 'Setting Up'
     return false if self.workdir_archived?
 
+    self.tool_config.validate_overlays_specs if self.tool_config  # validate on setup
+
     begin
       self.addlog("Setting Up.")
       self.record_cbraintask_revs
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -732,7 +732,7 @@ class ClusterTask < CbrainTask
       self.status == 'Setting Up'
     return false if self.workdir_archived?
 
-    self.tool_config.validate_overlays_specs if self.tool_config  # validate on setup
+    self.tool_config.validate_overlays_specs if self.tool_config  # validate and sync files on setup
 
     begin
       self.addlog("Setting Up.")
--- a/BrainPortal/app/models/tool_config.rb
+++ b/BrainPortal/app/models/tool_config.rb
@@ -377,7 +377,7 @@ class ToolConfig < ApplicationRecord
     return @_data_providers_with_overlays_ if @_data_providers_with_overlays_
     specs = self.singularity_overlays_specs.presence
     return [] if specs.blank?
-    specs = specs.(/([\s,]|#.*$)+/).map(&:presence).compact
+    specs = specs.split(/([\s,]|#.*$)+/).map(&:presence).compact
     @_data_providers_with_overlays_ = specs.map do |spec|
       next nil unless spec =~ /^dp:/i
       id_or_name = spec.sub(/^dp:/i,"")
--- a/BrainPortal/app/models/tool_config.rb
+++ b/BrainPortal/app/models/tool_config.rb
@@ -459,12 +459,9 @@ class ToolConfig < ApplicationRecord
           if  !(userfile.sync_to_cache() rescue nil)
             self.errors.add(:singularity_overlays_specs,
                    " contains invalid specification '#{spec}'. The fetching data from data provider to cache failed.")
-          else
-
-            if userfile.name !~ /^\/\S+\.(sqs|squashfs)$/i
+          elsif  ! userfile.name.ends_with?('.sqs') && ! userfile.name.ends_with?('.squashfs')
               self.errors.add(:singularity_overlays_specs,
-              " contains invalid specification '#{spec}'. It should be a full path that ends in .squashfs or .sqs")
-            end
+              " contains invalid specification '#{spec}'. It should end in .squashfs or .sqs")            
           end
         else
           self.errors.add(:singularity_overlays_specs,
--- a/BrainPortal/app/models/tool_config.rb
+++ b/BrainPortal/app/models/tool_config.rb
@@ -65,6 +65,7 @@ class ToolConfig < ApplicationRecord
   scope           :global_for_bourreaux , -> { where( { :tool_id => nil } ) }
   scope           :specific_versions    , -> { where( "bourreau_id is not null and tool_id is not null" ) }
 
+
   api_attr_visible :version_name, :description, :tool_id, :bourreau_id, :group_id, :ncpus
 
   # To make it somewhat compatible with the ResourceAccess module,
@@ -340,10 +341,8 @@ class ToolConfig < ApplicationRecord
 
     # break onto commands based on comas, whitespace or comments
     specs = specs.split(/([\s,]|#.*$)+/).map(&:presence).compact
-    i = 0
     specs.map do |spec|
       # Full path or pattern
-      i += 1
       next spec if spec =~ /^\// # just use that - FIXME remove after successful migration
       cb_error "Spec #{spec}  starts with semicolon" if spec.start_with?(':')
       knd, id_or_name = spec.split(':', 2)
@@ -357,8 +356,8 @@ class ToolConfig < ApplicationRecord
         cb_error "DataProvider #{id_or_name} does not have any overlays configured." if dp_ovs.blank?
         next dp_ovs
       when 'file'
-        cb_error "Provide absolute file path in spec #{spec}." if ! spec.start_with?('\\')
-        next id_or_name
+        cb_error "Provide absolute filepath in spec #{spec}." if (Pathname.new id_or_name).relative?
+        next id_or_name  # despite the variable name id_or_name for local file, id (inode) is not supported, only (full) name supported
       when 'userfile'
         # admin can access all files
         userfile = Userfile.find(id_or_name) rescue nil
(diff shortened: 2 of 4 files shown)


BrainPortal/app/models/cluster_task.rb
```diff
@@ -732,6 +732,8 @@ def setup_and_submit_job
       self.status == 'Setting Up'
     return false if self.workdir_archived?
 
+    self.tool_config.validate_overlays_specs if self.tool_config  # validate and sync files on setup
```
prioux, line 735: What does that even do? It doesn't do anything. And if it crashes, it will crash the entire Bourreau Worker.

Validation methods are part of the database-side life cycle of the ActiveRecord objects, and we're not at all in any sort of code that is part of that life cycle.

BrainPortal/app/models/tool_config.rb
```diff
@@ -326,22 +326,45 @@ def cbrain_task_class
...
+  #         dp:1234
+  #      # My special overlay in CBRAIN
+  #         userfile:1234
   def singularity_overlays_full_paths
     specs = self.singularity_overlays_specs.presence
     return [] if specs.blank?
 
-    specs = specs.split(/[\s,]+/).map(&:presence).compact
+    # break onto commands based on comas, whitespace or comments
+    specs = specs.split(/([\s,]|#.*$)+/).map(&:presence).compact
```
prioux, line 341: You're breaking text on hash signs here, that's not right. If you want to support embedded comments, you must split by lines first. This code will crash on a string like "hello#bye".

prioux: The new spec should not allow several specifications on the same line, only one per line.

BrainPortal/app/models/tool_config.rb
```diff
@@ -326,22 +326,45 @@ def cbrain_task_class
...
     specs.map do |spec|
-      # Full path or pattern
-      next spec if spec =~ /^\// # just use that
+      # Full path or pattern (legacy)
+      next spec if spec =~ /^\// # just use that - FIXME remove after successful migration
+      cb_error "Spec #{spec}  starts with semicolon" if spec.start_with?(':')
```
prioux, line 345: Why all the low level string logic? Just match! Something like

```ruby
cb_error "bad spec message here" unless spec =~ /\A(file|dp|userfile):(\S+)\s*\z/
scheme, id_or_name = Regexp.last_match[1,2]
```

BrainPortal/app/models/tool_config.rb
```diff
@@ -326,22 +326,45 @@ def cbrain_task_class
...
+        cb_error "Can't find DataProvider #{id_or_name} for fetching overlays" if ! dp
+        dp_ovs = dp.singularity_overlays_full_paths rescue nil
+        cb_error "DataProvider #{id_or_name} does not have any overlays configured." if dp_ovs.blank?
+        next dp_ovs
+      when 'file'
+        cb_error "Provide absolute filepath in spec #{spec}." if (Pathname.new id_or_name).relative?
+        next id_or_name  # despite the variable name id_or_name for local file, id (inode) is not supported, only (full) name supported
+      when 'userfile'
+        # admin can access all files
+        userfile = Userfile.find(id_or_name) rescue nil
```
prioux, line 361: instead of invoking the finder and raising an exception then trapping it, be gentler:

```
userfile = Userfile.where(:id => id_or_name).first
```

which will return nil if it's not there and is a lot kinder on the execution stack

BrainPortal/app/models/tool_config.rb
```diff
@@ -326,22 +326,45 @@ def cbrain_task_class
...
+        cb_error "File not found for spec #{spec}." if ! userfile
+        userfile.sync_to_cache()
+        next userfile.cache_full_path()
+      else
```
prioux, line 365: This check i no longer necessary after by my regexp match above

BrainPortal/app/models/tool_config.rb
```diff
@@ -352,7 +375,7 @@ def data_providers_with_overlays
     return @_data_providers_with_overlays_ if @_data_providers_with_overlays_
     specs = self.singularity_overlays_specs.presence
     return [] if specs.blank?
-    specs = specs.split(/[\s,]+/).map(&:presence).compact
+    specs = specs.split(/([\s,]|#.*$)+/).map(&:presence).compact
```
prioux, line 378: I think this parsing, which duplicates what we do above, should be extracted in a method to make it more formal.

I suggest a method that returns an array of pairs, `[ [ scheme, value ], [scheme, value] ... ]`

BrainPortal/app/models/tool_config.rb
```diff
@@ -400,40 +423,72 @@ def validate_container_rules #:nodoc:
   # Verify that the admin has entered a set of
   # overlay specifications properly. One or several of:
   #
-  #    /full/path/to/something.squashfs
-  #    /full/path/to/pattern*/data?.squashfs
+  #    file:/full/path/to/something.squashfs
```
prioux, line 426: You're not documenting `userfile:1234`

BrainPortal/app/models/tool_config.rb
```diff
@@ -400,40 +423,72 @@ def validate_container_rules #:nodoc:
...
+            " contains invalid specification '#{spec}'. It should be a full path that ends in .squashfs or .sqs")
+        elsif (Pathname.new id_or_name).relative?
+          self.errors.add(:singularity_overlays_specs,
+                          " contains invalid specification '#{spec}'. It should be an absolute path ")
         end
-        if ! dp.is_a?(SingSquashfsDataProvider)
-          self.errors.add(:singularity_overlays_specs, "DataProvider '#{spec}' is not a SingSquashfsDataProvider")
+
+      when 'userfile' # db-registered file specification: "userfile:ID" e.g. "userfile:42"
+        userfile = Userfile.find(id_or_name) rescue nil
```
prioux, line 465: Again, see my comment above

prioux (changes requested):

BrainPortal/app/models/tool_config.rb
MontrealSergiy: better diagnosis. 

but if not needed can use that as well.

BrainPortal/app/models/cluster_task.rb
MontrealSergiy: the requirements included 'sync userfile on setup', is there a more appropriate place to do it?

BrainPortal/app/models/tool_config.rb
MontrealSergiy: ok got. I thought 


```# folder a files
file:/a/1.sqs file:/a/2.sqs file:/a/3.sqs
# folder b files
/b/0.sqs
# data providers
dp:3  dp:4```

could be of use, but can limit to one liners

Comment (MontrealSergiy) [reply BrainPortal/app/models/tool_config.rb:341]: 1. basically I just try conform to the ticket's requirement for supporting old legacy specification format, which I believed allowed for multi spec lines. Alternatively, I can try to determine is spec of old or new format, e.g. by presence of colon (assuming no file name includes colon) and then apply either or validation function or new one. Not sure would it be overcomplication.
2. Advantage of statement per line notation, spaces in file names can be allowed `  /a/b/my beatiful brain.sqs` 
3. Would you like to have trailing comments `file:/x/x.sqs # that's file with my brain scans`
@prioux

Comment (prioux) [reply BrainPortal/app/models/tool_config.rb:341]: Ok more clearly:

1. one spec per line only
2. comments allowed on standalone lines and on spec lines too
3. don't allow spaces in the file specs

```
# comment
file:/a/b/c.sqs
file:/a/b/d.sqs #with comment
dp:3
dp:MainStore # by name shoudl work already
userfile:12455  # no support for 'by name' here though, must be an ID
```

Comment (prioux) [reply BrainPortal/app/models/tool_config.rb:341]: I suggest making an intermediate-level method `parsed_overlay_specs` that takes the text and returns something like

```
overlay_specs = tc.parsed_overlay_specs
-> [ [ 'file', '/a/b/c.sqs'], [ 'dp', '3' ] ... ]
```

File: BrainPortal/app/models/cluster_task.rb (hunk 1 above)
Comment (MontrealSergiy) [reply BrainPortal/app/models/cluster_task.rb]: well, removed, maybe redundant indeed

File: BrainPortal/app/models/tool_config.rb (hunk 2 above)
Comment (MontrealSergiy) [reply BrainPortal/app/models/tool_config.rb]: ok

File: BrainPortal/app/models/tool_config.rb (hunk 4 above)
Comment (MontrealSergiy) [reply BrainPortal/app/models/tool_config.rb]: ok

File: BrainPortal/app/models/tool_config.rb (hunk 6 above)
Comment (MontrealSergiy) [reply BrainPortal/app/models/tool_config.rb]: ok

File: BrainPortal/app/models/tool_config.rb (hunk 7 above)
Comment (MontrealSergiy) [reply BrainPortal/app/models/tool_config.rb:430]: added

File: BrainPortal/app/models/tool_config.rb (hunk 5 above)
Comment (MontrealSergiy) [reply BrainPortal/app/models/tool_config.rb:363]: somehow it seems easier and more readable with else

**Comment by MontrealSergiy** [comment]
Refactored, please take a look

File: BrainPortal/app/models/tool_config.rb [hunk 9]
```diff
@@ -326,22 +326,43 @@ def cbrain_task_class
...
+        cb_error "Can't find DataProvider #{id_or_name} for fetching overlays" if ! dp
+        dp_ovs = dp.singularity_overlays_full_paths rescue nil
+        cb_error "DataProvider #{id_or_name} does not have any overlays configured." if dp_ovs.blank?
+        dp_ovs
+      when 'file'
+        cb_error "Provide absolute filepath in spec #{spec}." if (Pathname.new id_or_name).relative?
+        id_or_name  # for local file, it could be full file name (no ids)
+      when 'userfile'
+        # db registered file, note admin can access all files
+        userfile = Userfile.where(:id => id_or_name).last
```
prioux, line 359: Small adjustements:
1) fetch the file using 'SingleFile' instead of 'Userfile', that way we make sure the admin has configured the ID of a file, not a directory
2) make the error message more clear the problem is in the overlay spec, because it will be difficult to understand the message `Userfile 12345 not found` in the logs if we don't know this error occursed during the stage where we build the overlays

BrainPortal/app/models/tool_config.rb
```diff
@@ -397,43 +416,75 @@ def validate_container_rules #:nodoc:
     return errors.empty?
   end
 
-  # Verify that the admin has entered a set of
+  # breaks down overlay spec onto a list of overlays
+  def parsed_overlay_specs(specs)
```
prioux, line 420: Make this a method that doesn't take any arguments. It should itself fetch the value of the raw specs with `self.singularity_overlays_specs.presence` and return the parsed version.

That means everywhere else you won't need to extract the specs and pass it in argument. It will make the code simpler at the beginning of `singularity_overlays_full_paths` and at the begining of `data_providers_with_overlays`

BrainPortal/app/models/tool_config.rb
```diff
@@ -397,43 +416,75 @@ def validate_container_rules #:nodoc:
...
+          self.errors.add(:singularity_overlays_specs, "DataProvider '#{id_or_name}' is not a SingSquashfsDataProvider")
         end
-        next
-      end
 
-      # Other
-      self.errors.add(:singularity_overlays_specs, "contains invalid specification '#{spec}'")
+      else
+        # Other errors
+        self.errors.add(:singularity_overlays_specs, "contains invalid specification'#{kind}:#{id_or_name}'")
```
prioux, line 486: Missing a space in the error message

BrainPortal/app/views/tool_configs/_form_fields.html.erb
```diff
@@ -212,17 +212,20 @@
...
-        a path with a pattern (e.g. <em>/a/b/data*.squashfs</em>),
+        a full path (e.g. <em>file:/a/b/data.squashfs</em>),
+        a path with a pattern (e.g. <em>file:/a/b/data*.squashfs</em>),
+        a registered file identified by ID (e.g. <em>userfile:123</em>)
         or a SquashFS Data Provider identified by its ID or name (e.g. <em>dp:123</em>, <em>dp:DpNameHere</em>).
         In the case of a Data Provider, the overlays will be the files that the provider uses.
-        Separate specifications with blank, commas, or newlines.
+        Each overlay specification should be on a separate line.
+        You can add comments, indicated with hash symbol <em>#</em>.
+        For example, <em>file:/a/b/atlas.squashfs # brain atlas</em>
```
prioux, line 228: Your parser doesn't seem to support comments on the same line as a spec... ? Either adjust the parser or remove the suggestion here.

prioux (changes requested):
Minor changes.

BrainPortal/app/views/tool_configs/_form_fields.html.erb
MontrealSergiy: That's only comment I am not agree. Same line comments are supported. Worked on my tests. But provide your examples, if you feel some corner case is missed.

BrainPortal/app/models/tool_config.rb
MontrealSergiy: ok, btw `presence`   seems to be redundant, as `blank?` is used in the following statement.

BrainPortal/app/models/tool_config.rb
MontrealSergiy: ok

BrainPortal/app/models/tool_config.rb
MontrealSergiy: ok

BrainPortal/app/views/tool_configs/_form_fields.html.erb
prioux: Okay, I didn't see the full effect of the split() in parsed_overlay_specs()

PR #1190, opened 2021-12-15
Add Boutiques integration for value-disables

natacha-beck:


prioux:
There is a known edge-case limitation about A disabling B disabling C or something, Natacha will create an issue about this.

PR #1199, opened 2022-01-25
Added Boutiques modules for output types

prioux:
See the documentation at the top of boutiques_output_file_type_setter.rb

Also, the NewMultiBoutiquesDemo task was upgraded to include an
example of the feature.


natacha-beck (approved):
Works fine.

PR #1200, opened 2022-01-25
BUGFIX blocking requests with offline bourreau fixes #1195

### LINKED ISSUE #1195: Prevent requests from creating tasks on bourreaux that are 'offline'
The user interface prevents this by not making the bourreau available in the selection box, but API requests can still POST to `/tasks` and create a task object on a bourreau that is offline.

When a bourreau is offline, such a request should return 403.

MontrealSergiy:
Hi I added 403 if bourreau_id belongs to offline server, or if id of tool_config.bourreux.id is different from bourreau_id. 

This is because bourreau id is overwriten by tool_config.bourreux.id .


prioux:
Please redo the pull request

You're modifying a bunch of unrelated files and make adjustements to white space for no reason.

BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -381,6 +382,15 @@ def create #:nodoc:
       @task.errors[:results_data_provider_id] = 'must be provided'
     end
 
+    if ! @task.bourreau.nil? && ! @task.bourreau.online? # if api mode bourreau is not automatically adjusted
```
prioux, line 385: This is too complicated. Just ad an error and let the rest of the code naturally handle it. No need to add a respons block here when there is already another one.

```ruby
if ! @task.tool_config.bourreau.online?
  @task.errors[:tool_config_id] = 'is on an executions erver that is currently offline'
end
```

prioux: Yes, that means we get :unprocessable_entity instead of :forbidden but I'm ok with that.

MontrealSergiy:
> Please redo the pull request
> 
> You're modifying a bunch of unrelated files and make adjustements to white space for no reason.

@prioux Hi, which bunch of unrelated files you think unrelated? I added a test because the controller has convoluted bourreau id logic the current api testing framework requires adding/changing several files per test . 

Do you suggest to drop the check of discrepancy in bourreau_id and tool_config.bourraux.id or can I keep it?

MontrealSergiy:
well I pushed next version

BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -24,7 +24,7 @@
 class TasksController < ApplicationController
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
-
+  require 'pry'
```
prioux, line 27: Once again I ask you, can you PLEASE REVIEW YOUR OWN CODE before you commit?

The git utilities in your work computer will show you all changes before you commit locally (use `git diff` and `git diff --staged`), and on top of that, when you make a pull request, the GitHub interface will show you all  lines changed in all files before you actually make the PR. Yes you keep pushing code that changes white space or you leave debug code in your commits. That's unacceptable.

MontrealSergiy: yes, sorry forgot to read it. Debug statement is removed now

prioux:
Right so this looks fine now.

The entire fix was only 3 lines, and they are basically the three lines that I wrote while doing your review.

Thanks for adding a test, however, that is always welcome. As an aside, if you need to add new objects to the test seed, give them ID that are greater than the objects that were already seeded. You added a ToolConfig 19 before the existing  ToolConfig 20. If you had created it as 21, after 20, it would have simplified your life as less changes would have been required in the rest of the tests.

PR #1204, opened 2022-02-21
store and validate task archive file md5 hash to detect corrupting or tampering resolves #1186

### LINKED ISSUE #1186: Store and validate checksum for task archives
CBRAIN has the ability to tar up and task's work directory. The results is a TarWorkdirArchive file registered in the task's owner's file space. 

To prevent tampering with the archive, when the archive is created, a checksum of the tar archive should be computed and stored in the meta data of the task. When unarchiving is requested, the unarchiving code should validate that the tar archive has the same checksum as when it was created.

I suggest storing the checksum in the meta data of the task, and using the 'cksum' program to compute it.

MontrealSergiy:
saves md5 hash on task archive file, blocks unarchiving if archive file is tampered or corrupted. md5 retrieval is lightly modified copy of the existing cbrain md5 code so I assume it works on mac

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1663,6 +1663,7 @@ class ClusterTask < CbrainTask
     end
 
     self.addlog("Attempting to restore TaskWorkdirArchive.")
+    
     taskarch_userfile.sync_to_cache
 
     md5 = self.meta[:workdir_archive_hash]
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1668,7 +1668,6 @@ class ClusterTask < CbrainTask
     md5 = self.meta[:workdir_archive_hash]
     if md5.present? and md5 != taskarch_userfile.cache_compute_md5sum  #  presence check is compatibility layer for old tasks archives
       self.addlog("MD5 hash does not match stored record: TaskWorkdirArchive #{self.name} is corrupted.")
-      # cb_error "Archive of #{self.name} task was tampered with! Unarchiving is aborted"
       return false
     end
 
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1663,12 +1663,12 @@ class ClusterTask < CbrainTask
     end
 
     self.addlog("Attempting to restore TaskWorkdirArchive.")
-    
+
     taskarch_userfile.sync_to_cache
 
     md5 = self.meta[:workdir_archive_hash]
     if md5.present? and md5 != taskarch_userfile.cache_compute_md5sum  #  presence check is compatibility layer for old tasks archives
-      self.addlog("MD5 hash does not match stored record: archive #{taskarch_userfile.name} of task #{self.name} is corrupted.")
+      self.addlog("MD5 hash does not match stored record: task archive #{taskarch_userfile.name} is corrupted.")
       return false
     end
 
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1621,6 +1621,7 @@ class ClusterTask < CbrainTask
       file.save!
       file.cache_copy_from_local_file(tar_file)
       file.cache_erase
+      file.get_hash
       file.save
       self.workdir_archive_userfile_id = file.id
       self.addlog_to_userfiles_created(file)
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1667,7 +1667,7 @@ class ClusterTask < CbrainTask
 
     md5 = self.meta[:workdir_archive_hash]
     if md5.present? and md5 != taskarch_userfile.cache_compute_md5sum  #  presence check is compatibility layer for old tasks archives
-      self.addlog("MD5 hash does not match stored record: TaskWorkdirArchive #{self.name} is corrupted.")
+      self.addlog("MD5 hash does not match stored record: archive #{taskarch_userfile.name} of task #{self.name} is corrupted.")
       return false
     end
 
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1661,6 +1661,8 @@ class ClusterTask < CbrainTask
       return false
     end
 
+    taskarch_userfile.map!(&:check_hash) # validate
+
     self.addlog("Attempting to restore TaskWorkdirArchive.")
 
     taskarch_userfile.sync_to_cache
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1664,7 +1664,10 @@ class ClusterTask < CbrainTask
 
     self.addlog("Attempting to restore TaskWorkdirArchive.")
 
-    taskarch_userfile&.check_hash # validate, save navigation (ampersand) is for smother migration
+    if taskarch_userfile&.integrity_violated?
+      self.addlog("MD5 hash does not match stored record: TaskWorkdirArchive #{self.name} is corrupted.")
+      cb_error "Task archive #{self.name} was tampered with! Unarchiving is aborted"
+    end
     taskarch_userfile.sync_to_cache # for compatibility remove in the next deploy
 
     self.make_cluster_workdir
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -1620,7 +1620,7 @@ class ClusterTask < CbrainTask
                  )
       file.save!
       file.cache_copy_from_local_file(tar_file)
-      self.meta[:workdir_archive_hash] = cache_compute_md5sum.presence
+      self.meta[:workdir_archive_hash] = file.cache_compute_md5sum.presence
       file.cache_erase
       file.save
       self.workdir_archive_userfile_id = file.id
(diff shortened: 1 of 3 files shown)


BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/task_workdir_archive/task_workdir_archive.rb
```diff
@@ -52,4 +54,45 @@ def before_destroy_adjust_task #:nodoc:
     return true
   end
 
+  # compute archive file hash to avoid tampering
```
prioux, line 57: While we're implementing a checksum method, move it to the SingleFile class so that all models can benefit.

Also:

* call it `cache_compute_md5sum` so we respect the conventions for method that accesses the cached content
* do NOT make it call `sync_to_cache()` it isn't this method's responsibility.
* let it raise an exception on any error, do not invoke 'addlog()', this would be poluting the log of the file
* remove the life cycle call back in task_workdir_archive (see other comment below)

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1661,9 +1661,14 @@ def unarchive_work_directory_from_userfile
       return false
     end
 
-    self.addlog("Attempting to restore TaskWorkdirArchive.")
+    if taskarch_userfile&.integrity_violated?
```
prioux, line 1664: invoke the method to compute the md5 and store it in the meta data of the task object, not the TarArchive, then compare at the moment of unarchiving

This context here already know about the status of the cache, so there is no need to invoke sync_to_cache at any time.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/task_workdir_archive/task_workdir_archive.rb
```diff
@@ -52,4 +54,45 @@ def before_destroy_adjust_task #:nodoc:
...
+      else
+        md5 = Regexp.last_match[1] if  md5.match(/\b([0-9a-fA-F]{32})\b/)
+        self.addlog('md5sum is malformed, the output format of md5 system untility may changed') if ! md5
+      end
+      md5
+    end
+  end
+
+  # saves archive file hash in his 'meta' data
+  def save_hash
```
prioux, line 84: remove

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/task_workdir_archive/task_workdir_archive.rb
```diff
@@ -52,4 +54,45 @@ def before_destroy_adjust_task #:nodoc:
...
+    return if self.meta[:content_hash]
+    self.meta[:content_hash] = get_hash.presence
+  end
+
+  # validate crypto hash
+  def integrity_violated?
```
prioux, line 90: remove

prioux (changes requested):
Refactor please. This is too complex.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/task_workdir_archive/task_workdir_archive.rb
MontrealSergiy: ok

BrainPortal/app/models/cluster_task.rb
MontrealSergiy: ok

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/task_workdir_archive/task_workdir_archive.rb
MontrealSergiy: ok

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/task_workdir_archive/task_workdir_archive.rb
MontrealSergiy: ok

MontrealSergiy:
might miss something still have to double check,

MontrealSergiy:
Not 100% sure is it find to use add_log in cluster task (rather than cbrain_error)

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
```diff
@@ -80,4 +80,25 @@ def gzip_content(operation = :compress) #:nodoc:
...
+          "md5sum"
+        when /Darwin/i
+          "md5"
+        else
+          "md5sum"
+        end
+    self.sync_to_cache
+
+    content_path = self&.cache_full_path
+    cb_error("Can't find cache path for file #{self.name}") if content_path.blank?
```
prioux, line 97: This cb_error is unncessary, the cache_full_path is guaranteed to be returned.

It also means that the `&.` at the previous line is also unnecessary

BrainPortal/app/models/cluster_task.rb
```diff
@@ -1662,9 +1663,14 @@ def unarchive_work_directory_from_userfile
     end
 
     self.addlog("Attempting to restore TaskWorkdirArchive.")
-
     taskarch_userfile.sync_to_cache
 
+    md5 = self.meta[:workdir_archive_hash]
+    if md5.present? and md5 != taskarch_userfile.cache_compute_md5sum  #  presence check is compatibility layer for old tasks archives
+      self.addlog("MD5 hash does not match stored record: TaskWorkdirArchive #{self.name} is corrupted.")
```
prioux, line 1670: In the message, you use `self.name` (the task's name) instead of `taskarch_userfile.name` (the archive's name)

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
```diff
@@ -80,4 +80,25 @@ def gzip_content(operation = :compress) #:nodoc:
...
+    md5 = IO.popen("#{md5command} < #{content_path.to_s.bash_escape}","r") { |fh| fh.read }
```
prioux, line 98: The computation code has a flaw. if the command's output is exactly 32 characters and not a MD5, the method will return those 32 characters:

* md5 is the raw output of the command
* it is present? but doesn't match the regex, so the assignment from last_match is not executed and md5 retains its original content
* thecondition on the cb_error thinks all is ok

prioux: Solution: use a different variable name for the command's output VS the md5 extracted from the match

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
```diff
@@ -80,4 +80,25 @@ def gzip_content(operation = :compress) #:nodoc:
...
+    md5command =
+        case CBRAIN::System_Uname
+        when /Linux/i
+          "md5sum"
+        when /Darwin/i
+          "md5"
+        else
+          "md5sum"
+        end
+    self.sync_to_cache
```
prioux, line 94: Do NOT do the sync to cache. Expect the programmer to have done it already. That's not within the scope of the method.

prioux: Another reason for this is that a programmer might want to compute the md5 of something that he has prepared in the cache but has not yet been pushed to the provider side. If you do a sync to cache, you destroy the cache.

prioux (changes requested):
It's almost there. See comments.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
```diff
@@ -80,4 +80,25 @@ def gzip_content(operation = :compress) #:nodoc:
...
+
+    content_path = self&.cache_full_path
+    cb_error("Can't find cache path for file #{self.name}") if content_path.blank?
+    md5 = IO.popen("#{md5command} < #{content_path.to_s.bash_escape}","r") { |fh| fh.read }
+    md5 = Regexp.last_match[1] if md5.present? && md5.match(/\b([0-9a-fA-F]{32})\b/)
+    cb_error("Can't compute MD5 for file #{self.name}", Errno::EIO) if md5.blank? ||
```
prioux, line 100: Also I'm not sure why you wrote that, but cb_error does NOT take a second argument like that. I tried it in my console:

```
CBRAIN> cb_error("Oh no", Errno::EIO)
NoMethodError: undefined method `[]' for Errno::EIO:Class
```

So this means you never exercised that code.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
MontrealSergiy: done

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
MontrealSergiy: done

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
MontrealSergiy: done

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
MontrealSergiy: I thought it depend on dp implementation, but removed

prioux:
I will test this now, but I think the code is fine.

prioux:
Works as expected.

PR #1206, opened 2022-02-28
Cookie notification, crude version for review

prioux:
This is a draft. Please have a look at the way I implemented this and make suggestions about layout and phrasing.

Also, let me know if you think it would work better with `window.sessionStorage` instead of 'window.localStorage`. An explanation of the difference between the two is here:

https://blog.logrocket.com/localstorage-javascript-complete-guide/#sessionstoragevslocalstorage

We also need to implement something similar in the NeuroHub interface side.

### DIFF
--- a/BrainPortal/app/assets/stylesheets/neurohub.scss.erb
+++ b/BrainPortal/app/assets/stylesheets/neurohub.scss.erb
@@ -2107,7 +2107,7 @@ Secondary navbar with application-wide search + breadcrumbs
   width: 100%;
   color: $PRIMARY_DK;
   background-color: $PRIMARY_BG;
-  opacity: 100%;
+  opacity: 95%;
   z-index: 10;
   line-height: 1;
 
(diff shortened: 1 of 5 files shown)


prioux:
Lines 36, 37 and 54 are debug code that will not be part of the final code, but allows you to reset the status of the 'click I agree' action.

BrainPortal/app/assets/stylesheets/neurohub.scss.erb
```diff
@@ -2092,6 +2093,44 @@ Secondary navbar with application-wide search + breadcrumbs
   }
 }
 
+/*
+=============================================================
+| COOKIE NOTIFICATION
```
MontrealSergiy, line 2098: what the difference from COOKIE NOTIFICATION and Cookie Notification Panel should not both have same name?

prioux: It's just a label for a comment section, see at the very top how Candice labels these sections.

PR #1208, opened 2022-03-02
[FIX] Userfiles error display

natacha-beck:


BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -1567,107 +1573,116 @@ def permission_check #:nodoc:
   #+attributes+ is a hash of attributes for all the files,
   #they must contain at least user_id and data_provider_id
   def extract_from_archive(archive_file_name, file_type = nil, attributes = {}) #:nodoc:
+    begin
```
prioux, line 1576: If you are trying to catch exception in the whole method, like ti seems ot be the case here, you don't need that pair of "begin" "end". You are doing:

```
def name
  begin
    all method code
  rescue
    code
  end
end
```

when you could be doing instead

```
def name
  all method code
rescue
  code
end
```

prioux (changes requested):
Small adjustments

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -1670,7 +1676,13 @@ def extract_from_archive(archive_file_name, file_type = nil, attributes = {}) #:
         :variable_text  => report
       )
     end
-
+    rescue => e
+      Message.send_message(current_user.own_group,
```
prioux, line 1680: At line 615 we have `send_message(current_user...` instead of `current_user.own_group` 

Which one is correct?

prioux (commented):
See comments

PR #1213, opened 2022-03-04
Added a auth_method 'keyboard-interactive' to SFTP

prioux:
Some SSHDs really want keyboard-interactive as an auth method,
probably because of 2FA, even if here in CBRAIN we'll never use
that. Hopefully this change won't block our servers when
sftp sessions are established (I expect not, as ssh should
realize no TTYs are available)

### DIFF
--- a/BrainPortal/app/models/en_cbrain_ssh_data_provider.rb
+++ b/BrainPortal/app/models/en_cbrain_ssh_data_provider.rb
@@ -87,7 +87,10 @@ class EnCbrainSshDataProvider < SshDataProvider
 
     # We should create a nice state machine for the remote rename operations
     self.master # triggers unlocking the agent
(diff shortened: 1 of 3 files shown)


prioux:
Test this by connecting to a browsable data provider in your development environment, and browsing the files, and syncing them back and forth. Check the shell where you started your rails server, maybe you'll see messages asking for a password but it is just echoed and never really asked for. Hopefully.

If you want to test deeper, comment-out your SSH key on the remote end, and see what happens when CBRAIN tries to browse/sync etc. All I want is to make sure that the server doesn't go into any mode where it REALLY asks for a password. Timeouts and crashes are all OK!

natacha-beck (approved):

natacha-beck:
I tested locally, works fine.

MontrealSergiy:
seems to work, but at some point I lost ability to edit file
vim gives me
'/data1/user-data/sboroday/dpsshflat/tstfile" E212: Can't open file for writing'

PR #1214, opened 2022-03-09
Boutiques Input Cache Cleaner integrator module  resolves #1203

### LINKED ISSUE #1203: Create new Boutiques module 'BoutiquesInputCacheCleaner'
This module would invoke 'cache_erase' on input files, once a task is complete, using a new specification in the descriptor:

```json
"custom": {
  "cbrain:integrator_modules": {
    "BoutiquesInputCacheCleaner": [
      "inputid1",
      "inputid2"
    ]
  }
}
```

That way when processing large datasets in parallel where the dataset is split in chunks (e.g. BidsSubjects of the UKBB), we can erase the subjec't data from the cache and free some disk space.

MontrealSergiy:
Hi, this is to expire input file cache as per #1230 . On my tests cache seems disappear, how reliable the current conflict prevention I am less than sure.

### DIFF
--- a/BrainPortal/lib/boutiques_input_cache_cleaner.rb
+++ b/BrainPortal/lib/boutiques_input_cache_cleaner.rb
@@ -68,7 +68,7 @@ module BoutiquesInputCacheCleaner
         last_cache_access_time = inputfile.local_sync_status.accessed_at
         next if last_cache_access_time > setup_time  # skip, perhaps the file is being accessed by another task
         inputfile.cache_erase
-        self.addlog("Boutiques Integrator's InputCacheCleaner deleted #{inputfile.name} file cache (#{input.cb_invoke_name})")
+        self.addlog("BoutiquesInputCacheCleaner deleted #{inputfile.name} file cache (#{input.cb_invoke_name})")
       end
 
     end
--- a/BrainPortal/lib/boutiques_input_cache_cleaner.rb
+++ b/BrainPortal/lib/boutiques_input_cache_cleaner.rb
@@ -51,9 +51,10 @@ module BoutiquesInputCacheCleaner
 
   def save_results #:nodoc:
 
-    return fail unless super # call all the normal code
+    return false unless super # call all the normal code
 
-    setup_time = self.meta[:setup_time] || Time.now
+    setup_time = self.meta[:setup_time]
+    return true if setup_time.nil?
 
     descriptor = self.descriptor_for_save_results
     inputs     = descriptor.custom_module_info('BoutiquesInputCacheCleaner')
@@ -64,8 +65,7 @@ module BoutiquesInputCacheCleaner
 
       Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
         inputfile = Userfile.find(inputfileid)
-        next if inputfile.local_sync_status.nil? # cache is already deleted  # cache is already deleted
-        last_cache_access_time = inputfile.local_sync_status.accessed_at
+        last_cache_access_time = inputfile.local_sync_status&.accessed_at
         next if last_cache_access_time > setup_time  # skip, perhaps the file is being accessed by another task
         inputfile.cache_erase
         self.addlog("BoutiquesInputCacheCleaner deleted #{inputfile.name} file cache (#{input.cb_invoke_name})")
--- a/BrainPortal/lib/boutiques_input_cache_cleaner.rb
+++ b/BrainPortal/lib/boutiques_input_cache_cleaner.rb
@@ -1,0 +1,77 @@
+
+#
+# CBRAIN Project
+#
+# Copyright (C) 2008-2021
+# The Royal Institution for the Advancement of Learning
+# McGill University
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# This module adds automatic verification of the
+# type of files selected in a File input of a Boutiques Task.
+#
+# To include the module automatically at boot time
+# in a task integrated by Boutiques, add a new entry
+# in the 'custom' section of the descriptor, like this:
+#
+#   "custom": {
+#       "cbrain:integrator_modules": {
+#           "BoutiquesInputCacheCleaner": [
+#             "my_input1",
+#             "my_input2"
+#           ]
+#       }
+#   }
+#
+# In the example above, any userfile cache selected for the file input
+# named 'my_input1' or 'my_input2' will be deleted after task execution unless CBRAIN
+# notices that another task uses same cache.
+# CBRAIN tries to handle conflict between tasks that share same file based on timestamps.
+# Yet this is a dangerous feature as it might not handle some of races, e.g. with legacy tools
+module BoutiquesInputCacheCleaner
+
+  def setup #:nodoc:
+    self.meta[:setup_time] = Time.now
+    super
+  end
+
+  def save_results #:nodoc:
+
+    return fail unless super # call all the normal code
+
+    setup_time = self.meta[:setup_time] || Time.now
+
+    descriptor = self.descriptor_for_save_results
+    inputs     = descriptor.custom_module_info('BoutiquesInputCacheCleaner')
+
+    inputs.each do |inputid| # 'my_input1', 'my_input2'
+
+      input = descriptor.input_by_id(inputid)
+
+      Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
+        inputfile = Userfile.find(inputfileid)
+        last_cache_access_time = inputfile.local_sync_status&.accessed_at        
+        next if !last_cache_access_time  # cache is already deleted
+        next if last_cache_access_time > setup_time  # skip, perhaps the file is being accessed by another task
+        inputfile.cache_erase
+        self.addlog("Boutiques Integrator's InputCacheCleaner deleted #{inputfile.name} file cache (parameter #{input.cb_invoke_name})")
+      end
+
+    end
+
+  end
+
+end
--- a/BrainPortal/lib/boutiques_input_cache_cleaner.rb
+++ b/BrainPortal/lib/boutiques_input_cache_cleaner.rb
@@ -66,6 +66,7 @@ module BoutiquesInputCacheCleaner
       Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
         inputfile = Userfile.find(inputfileid)
         last_cache_access_time = inputfile.local_sync_status&.accessed_at
+        next unless last_cache_access_time  # skip, already unsynch 
         next if last_cache_access_time > setup_time  # skip, perhaps the file is being accessed by another task
         inputfile.cache_erase
         self.addlog("BoutiquesInputCacheCleaner deleted #{inputfile.name} file cache (#{input.cb_invoke_name})")
--- a/BrainPortal/lib/boutiques_input_cache_cleaner.rb
+++ b/BrainPortal/lib/boutiques_input_cache_cleaner.rb
@@ -44,8 +44,9 @@
 module BoutiquesInputCacheCleaner
 
   def setup #:nodoc:
+    result = super
     self.meta[:setup_time] = Time.now
-    super
+    result
   end
 
   def save_results #:nodoc:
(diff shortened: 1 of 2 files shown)


BrainPortal/app/models/cluster_task.rb
```diff
@@ -747,6 +747,7 @@ def setup_and_submit_job
             self.status_transition(self.status, "Failed To Setup")
           else
             self.addlog("Setup and submit process successful.")
+            self.meta[:setup_time] = Time.now
```
prioux, line 750: You are changing the main CBRAIN code base. It is not proper to modify framework-level code that way.

The point of having special extensions and modules is that they provide their functionality INDEPENDENTLY from the rest of the framework, without requiring changes in that framework. The recording of the :setup_time must be made by code in the module.

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,65 @@
...
+    descriptor = self.descriptor_for_save_results
+    inputs     = descriptor.custom_module_info('BoutiquesInputCacheCleaner')
+    inputs.each do |inputid| # 'myinput1', 'myinput2' ]
+      input = descriptor.input_by_id(inputid)
+
+      next if !input
+      Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
+        inputfile = Userfile.find(inputfileid)
+        last_cache_access_time = inputfile.local_sync_status&.accessed_at || Time.now
+        if last_cache_access_time <= (self.meta[:setup_time] || Time.now) # an input cache was accessed before full setup is finished
```
prioux, line 57: extract the values of setup_time a the beginning of the method, since it's independent of the tw levels of loop you're going through

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,65 @@
...
+module BoutiquesInputCacheCleaner
+
+  def save_results #:nodoc:
+    super # call all the normal code
+    descriptor = self.descriptor_for_save_results
+    inputs     = descriptor.custom_module_info('BoutiquesInputCacheCleaner')
+    inputs.each do |inputid| # 'myinput1', 'myinput2' ]
+      input = descriptor.input_by_id(inputid)
+
+      next if !input
```
prioux, line 53: you don't need to do that check, the method input_by_id will always work

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,65 @@
...
+# notices that another task uses same cache.
+# CBRAIN tries to handle conflict between tasks that share same file based on timestamps.
+# Yet this is a dangerous feature as it might not handle some of races 
+module BoutiquesInputCacheCleaner
+
+  def save_results #:nodoc:
+    super # call all the normal code
+    descriptor = self.descriptor_for_save_results
+    inputs     = descriptor.custom_module_info('BoutiquesInputCacheCleaner')
+    inputs.each do |inputid| # 'myinput1', 'myinput2' ]
```
prioux, line 50: why is there a closing square bracket all alone in your comment?

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      next if !input
+      Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
+        inputfile = Userfile.find(inputfileid)
+        last_cache_access_time = inputfile.local_sync_status&.accessed_at || Time.now
+        if last_cache_access_time <= (self.meta[:setup_time] || Time.now) # an input cache was accessed before full setup is finished
+          inputfile.cache_erase
+        else
+          self.addlog("cannot delete #{inputfile.name} file cache (parameter #{input.cb_invoke_name}), due to a contention with another task")
+        end
```
prioux, line 61: Add a bit of clarity spacing so that you don't end the code with a ladder of 4 `end` like that?

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,65 @@
...
+      input = descriptor.input_by_id(inputid)
+
+      next if !input
+      Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
+        inputfile = Userfile.find(inputfileid)
+        last_cache_access_time = inputfile.local_sync_status&.accessed_at || Time.now
+        if last_cache_access_time <= (self.meta[:setup_time] || Time.now) # an input cache was accessed before full setup is finished
+          inputfile.cache_erase
+        else
+          self.addlog("cannot delete #{inputfile.name} file cache (parameter #{input.cb_invoke_name}), due to a contention with another task")
```
prioux, line 60: Only add a comment to the task log when you DO erase the input cache, and say nothing when you don't.

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,65 @@
...
+#
+# In the example above, any userfile cache selected for the file input
+# named 'my_input1' or 'my_input2' will be deleted after task execution unless CBRAIN
+# notices that another task uses same cache.
+# CBRAIN tries to handle conflict between tasks that share same file based on timestamps.
+# Yet this is a dangerous feature as it might not handle some of races 
+module BoutiquesInputCacheCleaner
+
+  def save_results #:nodoc:
+    super # call all the normal code
```
prioux, line 47: This is wrong. The save_results method's specification says that it can return true or false or raise an expection depending on different situations. You completely ignore the true/false values here. This line shoudl be

```ruby
return false unless super
```

(Do you see why?)

prioux (changes requested):
Minor changes and some bug fixes

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,65 @@
...
+    descriptor = self.descriptor_for_save_results
+    inputs     = descriptor.custom_module_info('BoutiquesInputCacheCleaner')
+    inputs.each do |inputid| # 'myinput1', 'myinput2' ]
+      input = descriptor.input_by_id(inputid)
+
+      next if !input
+      Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
+        inputfile = Userfile.find(inputfileid)
+        last_cache_access_time = inputfile.local_sync_status&.accessed_at || Time.now
```
prioux, line 56: Also, of local_sunc_status returns nil, just skip to the next entry in the loop, it means the file's no longer synchronized.

BrainPortal/app/models/cluster_task.rb
MontrealSergiy: ok, so you do not care of same file being used by an old style tool and this new boutiques integrator based tool?

Extensions is not even in plugin folder.

prioux (changes requested):
Fix and rebase please.

BrainPortal/app/models/cluster_task.rb
MontrealSergiy: What make no sense? Do you think a file used by old type of task cannot be used by a new type of boutiques task? Or perhaps this even is improbable... ok than

I understand that's easy to override just thought the 'good design' solution is  inferior.

prioux: It makes no sense to record the timestamp for the setup() method for tasks that are not integrated with this module.

THis timestamps is not associated with any input files. It is associated with the task's lifecycle.

prioux: The point is that this feature here is meant to be configured for tasks that use some inputs file while at the same task we expect no other tasks to use those same input files. And if any task is launched AFTER the current task and synchronize an input we use in the current task, then we do not delete the input. That's it.

prioux:
Please just do the basic implementation like it was in your original PR of yesterday, but with the fixes I asked for.

For a foolproof implemetation, I know how to do that I will code it myself by modifying your code.

prioux:
There is only about 10 minutes of work left to do, so that should be quick.

BrainPortal/lib/boutiques_input_cache_cleaner.rb
prioux: But that's what we want absolutely! We don't want to erase the input cache if ANYTHING goes wrong, so as to provide a recovery path. We only want to erase the cache when the task completes successfully.

BrainPortal/lib/boutiques_input_cache_cleaner.rb
MontrealSergiy: done

BrainPortal/lib/boutiques_input_cache_cleaner.rb
MontrealSergiy: done

BrainPortal/lib/boutiques_input_cache_cleaner.rb
MontrealSergiy: removed

BrainPortal/lib/boutiques_input_cache_cleaner.rb
MontrealSergiy: tried my best to unladder

BrainPortal/lib/boutiques_input_cache_cleaner.rb
MontrealSergiy: done

BrainPortal/lib/boutiques_input_cache_cleaner.rb
MontrealSergiy: done

BrainPortal/lib/boutiques_input_cache_cleaner.rb
MontrealSergiy: done

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def setup #:nodoc:
+    result = super
+    self.meta[:setup_time] = Time.now
+    result
+  end
+
+  def save_results #:nodoc:
+
+    return fail unless super # call all the normal code
```
prioux, line 54: What's `fail` ?!?

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    setup_time = self.meta[:setup_time] || Time.now
```
prioux, line 56: That's wrong, if you don't have a setup time, just don't do anything and return instead. There is nothing to do without it!

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,78 @@
...
+    descriptor = self.descriptor_for_save_results
+    inputs     = descriptor.custom_module_info('BoutiquesInputCacheCleaner')
+
+    inputs.each do |inputid| # 'my_input1', 'my_input2'
+
+      input = descriptor.input_by_id(inputid)
+
+      Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
+        inputfile = Userfile.find(inputfileid)
+        next if inputfile.local_sync_status.nil? # cache is already deleted  # cache is already deleted
```
prioux, line 67: why the double comment?

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,78 @@
...
+        last_cache_access_time = inputfile.local_sync_status.accessed_at
```
prioux, line 68: In line 67 and 68 you invoke `local_sync_status` twice, once to check it it's nil and once to get the access time, so loading it from the database twice too.

How about instead:

```
last_cache_access_time = inputfile.local_sync_status&.accessed_at
next if ! last_cache_access_time  # or unless if you prefer
```

prioux (changes requested):
Minor changes

BrainPortal/lib/boutiques_input_cache_cleaner.rb
prioux: Won't that raise an exception?

MontrealSergiy:
Note that multi inputs, like in MultiBoutiques demo (see below) not expand cbsvn list, BoutiquesInputCacheCleaner deleted list2.cbsv file cache (invoke[minput1][]), which is usually small text file. https://gist.github.com/MontrealSergiy/5cc364c9bbe6d7b729543bd2f61fdf6c

prioux (changes requested):
Bugs

BrainPortal/lib/boutiques_input_cache_cleaner.rb
MontrealSergiy: ok, I thought it is cached

prioux:
The code seems all correct now. I will test it.

BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      input = descriptor.input_by_id(inputid)
+
+      Array(invoke_params[inputid]).map(&:presence).compact.each do |inputfileid|
+        inputfile = Userfile.find(inputfileid)
+        last_cache_access_time = inputfile.local_sync_status&.accessed_at
+        next unless last_cache_access_time  # skip, already unsynch 
+        next if last_cache_access_time > setup_time  # skip, perhaps the file is being accessed by another task
+        inputfile.cache_erase
+        self.addlog("BoutiquesInputCacheCleaner deleted #{inputfile.name} file cache (#{input.cb_invoke_name})")
```
prioux, line 72: Why do you insert `input.cb_invoke_name` in the log message? It's ugly and useless for users and adminstrators.

prioux: Don't fix this, I will fix it when I make my own adjustments later.

prioux (approved):
OK will merge

PR #1215, opened 2022-03-10
[REFACTORING] of size condition for has_viewer

natacha-beck:


prioux:
If you want to be fancy and do some meta programming, you could add a method_missing handler to single file so that in subclasses, someone could do

```
has_viewer blah blah, :if => :size_allows_viewing_400mb?
has_viewer blah blah, :if => :size_allows_viewing_20kb?
has_viewer blah blah, :if => :size_allows_viewing_2000?
```

or any other method name with that pattern, and dynamically the method woudl be created and it would invoke `size_allows_viewing(number)`. E.g it would dynamically create

```
def size_is_at_least_400mb?
    size_allows_viewing(400_000_000)
end
```

it's a good exercise in meta programming. Supporting the suffixes (kb, mb etc) is optional, it could all be bytes.

prioux (changes requested):
Review

prioux:
Well, it's not meta programming, you're not defining the methods dynamically, you're just running the code in method_missing....

Also, the if/else structure is too complicated.

natacha-beck:
I still define the method on the fly.

I can simplify the `if else`: 

```ruby
  def method_missing(name, *args)
    # Define a method that perform size verifictaion on the file
    # in order to know if it's viewable or not
    size = name.to_s.match(/^size_allows_viewing_(.+)/)[1]
    return super if !size

    size_limit      = size.to_i
    userfile_errors = []
    if self.size.blank?
      userfile_errors.push("No size available for this file")  
    else
      userfile_errors.push("File is too large to be viewable (> #{size.humanize} kB)") if self.size > size_limit
    end

    rescue
      super
  end
```

prioux:
See my example. You are not defining any method, you are running the code check code within method_missing itself. That's not meat programming.

prioux:
Actually I just realized that in my example, `method_name` is the same as `name`

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/csv_file/csv_file.rb
```diff
@@ -27,7 +27,7 @@ class CSVFile < TextFile
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
 
-  has_viewer :name => 'CSV Viewer', :partial  => :csv_file, :if => :csv_is_viewable?
+  has_viewer :name => 'CSV Viewer', :partial  => :csv_file, :if => Proc.new { |u| u.size_allows_viewing_400_000 }
```
prioux, line 30: The entire point of the exercise was to simplify the job of the developer when configuring the `if` callback... and you still use a Proc... ? Did you forget?

prioux (changes requested):
See new comments. There are two Proc.new that are not needed anymore

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
```diff
@@ -97,4 +97,27 @@ def cache_compute_md5sum
...
+  
+  # Check if a file respect a size_limit in order to be viewable
+  # this method is used in the :if of the has_viewer used in specific 
+  # Userfile model
+  def size_allows_viewing?(size_limit=400_000)
+    userfile_errors = []
+    if self.size.blank?
+      userfile_errors.push("No size available for this file")  
+    else
+      userfile_errors.push("File is too large to be viewable (> #{size_limit.to_s.humanize} kB) ") if self.size > size_limit
```
prioux, line 118: What does the .humanize do ?

natacha-beck: For example: 

```ruby
NaN main :001 > "400_000".humanize
 => "400 000" 
```

prioux: But this can never happen in your code.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/single_file/single_file.rb
```diff
@@ -97,4 +97,27 @@ def cache_compute_md5sum
...
+  
+  # Check if a file respect a size_limit in order to be viewable
+  # this method is used in the :if of the has_viewer used in specific 
+  # Userfile model
+  def size_allows_viewing?(size_limit=400_000)
+    userfile_errors = []
+    if self.size.blank?
+      userfile_errors.push("No size available for this file")  
+    else
+      userfile_errors.push("File is too large to be viewable (> #{size_limit.to_s} kB) ") if self.size > size_limit
```
prioux, line 118: The message is wrong, it says `size_limit kb` but the limit is in bytes, as implemented.

You have to decide what the meaning of the 400_000 is. Is it bytes or kilobytes?

prioux: You can also just change the message, by replacing `kB` with `bytes`.

prioux: However, if you do that, then when a programmer wants to specify a limit of 5 gigabytes for viewing a file, they'll have to do 

```
has_viewer ...., :if => :size_allows_viewing_5_000_000_000
```

which is quite long and has a lot of zeros.

PR #1216, opened 2022-03-11
Improvements on boutiques input cache cleaner.

prioux:
The cleaning now only occurs if the input file was not already
synchronized at the moment we enter setup().

### DIFF
--- a/BrainPortal/lib/boutiques_input_cache_cleaner.rb
+++ b/BrainPortal/lib/boutiques_input_cache_cleaner.rb
@@ -38,14 +38,22 @@
 #
 # In the example above, the cached content of any userfile(s) selected for the
 # file inputs named 'my_input1' or 'my_input2' will be deleted after the task
-# completes properly. This happens unless CBRAIN notices that another task
-# uses that input too. The module tries to handle conflict between tasks that
-# share same file based on the SyncStatus timestamps.
+# completes properly.
+#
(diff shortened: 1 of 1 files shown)


BrainPortal/lib/boutiques_input_cache_cleaner.rb
```diff
@@ -36,44 +36,72 @@
...
+# In the example above, the cached content of any userfile(s) selected for the
+# file inputs named 'my_input1' or 'my_input2' will be deleted after the task
+# completes properly. This happens unless CBRAIN notices that another task
+# uses that input too. The module tries to handle conflict between tasks that
+# share same file based on the SyncStatus timestamps.
 module BoutiquesInputCacheCleaner
 
   def setup #:nodoc:
+
+    # BEFORE code: identify userfiles that are not synced at all
```
MontrealSergiy, line 48: `# BEFORE standard setup:  identify userfiles that are not synced at all `

would be clearer description. Also the class doc could mention that feature only work for nonsync files.

prioux: I improved the documentation.

MontrealSergiy:
On a few tests seems work as intended - sync files remain sync, async remain async

MontrealSergiy (approved):
The idea to reduce chances of erroneous erase seems valid, chances of erroneous erase are low

PR #1218, opened 2022-03-14
New Boutiques module for cleaning output caches

prioux:
See description in the module.

There are no issues associated with this feature.

### DIFF
--- a/Bourreau/lib/boutiques_output_cache_cleaner.rb
+++ b/Bourreau/lib/boutiques_output_cache_cleaner.rb
@@ -1,0 +1,1 @@
+../../BrainPortal/lib/boutiques_output_cache_cleaner.rb
--- a/BrainPortal/app/models/boutiques_portal_task.rb
+++ b/BrainPortal/app/models/boutiques_portal_task.rb
@@ -62,6 +62,14 @@ class BoutiquesPortalTask < PortalTask
     self.boutiques_descriptor
   end
 
+  # This method returns the same descriptor as
+  # boutiques_descriptor(), by default, but can be overriden
+  # by subclasses to change the behavior of what happens
+  # when generating the task's "show" page.
+  def descriptor_for_show_params
+    self.boutiques_descriptor
+  end
+
 
 
   ##############################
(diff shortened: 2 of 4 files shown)


prioux:
Note for the reviewers @natacha-beck and @MontrealSergiy .

You'll notice that this module has a new type of functionality that we've never
had before:  the module's function can be turned on or off through a new
checkbox in the task's form. This artificial checkbox is added by the module itself.

MontrealSergiy:
I tried to rerun an old task, (the same I used for input cache cleaning integrator). And output file appeared to be async. Is it normal or may new change affected old tasks?

prioux:
Please explain in more details. Did you rerun from scratch or restart an existing task? If restarting, at which stage? Is this with or without the new module directive in the descriptor? What do you mean by 'async' ? Do you mean the cache was erased? What did you see in the logs?

MontrealSergiy:
Did you rerun from scratch or restart an existing task? 
_both_

 If restarting, at which stage?
 _setup_
  Is this with or without the new module directive in the descriptor? 
  _tried only **input** cache cleaner and also both input and output cache cleaner in the same time directives so far, with the same result _
  What do you mean by 'async' ? Do you mean the cache was erased? 
  _Yes_ 
  What did you see in the logs?
  _ input cache deleted_

prioux:
You still have not made it clear to me what you think is unexpected. Is there a bug or no?

prioux:
We know the input file cleaner works, let's keep it out of the discussion. What about the output cache cleaner. Did it clean the output of the cache? If so, whats' the bug?

prioux:
When you check with the console, do you still find a SyncStatus record for the output files ?

MontrealSergiy:
no I do not see SyncStatus record


```CBRAINVagrant main :002 > f.local_sync_status
  SyncStatus Load (0.6ms)  SELECT  `sync_status`.* FROM `sync_status` WHERE `sync_status`.`userfile_id` = 182 AND `sync_status`.`remote_resource_id` = 1 ORDER BY `sync_status`.`id` ASC LIMIT 1
 => nil
```

for purely output cleaning I have some technical issues, so will take some time

prioux:
The method `local_sync_status` returns a differetn record depending on whether you are on a portal or on a bourreau. There's a different obejct to represent the cache status of each, because each has it's own independent cache. A file can be cached on the portal and not on the bourreau, and vice versa, or on both, or on none. To see all the sync status objects associated with a file, just use the plain association:

```
f.sync_status
```

which will return them all as an array. The attribute 'remote_resource_id' indicate which record is associated with which portal cache or bourreau cache.

MontrealSergiy:
oh, bourreau cache is kept or deleted as per checkbox

MontrealSergiy (approved):

PR #1219, opened 2022-03-16
Disk Quota System

prioux:
To start testing this, go to the data provider index page and you'll see a new button 'Disk Quotas'.

Note that although the interface shows the bytes limits in pretty sizes (e.g. 3.2 GB or 6.1 MB etc) the values you must enter in a quota is in `bytes`!

### DIFF
--- a/Bourreau/app/models/disk_quota.rb
+++ b/Bourreau/app/models/disk_quota.rb
@@ -1,0 +1,1 @@
+../../../BrainPortal/app/models/disk_quota.rb
--- a/Bourreau/lib/cbrain_disk_quota_exceeded.rb
+++ b/Bourreau/lib/cbrain_disk_quota_exceeded.rb
@@ -1,0 +1,1 @@
+../../BrainPortal/lib/cbrain_disk_quota_exceeded.rb
--- a/BrainPortal/app/controllers/disk_quotas_controller.rb
+++ b/BrainPortal/app/controllers/disk_quotas_controller.rb
@@ -39,6 +39,7 @@ class DiskQuotasController < ApplicationController
 
     respond_to do |format|
       format.html
+      format.js
     end
   end
 
--- a/BrainPortal/app/controllers/disk_quotas_controller.rb
+++ b/BrainPortal/app/controllers/disk_quotas_controller.rb
@@ -94,8 +94,8 @@ class DiskQuotasController < ApplicationController
     @disk_quota ||= DiskQuota.new(   :user_id => form_user_id, :data_provider_id => form_dp_id )
 
     # Update everything else.
-    @disk_quota.max_bytes = quota_params[:max_bytes].to_i if quota_params[:max_bytes].present?
-    @disk_quota.max_files = quota_params[:max_files].to_i if quota_params[:max_files].present?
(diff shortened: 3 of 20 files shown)


prioux:
Other notes:

This system has no support for quotas based on projects. From what I observed, it's fine this way.

prioux:
Hi @MontrealSergiy, yes you're right this is pretty strict, but this is a`hard quota` system. What you are referring to is a 'soft quota` system, and I think we should implement that separately, maybe with a `rake` task that we run from time to time. I will think about it and create an issue or pull request for the `soft quota` system. I can imagine a rake task scanning all user quotas and creating an internal message when a user is using more than 80% of their quota.

prioux:
I added a 'quotas' tab in the admin interface, and for normal users a 'quotas' menu entry in the 'resources' menu at the very top.

prioux:
Also, I've added coloring code that will turn the top bar 'green' when the currently checked out git branch is not one of the normal ones. I think it's useful for us, developers.

MontrealSergiy:
is limit on ":userfiles" registered files or file collections, or total number of files?

prioux:
Total number of files. As you can see at line 81 of the DiskQuota model.

MontrealSergiy:
in new quota form, if admin forgot to enter dp name (or is it id), the admin is redirected to quota index, which is different from how most form implemented in cbrain (stay on new page and hightlight erroneous field).

prioux:
OK, that seems a bug, I'll fix it.

prioux:
Ths fix was easy and you get nice little red boxes around the missing attributes.

MontrealSergiy:
The default quota on number of files does not seem work. Will try give an individual quota

MontrealSergiy:
I tried to set individual and default quota on file number neither did work: I was able to upload more files than allowed by quota

prioux:
Okay, I can have a look, but note that this quota system triggers only once the quota is already broken, not before. It will allow a process to create a file or directory tree when the quota is OK, even if by doing so the limit gets to be exceeded. Further files will then be prevented from being created.

MontrealSergiy:
For 'default' quota of 2 files I was able to upload 5 files and counting

prioux:
@MontrealSergiy There's something else I forgot to mention. The system caches the usage information for a pair [user, data_provider] for about one minute.

MontrealSergiy:
I waited 1 min with a timer still can upload

prioux:
Can you show me a screenshot of the quotas index table (as seen by the admin) and tell me on which DP you were uploading?

MontrealSergiy:
image

MontrealSergiy:
image

prioux:
Oh never mind, I was looking at the wrong thing

MontrealSergiy:
once again, I am testing 'default quota' for all users, so that's should not matter is it serge or sboroday users

prioux:
@MontrealSergiy Thanks for finding this bug. I pushed a new commit, just pull it and let me know.

MontrealSergiy:
I find it a bit strange to see dropdowns on the new quota page, but since your are the main user for it, I am sure you will not be confused, so up to you

MontrealSergiy (approved):
seems ok

PR #1223, opened 2022-03-24
Data Tracking Framework.

prioux:
This allows an administrator to mark a WorkGroup
with a 'data tracking' flag. When enabled, the CBRAIN
codebase will count, monthly, the files of those
projects as used in four contexts:

1) downloads using the interface
2) views in the browser
3) copies made to other data providers
4) files used to setup a CBRAIN task

For each context, a value of `+1` is added to a `{context}_count`
and a value of `+userfile.num_files` is added to a `{context}_num_files` counter
(so basically we can see the number of files used in file
collections).

The standard report maker of CBRAIN has been extended to add
four new report types for these values.

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -747,6 +747,7 @@ class ClusterTask < CbrainTask
             self.status_transition(self.status, "Failed To Setup")
           else
             self.addlog("Setup and submit process successful.")
+            self.track_usage_of_input_files_after_setup
             # the status is moving forward at its own pace now
           end
           self.meta[:submit_without_setup] = nil # reset special skip
@@ -2174,6 +2175,21 @@ exit $status
     return self.cluster_workdir_size
   rescue
     return nil
+  end
+
+  # This will increase the "task_setup" counts of the
+  # DataUsage records for any and all userfiles left in
+  # the :interface_userfile_ids array of the task's params.
+  def track_usage_of_input_files_after_setup #:nodoc:
+    userfile_ids = self.params[:interface_userfile_ids] || []
+    Userfile.joins(:group)
+      .where('userfiles.id' => userfile_ids)
+      .where('groups.track_usage' => true)
+      .to_a
+      .select { |userfile| userfile.is_locally_synced? }
+      .each do |userfile|
+        DataUsage.increase_task_setups(self.user, userfile)
+      end
   end
 
 
--- a/BrainPortal/app/models/data_usage.rb
+++ b/BrainPortal/app/models/data_usage.rb
@@ -32,7 +32,7 @@ class DataUsage < ApplicationRecord
   belongs_to :user,  :optional => false
   belongs_to :group, :optional => false
 
-  validates_uniqueness_of :user_id, :scope => [ :group_id ]
+  validates_uniqueness_of :user_id, :scope => [ :group_id, :yearmonth ]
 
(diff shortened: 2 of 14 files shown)


prioux:
Missing in my description is the fact that these records are also separate for each user, so we can figure out how many distinct users perform these actions.

prioux:
Instructions for testing.

* As admin, find any work group where you have files in it.
* Open the group's show page and click 'edit'
* There is a new checkbox called 'data tracking' which will turn tracking for files of this group
* At this point, as any users (and with multiple users preferably), do any of:
1. copy the files
2. launch tasks on them
3. view them in the interface
4. download them

In the console, the tracking model is `DataUsage`, so try `tv DataUsage.all` and `pv DataUsage.all`

In the interface, as admin, you will see four new report categories in the report maker (the url is `/report`, and also it's at the top of 'my account')

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2176,6 +2177,21 @@ def update_size_of_cluster_workdir
     return nil
   end
 
+  # This will increase the "task_setup" counts of the
+  # DataUsage records for any and all userfiles left in
+  # the :interface_userfile_ids array of the task's params.
+  def track_usage_of_input_files_after_setup #:nodoc:
```
natacha-beck, line 2183: I think it's possible for a user to select X files and to use only a part of these X files. In this case, should we increase the count for all the files?

BrainPortal/app/models/data_usage.rb
```diff
@@ -0,0 +1,77 @@
...
+  validates_presence_of :user_id
+  validates_presence_of :group_id
+  validates_presence_of :yearmonth
+
+  belongs_to :user,  :optional => false
+  belongs_to :group, :optional => false
+
+  validates_uniqueness_of :user_id, :scope => [ :group_id, :yearmonth ]
+
+  def self.increase_views(user, userfile)
```
natacha-beck, line 37: Should have :nodoc: or documentation no?
Same comment for other methods here.

natacha-beck:
I tested to launch a `diagnostic` task on a file present in a project that is set up to track the data and I end up with no count in the table:
Screen Shot 2022-03-28 at 9 38 36 AM

Not sure if I have made a wrong manipulation or if it is a bug. @MontrealSergiy can you try this one too?

BrainPortal/app/models/data_usage.rb
prioux: Yes, will make the adjustments.

BrainPortal/app/views/groups/show.html.erb
```diff
@@ -84,6 +76,25 @@
...
+    <% if current_user.has_role?(:admin_user) && @group.is_a?(WorkGroup) %>
+
+      <% t.edit_cell("Invisible", :content => check_box_tag(nil ,nil, @group.invisible?, :disabled => true)) do |f| %>
+        <%= f.check_box :invisible %>
+        <div class="field_explanation">
+          If checked, the project will not be shown in the list of projects.
+        </div>
+      <% end %>
+
+      <% t.edit_cell("Track Usage", :content => check_box_tag(nil ,nil, @group.track_usage?, :disabled => true)) do |f| %>
```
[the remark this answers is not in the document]
prioux: Good point, I will fix this.

prioux:
Make sure your bourreau has the branch checked out and that it really has been restarted ?

Issue #1043, opened 2020-10-20
Add editable 'News' section for the dashboard


prioux:
This is for both CBRAIN and NeuroHub. Let the admin customize and record one (or several?) news items in the dashboard.

Not sure where to store the info though. CBRAIN has a built-in `help_button()` helper which stores the text in a file on the filesystem, so maybe use this mechanism with an option such that it shows in page instead as a separate help panel.

MontrealSergiy:
Alternatively could be a message (send by admin or everybody... ) .  
If needed can be varying visibility or individually targeted news.

Bonus is internal links are supported out of box.

If needed is a good idea enable HTML or lightweight markup ( bbcode / markdown ...) support or text is fine?

I have seen cases other roles than admin where allowed to post news. (may be site manager?) Are existing news editable? Is draft mode/preview needed?

prioux:
There are many possibilities, let's talk about it out loud before any code is written.

PR #1228, opened 2022-04-16
Admins can now provide dashboard messages

prioux:
This implements #1043

### DIFF
--- a/BrainPortal/app/assets/stylesheets/cbrain.css.erb
+++ b/BrainPortal/app/assets/stylesheets/cbrain.css.erb
@@ -2283,6 +2283,16 @@ img {
   border-color: blue;
 }
 
+.message_display_table tr.cbrain_dashboard_display {
+  background-color: #f267c2;
+  border-color: black;
+}
+
+.message_display_table tr.neurohub_dashboard_display {
+  background-color: #f267c2;
+  border-color: black;
+}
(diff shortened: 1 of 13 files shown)


prioux:
To test: on CBRAIN side, as admin, go to Messages, click 'Leave Message' then two new buttons will appear at the top.

MontrealSergiy:
you do not want to keep the group dropdown? May be useful if you want to target different groups, like UK Biobank or admins

MontrealSergiy:
CBRAIN dashboard might look ugly, especially for narrow and wide messages. Should align and flow better 
image
image

MontrealSergiy:
wide message : credits should not show here, I think
image
image

Whole whiteboard disappears for wide (long) message
image

prioux:
Right, so there are some weird rendering problems. I'll look into this

prioux:
I think the first screenshot is because you made a mistake. Your HTML was invalid in your message. We can see that everything isn the page is in italics. Please make sure you enter valid HTML

MontrealSergiy:
besides italic on the first screenshot the panel alignment is not ideal.
It would be great to address isolating html errors to the message itself, not sure may be iframe or other container. One can expect admin know html, but occasional typo or accidental enter push are possible.

PR #1230, opened 2022-04-19
[ADD] method to generate fake files list for FileCollection

natacha-beck:
Linked to PR `https://github.com/aces/cbrain-plugins-neuro/pull/234` in `cbrain-plugins-neuro`


prioux:
I'm changing the destination branch too.

PR #1235, opened 2022-05-03
Allows the userfiles page to switch project

prioux:
Direct urls can link to the file manager with a
parameter that implement a direct switch to another
active project. e.g. /userfiles?switch_group_id=123
This is similar to the switch action in the Groups
controller, but it will also some differences:
* The list of persistently selected files will not be cleared
* If the switched project is public, or if it belongs
  to someone else, the 'view_all' files option will be set.

### DIFF
--- a/BrainPortal/app/controllers/groups_controller.rb
+++ b/BrainPortal/app/controllers/groups_controller.rb
@@ -25,6 +25,8 @@ class GroupsController < ApplicationController
 
   Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
 
+  include SwitchGroupHelpers
+
   api_available :only => [:index, :create, :switch, :update, :destroy, :show]
 
   before_action :login_required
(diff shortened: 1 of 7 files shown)


MontrealSergiy:
link did not work for me

prioux:
Point by point:
* yes that's normal
* not it goes to master, it will need to be deployed in production soon
* if you want, it's just details of other things I had on my backlog

MontrealSergiy (approved):
ok then

natacha-beck:
> when non-existing group is entered the user is redirected to the dashboard, without any message

Should it redirect to the Userfiles index page without switching group instead of redirecting to the dashboard?  (@prioux)

prioux:
No, it's fine the way it is. Providing a wrong group_id is an error. It raises a `RecordNotFound` exception, for which we have a standard handler. API requests will get a proper error message, and HTML requests are redirected to the dashboard *by convention*.

PR #1239, opened 2022-05-11
Improved cbrain_remote_ctl

prioux:
The script cbrain_remote_ctl was a Ruby script. Its job was to start or stop a bourreau, or start a bourreau console.

This PR replaces it with a bash script which does basically the same thing. The advantage is that the Ruby script used to require all the Ruby libraries specified in the app's bundle, whereas this bash script is pretty much plain unix shell.

To test this: deploy on a local and remote bourreau of your choice. Starts and stop it using the web interface and the 'ibc' command in your portal's console.

Monitor the content of 'Bourreau/log/server_start.log' to check for errors.

If you get an error that says something like:

```
! Unable to load application: Gem::LoadError: You have already activated nio4r 2.5.8, but your Gemfile requires nio4r 2.5.2. Prepending `bundle exec` to your command may solve this.
```

(it could be any gem, not just `nio4r`) then it means you have more than one version of a gem installed. you can force a clean of unused gems with:

```
cd Bourreau
bundle clean --force
```

but be careful, if your Bourreau is installed at the same place as your portal, both will be affected. The 'Gemfile.lock' on each side has the list of activated gems, make sure any gem in common is specified with the same version. As usual, run "bundle install" on each side to verify everything.

### DIFF
--- a/Bourreau/script/cbrain_remote_ctl
+++ b/Bourreau/script/cbrain_remote_ctl
@@ -1,4 +1,4 @@
-#!/usr/bin/env ruby
+#!/bin/bash
 
 #
 # CBRAIN Project
@@ -24,17 +24,16 @@
 # This script is a wrapper that launches the normal Rails server,
 # but with the added ability of receiving the content of the database.yml
 # file in standard output before it does so. Also, once the server is
-# started, the database.yml is deleted outright, closing a security hole.
+# started, the database.yml is deleted outright.
 #
 # This script it NOT usually executed by a human, instead it is
 # invoked by the Portal when trying to start or stop a Bourreau.
 #
 # The command-line args are to be provided strictly in the order shown
 # below, and all of them should be present (this is done simply so that
-# this bash script is kept simple, as anyway this script is not meant
-# to be started manually by users).
+# this bash script is kept simple).
 #
-#
+# -----------------------
 # To start the Rails app:
 #
 #   $0 start -e <environment>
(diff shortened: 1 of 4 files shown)


MontrealSergiy:
works with local boureau

MontrealSergiy:
Diagnostics task runs well on local (wsl, ruby 2.6) + remote (ace-vh-2 ruby 2.4.1), did not notice any exceptions in the log file Bourreau/log/server_start.log

prioux:
Once the Bourreau is started there really shouldn't be any difference in the operations of tasks. What really needs to be verified is that the Bourreau does start and stop, using 'ibc' or the 'start' or 'stop' buttons in the interface.

Bourreau/script/cbrain_remote_ctl
```diff
@@ -24,113 +24,110 @@
...
 # Values from previous 'start' operation
-prevpid    = File.read(pidfile).to_i rescue 0
-prevhost   = (File.read(pidhost) rescue "").strip
-
-# To help diagnose problems, print Ruby's version.
-puts "You are using Ruby #{RUBY_VERSION}"
-if RUBY_VERSION =~ /^1\.[0-8]/
-  puts "Warning! This Ruby version seems to be way too old. Maybe your environment wasn't initialized properly?"
-end
+prevpid=$(cat $pidfile 2>/dev/null)
```
MontrealSergiy, line 102: not sure what are you doing. Seems to evaluate to empty string

prioux: It's so the statement works when the file doesn't exist and don't print error messages. `2>` is a redirection of channel #2 (stderr), not stdout (which is channel #1).

Why don't you try it on the command line yourself:

```shell
echo 1234 > pidfile
x=$(cat pidfile 2>/dev/null) # with the file existing
echo "-$x-"
rm pidfile
x=$(cat pidfile 2>/dev/null) # with the file not existing
echo "-$x-"
```

MontrealSergiy:
IBC does not seem to work : the bourrea stop request stays in the queue for a long time, but nothing happens. Though I did not use ibc for a long time

prioux:
Try cleaning up manually:

* kill the worker processes (if any)
* kill the bourreau process (if any)
* remove the pidfile
* remove the two socketfiles in tmp/sockets (if any)

Then from `ibc` try start and stop sequences.

MontrealSergiy:
I think I got it, will double check

MontrealSergiy:
ibc seems work now

MontrealSergiy:
ok with me

natacha-beck (approved):

PR #1243, opened 2022-05-31
Improve the `select_all` mechanism

natacha-beck:
This PR should improve the mechanism of the `select_all` mechanism (used in CIVET tools for example). 

Multiple cases: 
- If the `select_all` checkbox is clicked all the attached checkboxes are selected.
- Then if an attached checkbox is unselected the `select_all` checkbox is deselected.
- Then if all the attached checkboxes are selected again the `select_all` is checked again.

In meantime, I add a `hidden` input in order to keep track of the state of the `select_all` that can be `all` or `none`.


MontrealSergiy:
How to test it? Is there description of issues (or #) that this PR addresses?

prioux:
This PR has design issues that were identified last week, Natacha has to work on it some more.

natacha-beck:
I think it is ready to test now. 

It come with the following PR in cbrain-plugins-neuro: 

- https://github.com/aces/cbrain-plugins-neuro/pull/236 

To test it you need for civet_macaque: 
- 2 t1 files. 

And t2 and pd files associated with them: 

-  For example if you have `abc_t1.mnc`, you should have `abc_t2.mnc` and `abc_pd.mnc` 

The ideas is to allow the user to save preset, without the ` Use multispectral classification?` for example. 

Due to limitation it will save as all or none. We cannot save the list of all the files with a status of all the `select_all` checkbox.

natacha-beck:
@MontrealSergiy please test it with Chrome instead of Firefox it will maybe fix your issue. 
I had the same issue on my side, it was due to issue with the assets loading in dev mod.

MontrealSergiy:
file type detection still works bad but  saving preset worked this time

MontrealSergiy:
the 'all' checkbox also disappears when one of sub checkboxes is unchecked, as intended

I installed a vm with new chrome, I do not have control over my workstation chrome versions image

PR #1253, opened 2022-08-01
Admins can force a user to authenticate with globus

prioux:
When creating an account or editing an account, an admin can configure
it such that the user must link it to a Globus identity provider.

The admins does so by entering the name of the identity provider(s) in user's
account form, something like:

```
Sydney Opera House, Hogwarts School of Witchcraft and Wizardry, University of Minnesota
```

(These names are case sensitive by the way).


prioux:
@MontrealSergiy @natacha-beck No need to review this PR. I am making it only for documentation purposes.

Issue #1188, opened 2021-12-09
Create new Boutiques support modules InputValueFixer


prioux:
Using the new support for modules in the new Boutiques integrator (#1181), I'd like a new module that fixes an input to a particular value.

The input would no longer be shown to the user in the form.

In the descriptor, the spec would look like:

```json
"custom": {
  "cbrain:integrator_modules": {
    "InputValueFixer": {
      "inputid1": "somevalue",
      "inputid2": 2345
    }
  }
}
```

prioux:
The PR that will implement this will have only 2 new files (one of them a symlink) and should consist in about 20 lines of code, total (rough estimate).

MontrealSergiy:
@prioux what about the input-value dependency (disables and requires and such), group validation , should those be adjusted (might need more than 20 lines of code) or dropped?

prioux:
No, don't make any adjustments. We only want to bare functionality I describe here.

We expect a programmer who use the module will do it responsibly and not choose an input that has other interfering rules applied to it.

PR #1254, opened 2022-08-01
Boutiques parameter values Fixer module

MontrealSergiy:
addresses issue #1188
It did not fit in 20 lines due to some basic handling of constraints/dependencies.

If must I can do stripped down version, yet I feel there could be constraints/dependencies among preset params
even in case of the resource control, for example options to use min memory possibly and memory below threshold could be mutually exclusive.

test file with dependencies could be

https://github.com/MontrealSergiy/cbrain-plugins-test/blob/fixer/boutiques_descriptors/dependencyhell.json

### DIFF
--- a/BrainPortal/lib/boutiques_input_value_fixer.rb
+++ b/BrainPortal/lib/boutiques_input_value_fixer.rb
@@ -63,7 +63,7 @@ module BoutiquesInputValueFixer
     # the major use case are Flags, but also may be useful to address params with 'default' (
     # or, for flags, null-like values)
 
-    descriptor_dup = descriptor.deep_dup
+    descriptor_dup = descriptor.dup
     skipped = invocation.keys.select do |i_id|
       begin
         input = descriptor_dup.input_by_id(i_id)
@@ -77,15 +77,22 @@ module BoutiquesInputValueFixer
 
     descriptor_dup.groups.each do |g| # filter groups
       members = g.members - invocation.keys
-      # delete a mutualy exclusive group if its member(s) fixed to a value(s)
-      if g.mutually_exclusive && members.length != g.members.length
-        if (invocation.keys & g.members - skipped).present?
-           g.mutually_exclusive == false
-           members = nil
+      # disable a mutualy exclusive group if its param assigned fixed value by this modifier
+      # if one simply deletes the fixed param,
+      if g.mutually_exclusive && members.length != g.members.length # params can be mutually exclusive e.g. --use-min-mem vs --mem-mb
+        if (invocation.keys & g.members - skipped).present? # at least some group members are actually assigned vals rather than deleted
+           if members.length == 1
+             g.mutually_exclusive = false # drop the restriction, has no point for one element
+             # (though ideally input should be suppressed)
+           else # when more than one member
+             g.all_or_none = true # adding all-or-none will block task submission.
+           end
+           # a better solution is to delete rest of group params completely
+           # a bit more complex though and might result in recursive code or nested loops
         end
       end
 
-      # removes one is required flag if one element fixed  --use-min-mem vs --mem-mb
+      # removes one-is-required flag if one element fixed
       if g.one_is_required && members.length != g.members.length
         if (invocation.keys & g.members - skipped).present?
           g.one_is_required == false
@@ -95,7 +102,8 @@ module BoutiquesInputValueFixer
       # removes one is required flag if one element fixed, e.g.
       if g.all_or_none && members.length != g.members.length
         if (g.members & skipped).present?
-          g.all_or_none == false
+          #g.all_or_none == false
+          g.mutually_exclusive = true # if more than 1 params remains remaining
           # todo delete all member inputs
         elsif (invocation.keys & g.members - skipped).present? # if one is set, rest should be to
           g.members.each do |i_id|
@@ -113,16 +121,9 @@ module BoutiquesInputValueFixer
       # with at most one quantifier flag per group
       # and in mutually exclusive (i.e. non-overlapping) groups.
       # if not, perhaps, more can be done
-
-      g.members = members.presence
-
-      # todo propagate dependencies described with input or value disables, enalbes and requires
-      # (it's easier to add internal 'hidden' attribute to inputs in the main codebase)
-      # or maybe just delete dependencies or maybe just check that fixed vars are not involved
-      # in dependecier
-
+      g.members = members
     end
-    descriptor_dup.groups = descriptor_dup.groups.compact
+    descriptor_dup.groups = descriptor_dup.groups.select {|g| g.members.present?} # delete empty group
 
     # delete fixed inputs
     descriptor_dup.inputs = descriptor_dup.inputs.select { |i| ! invocation.key?(i.id)} # filter out fixed inputs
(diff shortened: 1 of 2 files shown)


natacha-beck:
You said, it fix the issue 1254:

> addresses issue https://github.com/aces/cbrain/pull/1254 ,

But you link to the pull request can you point to the issue.

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,182 @@
...
+      #input.disables_if input.disables_inputs.present?
+      input.disables_inputs ||= []
+      input.disables_inputs |= [input_id]
+      input.name += " --- disabled by admin ---"
+    end
+  end
+
+  # adjust descriptor to allow check # of supplied files
+  def descriptor_for_before_form
+    delete_fixed_inputs(super)
```
prioux, line 159: I am not sure this is necessary. I don't really see why someone configuring a custom section of a desccriptor woudl provide a fixed value for a File input. Can you think of a case like that?

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,182 @@
...
+  # does its best to avoid deadlock and issues, but, probably,
+  # might fails in edge cases. It is recommended to supply fixation is "closed" in the sense that
+  # contains all the implied parameter fixes or deletions
+  def delete_fixed_inputs(descriptor)
+    # input parameters are marked by null values will be excluded from the command line
+    # the major use case are Flags, but also may be useful to address params with 'default' (
+    # or, for flags, null-like values)
+
+    descriptor_dup = descriptor.dup
+    skipped = fixation.keys.select do |i_id|
```
prioux, line 68: What is `skipped` exactly? I see a big select loop but there is no explanation as to what this contains at the end, exactly. It seems to be a subset of the IDs of the fixed inputs? But which subset, and what is it for?

prioux (changes requested):
Cleanup required

BrainPortal/lib/boutiques_input_value_fixer.rb
MontrealSergiy: atlas? config file?

BrainPortal/lib/boutiques_input_value_fixer.rb
MontrealSergiy: maybe `to_be_deleted` ? `to_abstract_away` ?

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,182 @@
...
+
+  # Note: to access the revision info of the module,
+  # you need to access the constant directly, the
+  # object method revision_info() won't work.
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  # the hash of param values to be fixed or be ommited ()  #
+  def fixation
+    fixation = self.boutiques_descriptor.custom_module_info('BoutiquesInputValueFixer')
+    cb_error "module BoutiquesInputValueFixer requires a hash" unless fixation.is_a? Hash
```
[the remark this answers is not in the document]
MontrealSergiy: Can it fail only for some parameters and not others and without exception, so admin would thing the issues in user data/parameters?

More, it is not even about configuration correct, are you sure that no information is leaking from/to all these descripter_before or descriptor_after? Seems like a good place to plant a bug/leak

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,182 @@
...
+#     "cbrain:integrator_modules": {
+#         "BoutiquesInputValueFixer": {
+#             "n_cpus": 1,
+#             "mem": "4G",
+#             "customquery": nil,
+#             "level": "group"
+#         }
+#     }
+# }
+# COULD BE TRICKY, PRESENTLY PARAMETER DEPENDENCY are not supported
```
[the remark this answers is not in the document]
MontrealSergiy: to warn the user (boutique integrator)

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,182 @@
...
+    fixation
+  end
+
+  # deletes fixed inputs listed in the custom 'integrator_modules'
+  # no input or values dependencies for fixed variables are fully supported,
+  # in the presence of dependencies involving fixed params, the module
+  # does its best to avoid deadlock and issues, but, probably,
+  # might fails in edge cases. It is recommended to supply fixation is "closed" in the sense that
+  # contains all the implied parameter fixes or deletions
+  def delete_fixed_inputs(descriptor)
```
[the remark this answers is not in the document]
MontrealSergiy: renamed

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,182 @@
...
+# string 'null' is considered string for String params
+module BoutiquesInputValueFixer
+
+  # Note: to access the revision info of the module,
+  # you need to access the constant directly, the
+  # object method revision_info() won't work.
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  # the hash of param values to be fixed or be ommited ()  #
+  def fixation
```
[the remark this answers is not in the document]
MontrealSergiy: ok

BrainPortal/lib/boutiques_input_value_fixer.rb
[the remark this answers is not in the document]
MontrealSergiy: ok

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,192 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# This module allows one to fix some of input parameters to specific values
+# The fixed input(s) would no longer be shown to the user in the form.
+#
+# In the descriptor, the spec would look like:
+#
+#    "custom": {
```
prioux, line 27: Please adjust this description block.

* the json example is badly indented
* add a blank line after the json (still commented though) so that rdoc creates a proper code block for it
* some sentences are not finished (`or their values or`)

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,192 @@
...
+  # you need to access the constant directly, the
+  # object method revision_info() won't work.
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  # the hash of param values to be fixed or be ommited ()  #
+  def fixed_values
+    self.boutiques_descriptor.custom_module_info('BoutiquesInputValueFixer')
+  end
+
+  require 'pry'
```
prioux, line 78: this is debut code that shouldn't be a in finished product

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,192 @@
...
+
+  # deletes fixed inputs listed in the custom 'integrator_modules'
+  def descriptor_without_fixed_inputs(descriptor)
```
prioux, line 81: Overall throughout this method you use `fixed_values.keys` a lot, and every time it is several layers of method calls that are executed just to get, over and over, the same set of input ids (something like `[ 'input1', 'input2', 'input3' ]`).

Can you just not, right at the beginning of the method, just extract this once?

Like in:

```ruby
fixed_input_ids = fixed_inputs.keys
```

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,192 @@
...
+      members = g.members - fixed_values.keys
+      # disable a mutualy exclusive group if its param assigned fixed value by this modifier
+      # if one simply deletes the fixed param,
+      if g.try(:mutually_exclusive) && members.length != g.members.length # params can be mutually exclusive e.g. --use-min-mem vs --mem-mb
+        if (fixed_values.keys & g.members - todelete).present? # at least some group members are actually assigned vals rather than deleted
+          g.mutually_exclusive = false
+          if g.try(:one_is_required).present?
+            g.one_is_required = false
+          end
+          block_inputs(descriptor_dup, members - fixed_values.keys)
```
prioux, line 108: I don't understand why you are subtracting `fixed_values.keys` from `members` here, given already at the top you initialize `members` to be a set where you remove them already (line 99)

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,192 @@
...
+      # removes one-is-required flag if one element fixed
+      # if g.one_is_required && members.length != g.members.length
+      #   if (fixed_values.keys & g.members - todelete).present?
+      #     g.one_is_required == false
+      #   end
+      # end
+
+      g.members = members
+    end
+    descriptor_dup.groups = descriptor_dup.groups.select {|g| g.members.present? } # delete empty group
```
prioux, line 141: You're already working on a full dup of the descriptor, so use the mutator methods instead, that will avoid a lot of copying:

```ruby
descrtiptor_dup.groups.reject! { |g| condition }
```

That also makes the code way easier to read.

BrainPortal/lib/boutiques_input_value_fixer.rb
```diff
@@ -0,0 +1,192 @@
...
+    # after submission of parameter).
+
+    descriptor_dup.groups.each do |g| # filter groups, relax restriction to ensure that form can still be submitted
+      members = g.members - fixed_values.keys
+      # disable a mutualy exclusive group if its param assigned fixed value by this modifier
+      # if one simply deletes the fixed param,
+      if g.try(:mutually_exclusive) && members.length != g.members.length # params can be mutually exclusive e.g. --use-min-mem vs --mem-mb
+        if (fixed_values.keys & g.members - todelete).present? # at least some group members are actually assigned vals rather than deleted
+          g.mutually_exclusive = false
+          if g.try(:one_is_required).present?
```
prioux, line 105: Why is there a `try` here? `g` is always not null, so you will always invoke `g.one_is_required`

prioux (changes requested):
Minor changes

BrainPortal/lib/boutiques_input_value_fixer.rb
prioux: You have not addressed this.

prioux (changes requested):
I think no work was done since my last review in February.

prioux (changes requested):
See comments?

MontrealSergiy:
not sure is it ok to rebase on master (most recent) or dev

prioux:
The new version seems OK, I will test it and if it works I will merge it

PR #1258, opened 2022-08-22
Prefetch tasks git file revision info resolves #1248

### LINKED ISSUE #1248: Bourreau Workers should prefetch file revision info
The CbrainFileRevision class caches GIT file revision info for the CBRAIN codebase the first time it's accessed.

This works well as long as the request for revision info, over time, happen within the same process.

On the Bourreau side however we spawn workers and subworkers, and when they fetch revision info, the info is not cached in the main Bourreau.

This means that whenever a subworker processes a task, say `BoutiquesTask::SomeSuperTool` and logs the GIT information in the task's log, the cached information is not available the next time around the subworker is spawned.

So I suggest the BourreauWorker, on its initial setting up phase, per-fetches the GIT revision info for all the possible task classes that it is configured to handle.

Something like:

```ruby
ToolConfig
  .where(:bourreau_id => current_bourreau_id)
  .map(&:tool)
  .uniq
  .map(&:cbrain_task_class)
  .each { |klass| klass.revision_info.self_update }
```

That way, when the Worker spawns its subworkers, they will already have all the revision info code in memory, and won't have to issue any redundant git commands.

MontrealSergiy:
Prefetches tasks git file revision info #1248

Note, this will fail e.g. if there are ghost tools (configs) from deleted boutiques. 

I can add some rescues if that's an issue, or just clean up the db from outdated configs

### DIFF
--- a/Bourreau/lib/bourreau_system_checks.rb
+++ b/Bourreau/lib/bourreau_system_checks.rb
@@ -460,23 +460,6 @@ class BourreauSystemChecks < CbrainChecker #:nodoc:
 
 
 
-  def self.a110_ensure_task_class_git_commits_cached
-
-    #----------------------------------------------------------------------------
-    puts "C> Ensuring git commits for tasks classes are pre-cached..."
-    #----------------------------------------------------------------------------
-
-    myself = RemoteResource.current_resource
-
-    ToolConfig.where(:bourreau_id => myself.id)
-        .map {|tc| tc.cbrain_task_class rescue nil}
-        .uniq
-        .compact  # to remove the nil
-        .each { |klass| klass.revision_info.self_update }
-  end
-
-
-
   def self.a100_ensure_dp_cache_symlink_exists #:nodoc:
 
     myself        = RemoteResource.current_resource
@@ -501,6 +484,23 @@ class BourreauSystemChecks < CbrainChecker #:nodoc:
 
 
 
+  def self.a110_ensure_task_class_git_commits_cached
+
+    #----------------------------------------------------------------------------
+    puts "C> Ensuring git commits for tasks classes are pre-cached..."
+    #----------------------------------------------------------------------------
+
+    myself = RemoteResource.current_resource
+
+    ToolConfig.where(:bourreau_id => myself.id)
+        .map {|tc| tc.cbrain_task_class rescue nil}
+        .uniq
+        .compact  # to remove the nil
+        .each { |klass| klass.revision_info.self_update }
+  end
+  
+  
+  
   def self.z000_ensure_we_have_a_forwarded_ssh_agent #:nodoc:
 
     #----------------------------------------------------------------------------
@@ -518,4 +518,3 @@ class BourreauSystemChecks < CbrainChecker #:nodoc:
   end
 
 end
-
--- a/Bourreau/lib/bourreau_system_checks.rb
+++ b/Bourreau/lib/bourreau_system_checks.rb
@@ -460,6 +460,23 @@ class BourreauSystemChecks < CbrainChecker #:nodoc:
 
 
 
+  def self.a110_ensure_task_class_git_commits_cached
+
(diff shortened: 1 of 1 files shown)


natacha-beck:
@MontrealSergiy works on my side reduce the execution of git command. 

But you said: 

> Note, this will fail e.g. if there are ghost tools (configs) from deleted boutiques.

How it will failed, what's happen in this specific case.

MontrealSergiy:
In a inconsistent db 
 the feature can raise exception  NameError while constatizing. But purging db and reinstall of plugins heped

MontrealSergiy:
a rescue will do, smth like `(ToolConfig.where(:bourreau_id => @rr_id).map {|tc| tc.cbrain_task_class rescue next}  ).uniq.each { |klass| klass.revision_info.self_update }`

natacha-beck:
According to your commet: 

> In a inconsistent db the feature can put down cbrain workers with exception NameError  > while constantizing. But purging db and reinstall of plugins helps. The exception, as the 
> cause of worker death only reported in logs, w.r.t GUI a worker dies silently

You should add rescue we cannot let worker die without any message. We cannot let worker dies just for that. 

> that

natacha-beck:
What about the following: 

```ruby
  ToolConfig.where(:bourreau_id => @rr_id)
            .map {|tc| tc.cbrain_task_class rescue nil}
            .uniq
            .compact   # to remove nil
            .each { |klass| klass.revision_info.self_update }
```

MontrealSergiy:
A bit longer than what I suggested above yet should work too

natacha-beck (approved):

prioux (changes requested):
The code is ok but it's done at the wrong time and place.

Right now, every time a worker starts, it will run a bunch of git commands to fetch the git revision info for each task class. This will happen in parallel and redundantly when lots of workers are being started at the same time.

The code should be performed before the workers are started so that they all inherit the same pre-fetched git information.

prioux (changes requested):
See my comment

MontrealSergiy:
@prioux  your comments addressed

MontrealSergiy:
if you like I can rebase

BrainPortal/app/models/bourreau.rb
```diff
@@ -406,6 +406,13 @@ def self.process_command_start_workers(command)
     # Returns a logger object or the symbol :auto
     logger = self.initialize_worker_logger(log_to,verbose)
 
+    # prefetch git commits info just before workers are started (reduces git invocation #)
```
prioux, line 409: This is still performed at the wrong place in the life cycle of the bourreau.

You are fetching the git revision numbers DURING the processing of the start workers command on the controls channel, this is not the role of the command interface. It doesn't know or shouldn't know anything about these specifics. This has lots of negative implications:

1. It would block the processing of the command until all the git revision info are received. On a slow filesystem (e.g. the home directory of a cluster), it would mean the start workers command would time out on the client side (suck commands are assumed to run in less than 10 seconds).
2. we'll be running this fetching code multiple times, every time we stop and start the workers, which is unnecessary

Move the code to a Bourreau system check instead

prioux (changes requested):
See comments

BrainPortal/app/models/bourreau.rb
MontrealSergiy: Code is moved to the system check. I pushed new branch to reduce commit numbers, as it is same code being pushed around

prioux:
OK I'll check this soon

prioux:
This seems ok to me exact for one minor thing: you created your method as `a110` and in the file you put tha method between `a090` and `a100`. Just put it at the proper place please.

MontrealSergiy:
> This seems ok to me except for one minor thing: you created your method as `a110` and in the file you put that method between `a090` and `a100`. Just put it at the proper place please.

ok, the method is moved toward the end

PR #1259, opened 2022-08-23
adding zip_archive userfile

tperezdevelopment:


prioux:
Looks good to me. @tperezdevelopment Remember to remove the code that was originally in your plugins. The definition of the ZipArchive model can only be at one place.

Issue #1164, opened 2021-10-22
File selection do not update propely in Userfile index


natacha-beck:
When we select a tool to launch without selecting a file we have the following message at the top of the Userfile index: 

<img width="1436" alt="Screen Shot 2021-10-22 at 4 10 24 PM" src="https://user-images.githubusercontent.com/777590/138516926-beecc6e1-000a-4d22-a368-033ba9e84519.png">

If we the select files the counter increase: 

<img width="1432" alt="Screen Shot 2021-10-22 at 4 12 25 PM" src="https://user-images.githubusercontent.com/777590/138517211-91e0c391-e4a6-44de-9a8d-cfb7e6426f93.png">

But if we filter the table and the  select a file the counter (and maybe the selection) stop to work.

Issue #1240, opened 2022-05-13
The 'prepare_tool_id' feature doesn't work in Firefox


prioux:
There's a problem with the JavaScript I think. It works in Chrome and Safari, but not Firefox.

I do get prompted to select files, but the file counter near the launch button never registers any selected files when I click the checkboxes.

Natacha, since you're the person who implemented the feature, can you debug this?

natacha-beck:
I wasn't able to reproduced this issue (tested on my laptop and on my desktop) both with Firefox.

Did you test it with an `incognito` windows ? The files selection did not work properly in `incognito` mode.

PR #1261, opened 2022-09-01
Fix counter in userfile index page

natacha-beck:
Fix issue #1164 and fix the #1240.

Note: it is better to use option `w=1` in GitHub to compare.


MontrealSergiy (approved):
works to me

MontrealSergiy:
not sure how urgent, but usually PR go on dev

BrainPortal/public/javascripts/userfiles.js
```diff
@@ -33,7 +33,7 @@ function cbrain_userfile_launch_bar(tool_name, file_status_text, have_selection,
 
   $('#menu_bar').after(
     $('<div class="launch_bar">')
-      .append($('<span class="info">Select some files to launch ' + tool_name + '</span>'))
+      .append($('<span class="info">Select some files to launch <span id="tool_name">' + tool_name + '</span></span>'))
```
prioux, line 36: Ho do you know that 'tool_name' is an ID that you can use? Is it used by some other pages in CBRAIN? How about giving it a name that is really sure to be unique for that feature?

BrainPortal/public/javascripts/userfiles.js
```diff
@@ -441,49 +483,6 @@ $(function() {
 
     /* Show/Hide dynamic actions/menu elements according to current selection */
     (function () {
-      // Generate the launch_task div if prepare_tool_id present as a parameter
-      function launch_task() {
```
prioux, line 445: So what happened here? This function was moved? Was it necessary?

And what happens to the launch_task() call at line 499 ? It's no longer in scope I think?

prioux (changes requested):
I think this breaks. See my comments.

BrainPortal/public/javascripts/userfiles.js
natacha-beck: I moved the function  in order to be able to call it at 2 different places:
- at line 56 in `cbrain_attach_userfile_checkboxes` (before some specific (buggy) code was called to update the number of files)
- at line 499 as before ( I test it and the launch_task function is reachable). 

This way I have a single function that is able to update the number of task selected, instead of having 2 differents pieces of code to do that.

Issue #1231, opened 2022-04-22
Add a watchdog process for the Bourreau Rails app.


prioux:
There is a SSH control master that provides DB, SshAgent and ActiveResource tunnels to the Bourreau app.

Sometimes, that SSH connection dies (for network issues or otherwise. Darn you CISCO routers!). Since the Bourreau is entirely passive and wakes up only when an XML message is received on its ActiveResource port, it stays around indefinitely and doesn't detect that this has happened.

I suggest that when a Bourreau is booted up, a small bash script (the watchdog) is launched in background that would check the aliveness of any of the SSH master's tunnels. This could be done indirectly, by calling the PING function of the Bourreau itself. If the connection is down, the Bourreau would exit (and also clean up its pidfile). The watchdog would check every 20 minutes or so. The watchdog also exits if the SSH master is found to have failed.

This would save us a lot of manual trouble, where we often have to connect to a Bourreau's site and check for leftover Bourreau processes and PID files.

The BourreauWorkers do not suffer from this problem, however. They make a periodic check on the DB for tasks to launch, so they do exit when this happens.

Suggestion: add this as part of PR #1229

prioux:
This is implemented in https://github.com/aces/cbrain/pull/1229/commits/82579a68d2919f9748b110b9bb9adbaa56709d13 in PR #1229

prioux:
Implemented in #1229. Yeah!

PR #1263, opened 2022-09-05
[RM] attribute from NH contact info table

### LINKED ISSUE #1227: Remove attribute nh_email_delivery_options in RemoteResource
The attribute `nh_email_delivery_options` is never used for anything; the current code base use `email_delivery_options` instead.

This requires a migration and removing one line in the model file `remote_resource.rb`

natacha-beck:
Closed #1227


natacha-beck:
Will do #1227 and #1231 together instead.

PR #1264, opened 2022-09-06
[RM] unused attributes from Remote Resource

### LINKED ISSUE #1227: Remove attribute nh_email_delivery_options in RemoteResource

### LINKED ISSUE #1232: Cleanup: remove attributes no longer used after PR #1229
PR #1229 cleans up a bunch of stuff but leaves in the DB two attributes of the RemoteResource model that are no longer necessary:

* tunnel_mysql_port
* tunnel_actres_port

These attributes can be removed from the model.

We should wait a few weeks after production deployment of #1229 to make sure everything is OK, in case we have to revert back to the old mechanism.

Also: clean up/adjust all the CBRAIN installation instructions relative to these values.

natacha-beck:
Closed #1227 and closed #1232

### DIFF
--- a/BrainPortal/spec/models/remote_resource_spec.rb
+++ b/BrainPortal/spec/models/remote_resource_spec.rb
@@ -334,7 +334,6 @@ describe RemoteResource do
     end
     it "should return a 'localhost' url if ssh control and active resource tunnel info given" do
       allow(remote_resource).to receive(:has_ssh_control_info?).and_return(true)
-      allow(remote_resource).to receive(:tunnel_actres_port).and_return(true)
       expect(remote_resource.site).to match(/^http:\/\/localhost/)
     end
   end
--- a/BrainPortal/app/models/remote_resource.rb
+++ b/BrainPortal/app/models/remote_resource.rb
@@ -58,7 +58,6 @@ class RemoteResource < ApplicationRecord
   # can only be set by the admin using the Rails console; leaving them blank
   # means the mailers use whatever is configured in the Rails environment config.
   serialize             :email_delivery_options
-  serialize             :nh_email_delivery_options
 
   validates             :name,
                         :uniqueness        => true,
--- a/BrainPortal/app/views/bourreaux/_notes.html.erb
+++ b/BrainPortal/app/views/bourreaux/_notes.html.erb
@@ -18,7 +18,7 @@
 # GNU General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.  
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 -%>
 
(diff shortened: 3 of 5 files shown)


MontrealSergiy (approved):
looks ok

BrainPortal/db/migrate/20220905142228_remove_old_remote_ressources_attributes.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveOldRemoteRessourcesAttributes < ActiveRecord::Migration[5.0]
+  def change
+    remove_column :remote_resources, :nh_email_delivery_options, :text,    :after => :nh_system_from_email
```
prioux, line 3: I don't think you need to provide type information for the remote_column operation.

See for instance

https://github.com/aces/cbrain/blob/master/BrainPortal/db/migrate/20170418124708_remove_docker_present_and_singularty_present_from_remote_resources.rb

prioux:
This PR is leaving other places in the code base that refer tothe port numbers. See the files in

BrainPortal/config/console_rc/lib

prioux:
Also, rebase.

prioux (changes requested):
This PR is not complete

BrainPortal/spec/models/remote_resource_spec.rb
```diff
@@ -333,8 +333,6 @@
       expect(remote_resource.site).to match(/^http:\/\//)
     end
     it "should return a 'localhost' url if ssh control and active resource tunnel info given" do
-      allow(remote_resource).to receive(:has_ssh_control_info?).and_return(true)
```
prioux, line 336: has_ssh_control_info? is still a valid method, why remove this from the test?

prioux:
I would like this file also updated:

https://github.com/aces/cbrain/blob/master/BrainPortal/app/views/bourreaux/_notes.html.erb

and the wiki:

https://github.com/aces/cbrain/wiki/Bourreau-Setup#3-bourreau-configuration

BrainPortal/db/migrate/20220905142228_remove_old_remote_ressources_attributes.rb
natacha-beck: I think I need to since is not `def down` but `def change` that I use. 

https://www.rubyinrails.com/2014/05/14/rails-migration-change-vs-up-down-methods/

natacha-beck:
> I would like this file also updated:
> 
> https://github.com/aces/cbrain/blob/master/BrainPortal/app/views/bourreaux/_notes.html.erb

--> I am not 100% sure, should I remove everything that is after line 75 https://github.com/aces/cbrain/blob/master/BrainPortal/app/views/bourreaux/_notes.html.erb#L75

natacha-beck:
> and the wiki:
> 
> https://github.com/aces/cbrain/wiki/Bourreau-Setup#3-bourreau-configuration

And here I should remove the `In the Tunnelling Configuration section` section right ?

prioux:
About "_notes.html.erb" : There are several sections that need adjustments:

* Hardcoded DB and ActiveResource (line 50): no longer supported
* Hardcoded, But With Remote Control (line 64): no longer supported
* Remote Control And Optional Tunnels (line 77): this is now the only way to configure the connection, and the admin does not have the ability to specify port numbers anymore. Also, the database.db is never kept on the server side anymore.

prioux:
For the wiki page, yes, just remove the paragraph.

natacha-beck:
I think I fixed the code according to your PR @prioux. 

Find here the [wiki modifications](https://github.com/aces/cbrain/wiki/Bourreau-Setup/_compare/2ecf46c894e48872378faf63933a61a36446f84a).

prioux:
I think you made a msiatke when you rebased/re arranged this PR. The migration file is no longer in the PR. Can you compare plase the set of files modified here (4 files) with the set of files that were modified before you re-arranged everything, just to make sure, in case other file changes are also missing?

PR #1267, opened 2022-09-15
[ADD] boutiques_dir_maker module

natacha-beck:
Some tools take expect that a file/directory input is
placed input a parent directory.

For example we have the following Boutiques command-line:

>     apptool [PRECOMPUTED_INPUT]

with the following input in Boutiques:

>     {
>       "description": "A directory that contain the precomputed data process by...",
>       "id": "precomputed_input",
>       "name": "Precomputed input",
>       "optional": false,
>       "type": "File",
>       "value-key": "[PRECOMPUTED_INPUT]"
>     }

and in the `cbrain:integrator_modules` section:

>     "BoutiquesDirMaker": {
>         "precomputed_input": "fake_parent_dir"
>      }

In CBRAIN the user will select a userfile for example `sub-n`,
the final command line will become:

>     apptool precomputed/sub-n


prioux:
I have several comments to make about the design and implementation, but for the moment I'll point out that the module is not installed on the bourreau side so this cannot work.

MontrealSergiy:
Yup, @natacha-beck you need to add soft links to Bourreau side. Does any consumers of derivative/precomputed bids require participant or any other home folder files? (May be ask Eric?)

MontrealSergiy:
Natacha are we targeting Nibabies + Cabinet? I believe command line should become

```apptool fake_parent_dir```

do you have some examples of command lines and data for these pipelines to test?

**Comment by prioux** [comment]
I'm just seeing lots of places where the changes will make things crash...

File: BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_task_params.html.erb [hunk 1]
```diff
@@ -34,6 +34,16 @@
   groups      = @descriptor.groups          || []
   params_in_groups = groups.map(&:members).flatten.uniq # list of param IDs
   interface_files  = Userfile.where(:id => @task.params[:interface_userfile_ids])
+
+  interface_files_name   = interface_files.pluck(:name)
```
prioux, line 38: I don't understand this at all, and as a change to the framework I'm pretty it will crash in plenty of places. The partials expect interface_files to be an ActiveRecord relation and they will issue methods specific to AR (e.g. `.pluck()`) which are going to crash when they're given this array of small objects. Also, why the weird "name" value that contains a data provider name in parenthesis? If oyu wanted to chaneg the appearance of the names of the files in the form, you could replace in the view code `userfile.name` with the method `userfile.browse_name` instead, maybe?

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,166 @@
...
+  # For input in the `BoutiquesDirMaker` section,
+  # create a directory based on the value BoutiquesDirMaker[input_id]
+  # that will contains a symlink to the orginal selected Userfile.
+  def setup #:nodoc:
+    return false unless super
+
+    descriptor = self.descriptor_for_setup
+    self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
+
+    boutiques_dir_maker_list = descriptor.custom_module_info('BoutiquesDirMaker')
```
prioux, line 94: Why is this variable called `_list` when it's a hash?

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,166 @@
...
+  # the fake parent directory
+  def descriptor_for_form #:nodoc:
+    descriptor = super.dup
+    boutiques_dir_maker_list = descriptor.custom_module_info('BoutiquesDirMaker')
+
+    boutiques_dir_maker_list.each do |inputid,dirname|
+      # Adjust the description
+      input        = descriptor.input_by_id(inputid)
+      description  = input.description.presence || ""
+      description += "\n" if description
```
MontrealSergiy, line 73: to avoid endofline in the beginning of  description, either ```description += "\n" if description.present?```

BrainPortal/lib/boutiques_dir_maker.rb
prioux: When naming a hash, it is customary to name it after a description of the VALUES in it. E.g.

```ruby
fruits = { banana: 'yellow', apple: 'red' } # BAD name
colors = { banana: 'yellow', apple: 'red' } # BETTER name

# When accessing, it makes reading the code easier when the hash is named after the values:
color = fruits['banana'] # BAD
color = colors['banana'] # BETTER
```

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,166 @@
...
+    boutiques_dir_maker_list = descriptor.custom_module_info('BoutiquesDirMaker')
+    descriptor.file_inputs.each do |input|
+      userfile_id = invoke_params[input.id]
+      next if userfile_id.blank? # that happens when it's an optional file
+      userfile    = Userfile.find(userfile_id)
+      dirname     = boutiques_dir_maker_list[input.id]
+
+      # Most common situation
+      if ! input.list || ! userfile.is_a?(CbrainFileList)
+        make_available(userfile, dirname.present? ? "#{dirname}/#{userfile.name}" : userfile.name)
```
prioux, line 103: The method make available was already invoked in super(), why do it again?!? Just create a symlink to whatever make_available created for you?

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,166 @@
...
+        next
+      end
+
+      # In case the input is a list and is assigned a CbrainFileList
```
prioux, line 107: I don't understand this either... you're duplicating exactly a bunch of work done by the framework in super(). And you're not even implementing the necessary code to support file lists inside your "parent dir".

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,166 @@
...
+  def descriptor_for_cluster_commands
+    descriptor = super.dup
+
+    command    = descriptor.command_line
+
+    boutiques_dir_maker_list = descriptor.custom_module_info('BoutiquesDirMaker')
+    boutiques_dir_maker_list.each do |inputid,dirname|
+      input         = descriptor.input_by_id(inputid)
+      token         = input.value_key # e.g. '[PRECOMPUTED_INPUT]'
+      userfile_id   = invoke_params[inputid]
```
prioux, line 142: This is all wrong, you're substituting filenames inside the command line before bosh is given the task to do it. Bosh will complain and maybe crash when it doesn't find the substitution tokens it wants. If you're modifying the command-line, why not just replace

```ruby
   "blah blah [PRECOMPUTED_INPUT] blah blah"
```

by

```ruby
   "blah blah #{dirname.bash_escape}/[PRECOMPUTED_INPUT] blah blah"
```

that way the the substitution is still the responibility of bosh?

Also, the way you implemented it, doing a string substitution inside the command line, is a security risk for code injection! You can't just substitute things under user control without escaping.

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,166 @@
...
+      # In order to prevent bosh from complaining if the value-key is no longer found
+      # anywhere in the command-line, we re-insert a dummy no-op bash statement at the
+      # beginning of the command with at least one use of that value-key. It will look
+      # like e.g.
+      #
+      #   "true [PRECOMPUTED_INPUT] ; real command here"
+      #
+      # In bash, the 'true' statement doesn't do anything and ignores all arguments.
+      if ! command.include? token
+        command = "true #{token} ; #{command}"
```
prioux, line 157: See my comment above, when adjusting the command line properly, you don't need any of these hacks here.

prioux (changes requested):
Lots of bugs and redesign needed.

BrainPortal/lib/boutiques_dir_maker.rb
MontrealSergiy: or use following hack to avoid conditionals at all

```
    
      # Adjust the description
      input        = descriptor.input_by_id(inputid)
      new_paragraphs =   "This input will be copied in a parent folder `#{dirname}`.\n The parent folder will be used in the command line"
      input.description  = [input.description.presence,  new_paragraphs].compact.join('\n')

Comment (MontrealSergiy) [reply BrainPortal/lib/boutiques_dir_maker.rb:73]: or, with save navigation operator
```
      # Adjust the description
      input        = descriptor.input_by_id(inputid)
      new_paragraphs =   "This input will be copied in a parent folder `#{dirname}`.\n The parent folder will be used in the command line"
      input.description  = input.description.presence&.append(\'n')  + new_paragraphs
```

**Comment by natacha-beck** [comment]
> `apptool fake_parent_dir`



> Natacha are we targeting Nibabies + Cabinet? I believe command line should become
> 
> `apptool fake_parent_dir`
> 
> do you have some examples of command lines and data for these pipelines to test?

Here it is, according to Eric email's: 

> We do not want to provide sub-n directly to NiBabies, we want to provide something like ./precomputed/sub-n (although the folder doesn't explicitly need to be named precomputed, it just has to have the sub-n "precomputed" results directly beneath it)

File: BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_task_params.html.erb (hunk 1 above)
Comment (natacha-beck) [reply BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_task_params.html.erb:38]: If you have an Array of Hash (it is my case here). 
You can use pluck, for example: 

> Userfile.last(10).map {|x| {type: x.type, name: x.name, id: x.id}}.pluck(:type) 

But right I can use browse_name instead.

**Comment by MontrealSergiy** [comment]
@natacha-beck for some reason I cannot directly continue thread. What concerns Eric comment are you sure he talks about the command line and not CBRAIN tool? Do not you recall Eric was basically asking to generalize the SingleSubjectMaker onto many bids datasets (and Pierre's SingleSubjectMaker does not contain sub-xxxx subfolder in the command line). Thus, (unless sub-xxx has another sub-xxx/sub-xxx under it), sub-xxx should be dropped from the command line.

```
according to Eric email's:

We do not want to provide sub-n directly to NiBabies, we want to provide something like ./precomputed/sub-n (although the folder doesn't explicitly need to be named precomputed, it just has to have the sub-n "precomputed" results directly beneath it)
```

**Review by MontrealSergiy — CHANGES_REQUESTED** [review]
major issue is command line should be `apptool precomputed` and not `apptool precomputed/sub-n`, single subject input is cbrain thing, all command line tool accept bids datasets.

See also few minor comments

File: BrainPortal/lib/boutiques_create_fake_parent_dir.rb [hunk 8]
```diff
@@ -0,0 +1,139 @@
...
+  end
+
+  ############################################
+  # Bourreau (Cluster) Side Modifications
+  ############################################
+
+  # For input in the `BoutiquesCreateFakeParentDir` section,
+  # create a fake parent directory that will contains a symlink
+  # to the orginal selected Userfile.
+  def setup #:nodoc:
```
prioux, line 107: I think the overall solution of setting aside the IDs of the input file to prevent the standard setup() to try to run make_available() on them is correct, but we're doing it the wrong way (and yes I know it was my suggestion). We can't set them aside in the after_form() method on the portal side. If we do that, the system won't work for task arrays (e.g. when a CbrainFileList is provided as an input for a File). 

The code can be adjusted in a much simpler way: everything is done in setup(), by insterting code BEFORE and AFTER the super() call:

```ruby
   def setup
      remove_ids_from_invoke_struct_here
      result = super
      reinstert_ids_here
      return result if ! result
      do_special_make_available_here
      return true or false # depending on maybe other checks
   end
```

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
```diff
@@ -0,0 +1,139 @@
...
+      params["BoutiquesCreateFakeParentDir"][inputid] = fake_parent_dirname
+      make_available(userfile, "#{fake_parent_dirname}/#{userfile.name}")
+    end
+
+    true
+  end
+
+  # Overrides the same method in BoutiquesClusterTask, as used
+  # during cluster_commands()
+  def finalize_bosh_invoke_struct(invoke_struct) #:nodoc:
```
prioux, line 133: There is no need to do that if we follow my previous suggestion.

prioux (changes requested):
New mechanism suggested

BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_boutiques_input.html.erb
```diff
@@ -81,7 +81,10 @@
       <%= render :partial => task_partial(:dropdown), :locals =>
             { :id        => input.cb_invoke_html_id(false), # false disable list input's '[]'
               :name      => input.cb_invoke_html_name(false), # false disable list input's '[]'
-              :options   => interface_files.pluck(:id, :name),
+              :options   => interface_files.map{|file|  [file.id, file.browse_path.present? ?
```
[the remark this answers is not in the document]
prioux: Improve this please, like I suggested?

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
```diff
@@ -0,0 +1,152 @@
...
+  # to the orginal selected Userfile.
+  def setup #:nodoc:
+
+    descriptor = self.descriptor_for_setup
+    self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
+
+    parent_dirname_by_inputid = descriptor.custom_module_info('BoutiquesCreateFakeParentDir')
+
+    # Remove IDs from invoke_params
+    params["BoutiquesCreateFakeParentDir"] = {}
```
prioux, line 102: You don't need to save the IDs in the params structure. Just keep it in a hash variable local to the method!

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
```diff
@@ -0,0 +1,152 @@
...
+    # Special make_available who need to have a parent folder
+    parent_dirname_by_inputid.each do |inputid,fake_parent_dirname|
+      userfile_id = params["BoutiquesCreateFakeParentDir"][inputid]
+
+      params['BoutiquesCreateFakeParentDir'][inputid] = nil if userfile_id.blank?
+      next if userfile_id.blank?
+
+      userfile    = Userfile.find(userfile_id)
+
+      if !Userfile.is_legal_filename?(fake_parent_dirname)
```
prioux, line 127: Don't make checks like this, it's not like users can control these strings.

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
```diff
@@ -0,0 +1,152 @@
...
+    return false if !super
+
+    # Re-insert IDs in invoke_params
+    invoke_params.each do |inputid, userfile_id|
+      next if userfile_id.present?
+      invoke_params[inputid] = params["BoutiquesCreateFakeParentDir"][inputid]
+    end
+
+    # Special make_available who need to have a parent folder
+    parent_dirname_by_inputid.each do |inputid,fake_parent_dirname|
```
prioux, line 119: This entire loops does a bunch of unnecessary things

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
```diff
@@ -0,0 +1,152 @@
...
+      params['BoutiquesCreateFakeParentDir'][inputid] = fake_parent_dirname
+    end
+
+    true
+  end
+
+  # Overrides the same method in BoutiquesClusterTask, as used
+  # during cluster_commands()
+  def finalize_bosh_invoke_struct(invoke_struct) #:nodoc:
+    self.params['BoutiquesCreateFakeParentDir'].each do |inputid, dirname|
```
prioux, line 142: super() is a method call, and every time you'll be editing potentially a different hash; you cannot have a loop that has super() in it, you're supposed to just call it once, override it's content and return the new content.

prioux (changes requested):
Several bugs still

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
natacha-beck: User no but the creator of the Boutiques descriptor can, that´s why I have a checklist. 
If you really think that it´s not useful I can remove.

prioux:
One more overall comment. I'd like the entire module to be renamed:

`BoutiquesInputSubdirMaker`

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
```diff
@@ -0,0 +1,159 @@
...
+
+    # Special make_available who need to have a parent folder
+    parent_dirname_by_inputid.each do |inputid,fake_parent_dirname|
+      dirname     = fake_parent_dirname[0]
+      userfile_id = original_userfile_ids[inputid]
+
+      next if userfile_id.blank?
+
+      self.addlog("Creating parent directory '#{dirname}'")
+      self.addlog("#{basename} rev. #{commit}")
```
prioux, line 127: Log the revision information at the beginning of tje method., not inside this loop! You'll be adding one line to the log for each file, completely redundantly.

prioux (changes requested):
Change log.

Also see my comment in the main discussion about renaming completely this module.

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
```diff
@@ -0,0 +1,159 @@
...
+  def descriptor_for_form #:nodoc:
+    descriptor                = super.dup
+    parent_dirname_by_inputid = descriptor.custom_module_info('BoutiquesCreateFakeParentDir')
+
+    parent_dirname_by_inputid.each do |inputid,fake_parent_dirname|
+      # Adjust the description
+      input              = descriptor.input_by_id(inputid)
+      dirname            = fake_parent_dirname[0]
+      input.description  = input.description.to_s +
+                           "\nThis input will be copied in a parent folder: #{dirname}.The parent folder will be used in the command line"
```
prioux, line 85: Spaceing in text

BrainPortal/lib/boutiques_create_fake_parent_dir.rb
```diff
@@ -0,0 +1,159 @@
...
+  ensure
+    original_userfile_ids.each do |inputid, userfile_id|
+      invoke_params[inputid] = original_userfile_ids[inputid]
+    end
+  end
+
+  # Overrides the same method in BoutiquesClusterTask, as used
+  # during cluster_commands()
+  def finalize_bosh_invoke_struct(invoke_struct) #:nodoc:
+    invoke_params = super.dup
```
prioux, line 143: This variable is just a local variable but you gave it the exact same name as an important method of the the task object. Rename it to avoid confusion.

BrainPortal/lib/boutiques_input_subdir_maker.rb
```diff
@@ -0,0 +1,163 @@
...
+    override_invoke_params = super.dup
+
+    descriptor = self.descriptor_for_cluster_commands
+    parent_dirname_by_inputid = descriptor.custom_module_info('BoutiquesInputSubdirMaker')
+    parent_dirname_by_inputid.each do |inputid,fake_parent_dirname|
+      if override_invoke_params[inputid].blank?
+        override_invoke_params.delete(inputid)
+      else
+        (dirname, append_userfile_name) = fake_parent_dirname
+        override_invoke_params[inputid] = append_userfile_name ? "#{dirname}/#{override_invoke_params[inputid]}" : dirname
```
prioux, line 156: This doesn't work, you're going to set the entry in `override_invoke_params` to `dirname/userfileID` instead of `dirname/userfileNAME`

BrainPortal/lib/boutiques_input_subdir_maker.rb
```diff
@@ -0,0 +1,163 @@
...
+# and in the `cbrain:integrator_modules` section:
+#
+#     "BoutiquesInputSubdirMaker": {
+#       "precomputed_input": ["precomputed", true],
+#       "derivatives_input": ["derivatives", false]
+#      },
+#
+# In CBRAIN the user will select 2 userfiles for example:
+#
+# For precomputed option `precomputed/sub-n`,
```
prioux, line 60: I don't understand this explanation, in CBRAIN userfiles don't select `precomputed/sub-2` or anything of like that

BrainPortal/lib/boutiques_input_subdir_maker.rb
prioux: Ok this works, the name has been substituted by the time this method is reached.

prioux (approved):
Looks good to me

PR #1270, opened 2022-09-26
ToolConfigs to auto-propagate env variables to singularity fixes #1266

### LINKED ISSUE #1266: ToolConfigs to auto-propagate env variables to singularity
Our ToolConfig model allows the admin to configure environment variables settings for the tool, e.g. `ABC=xyz`. These setting are transformed into a set of `export ABC=xyz` statements in the job script.

When a tool is run within Singularity, these environment variables will be propagated by default into the container, except if the admin configures the `--cleanenv` (or `-c`) option when singularity is invoked (in the ToolConfig attribute `container_exec_args` which shows up as 'Misc Singularity options in the page). In that case the admin must either

1. add explicitly another `--env ABC=xyz` (or better `--env ABC=$ABC`) in the Singularity options
2. change the name of the environment variables in the ToolConfig env table so they get prefixed with "SINGULARITYENV_" (e.g.  "SINGULARITYENV_ABC")

Option 1 means redundancy in defining the environment variables, and option 2 means that the tool config's environment table is now 'aware' of the containerized situation, which can lead to inconsistency when configuring the exact same tool without a container.

In order to make the life of the admin easier, I suggest that whenever a tool is launched through singularity, the set of export statements created by the **ClusterTask** object be replaced (or extended) with assignments of the form `export SINGULARITYENV_ABC=xyz`.

MontrealSergiy:


MontrealSergiy:
CI fail fixing bugs

PR #1272, opened 2022-09-28
Propagate environment variables to container fixes #1266

### LINKED ISSUE #1266: ToolConfigs to auto-propagate env variables to singularity

MontrealSergiy:


BrainPortal/cbrain_plugins/cbrain-plugins-base/boutiques_descriptors/singularity_env_demo.json
```diff
@@ -0,0 +1,45 @@
...
+            "list":          false,
+            "path-template": "old*"
+        },
+        {
+            "name":          "new file",
+            "id":            "s2",
+            "description":   "The output of the 'du' command in human units",
+            "optional": true,
+            "list":          false,
+            "path-template": "new"
```
natacha-beck, line 35: Should be `new*` no ?

MontrealSergiy:
@natacha-beck ok the description is  contrived. It is test/demo descriptor for the new module. May be better delete it or place in a separate  elsewhere?

BrainPortal/cbrain_plugins/cbrain-plugins-base/boutiques_descriptors/singularity_env_demo.json
```diff
@@ -0,0 +1,45 @@
+{
+    "name":           "ENVPrefix",
+    "tool-version":   "1.0.0",
+    "schema-version": "0.5",
+    "author":         "Serge",
+    "description":    "A demo CBRAIN task that adds SINGULARITYENV_ prefix to config environment variable names ",
+
+    "command-line": "touch old_{$MYVARIABLE}; touch new_{$SINGULARITYENV_MYVARIABLE} ; du -a -H -h [CBRAIN_INPUT] > out",
```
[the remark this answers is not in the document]
prioux: That is bad bash syntax.

Substitution of variables in bash is either plain `$var` or `${var}`. This code says `{$var}`, which always leave the curly brackets there.

BrainPortal/cbrain_plugins/cbrain-plugins-base/boutiques_descriptors/singularity_env_demo.json
MontrealSergiy: I removed that output to make tool seem more plausible

natacha-beck:
One general question. Should we update the tool.config id a container engine exist ?

MontrealSergiy:
> One general question, should we update the tool.config id a container engine exist ?

Not sure since most tools deployed in a singularity, but, If you like, ok

prioux:
This pool request doesn't do at all what I wanted.

The original issue #1266 is about the singularity execution layer, not about Boutiques integration. The code that needs to be changed is in the CBRAIN framework.

I am closing this PR because CBRAIN doesn't need at all a solution at this level of the execution stack.

PR #1276, opened 2022-10-10
Boutiques modules integrator information

### LINKED ISSUE #1202: Log more information about boutique descriptors
Task logs could include some more information about the descriptors. I think we already log the GIT revision info of the descriptor itself, but we could add:

* the revision info of the included modules, if any
* the content of "custom" -> "cbrain:author"

I'm not sure at what stages of the task's lifecycle to add all that, however.

Along with this, mayby make sure all overriden methods in included modules add at least one log entry at run time, just FYI.

natacha-beck:
Close #1202

### DIFF
--- a/BrainPortal/app/models/boutiques_cluster_task.rb
+++ b/BrainPortal/app/models/boutiques_cluster_task.rb
@@ -76,7 +76,7 @@ class BoutiquesClusterTask < ClusterTask
     descriptor = self.descriptor_for_setup
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
 
-    module_information;
+
 
     descriptor.file_inputs.each do |input|
       userfile_id = invoke_params[input.id]
@@ -420,19 +420,6 @@ class BoutiquesClusterTask < ClusterTask
   # for output.
   def file_access_symbol
     @_file_access ||= (self.class.properties[:readonly_input_files].present? || self.tool_config.try(:inputs_readonly) ? :read : :write)
-  end
-
-  private
-
-  def module_information
-    descriptor = self.descriptor_for_setup
-
-    integrator_modules = descriptor['custom']['cbrain:integrator_modules']
-    integrator_modules.each_key do |module_name|
-      klass    = module_name.constantize
-      rev_info = klass::Revision_info
-      self.addlog("#{rev_info.basename} rev. #{rev_info.short_commit} #{rev_info.time} (author: #{rev_info.author})");
-    end
   end
 
 end
--- a/BrainPortal/app/models/boutiques_cluster_task.rb
+++ b/BrainPortal/app/models/boutiques_cluster_task.rb
@@ -76,8 +76,6 @@ class BoutiquesClusterTask < ClusterTask
     descriptor = self.descriptor_for_setup
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
 
-
-
     descriptor.file_inputs.each do |input|
       userfile_id = invoke_params[input.id]
       next if userfile_id.blank? # that happens when it's an optional file
--- a/BrainPortal/app/models/boutiques_cluster_task.rb
+++ b/BrainPortal/app/models/boutiques_cluster_task.rb
@@ -76,6 +76,8 @@ class BoutiquesClusterTask < ClusterTask
     descriptor = self.descriptor_for_setup
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
 
+    module_information;
+
     descriptor.file_inputs.each do |input|
       userfile_id = invoke_params[input.id]
       next if userfile_id.blank? # that happens when it's an optional file
@@ -418,6 +420,19 @@ class BoutiquesClusterTask < ClusterTask
   # for output.
   def file_access_symbol
     @_file_access ||= (self.class.properties[:readonly_input_files].present? || self.tool_config.try(:inputs_readonly) ? :read : :write)
+  end
+
+  private
+
+  def module_information
+    descriptor = self.descriptor_for_setup
+
+    integrator_modules = descriptor['custom']['cbrain:integrator_modules']
+    integrator_modules.each_key do |module_name|
+      klass    = module_name.constantize
+      rev_info = klass::Revision_info
+      self.addlog("#{rev_info.basename} rev. #{rev_info.short_commit} #{rev_info.time} (author: #{rev_info.author})");
+    end
   end
 
 end
--- a/BrainPortal/app/models/boutiques_portal_task.rb
+++ b/BrainPortal/app/models/boutiques_portal_task.rb
@@ -267,16 +267,10 @@ class BoutiquesPortalTask < PortalTask
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
     valid_input_keys = descriptor.inputs.map(&:id)
 
-<<<<<<< HEAD
     # Add information about Boutiques module
     module_information = boutiques_module_information();
     module_information.map do |info|
-=======
-    # add information about Boutiques module
-    module_information = boutiques_module_information()
-    module_information.each do |info|
->>>>>>> [FIX] after PR review
-       self.addlog(info)
+       self.addlog(log_info)
     end
 
     # --------------------------------------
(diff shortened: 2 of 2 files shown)


BrainPortal/app/models/boutiques_cluster_task.rb
```diff
@@ -76,6 +76,8 @@ def setup #:nodoc:
     descriptor = self.descriptor_for_setup
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
 
+    module_information;
```
prioux, line 79: That's a bad name. With a name like that, it feels it's returning something, but there is no assignment.

What the method does is log informaton about the module. So name it appropriately: "addlog_boutiques_modules_information". All methods that add to the logging system start with "addlog_", after all.

Then, invoke it with self so that it looks like the previous line:

```ruby
   self.addlog_boutiques_module_information
```

For discussion: all these addlogs are added at the time of setup(). I am not sure that's the best place. Maybe in after_final_task_list_saved() ?

prioux: Finally, if this works out, then remove the addlog() statements in the modules themselves, they become redundant.

BrainPortal/app/models/boutiques_cluster_task.rb
```diff
@@ -420,4 +422,17 @@ def file_access_symbol
...
   end
 
+  private
+
+  def module_information
+    descriptor = self.descriptor_for_setup
+
+    integrator_modules = descriptor['custom']['cbrain:integrator_modules']
+    integrator_modules.each_key do |module_name|
+      klass    = module_name.constantize
```
prioux, line 432: It's not a `klass`, it's a `modname`

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -264,6 +264,12 @@ def final_task_list #:nodoc:
     descriptor = self.descriptor_for_final_task_list
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
 
+    # add information about Boutiques module
+    module_information = boutiques_module_information();
+    module_information.map do |info|
```
prioux, line 269: why map() instead of each() ?

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -622,4 +628,21 @@ def file_access_symbol
     @_file_access ||= (self.class.properties[:readonly_input_files].present? || self.tool_config.try(:inputs_readonly) ? :read : :write)
   end
 
+  private
+
+  def boutiques_module_information
```
prioux, line 633: Add comment block before method, add `#:nodoc:`

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -622,4 +628,21 @@ def file_access_symbol
...
+    descriptor = self.descriptor_for_final_task_list
+
+    module_information = []
+
+    integrator_modules = descriptor['custom']['cbrain:integrator_modules']
```
prioux, line 638: This code will  crash for any boutiques descriptor that has no integrator module (or no custom section)

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -622,4 +628,21 @@ def file_access_symbol
...
+    integrator_modules.each_key do |module_name|
```
prioux, line 639: don't do

```ruby
blah = []
something.each do
  blah << xyz
end
```

when you can do

```ruby
blah = something.map do
   xyz
end
```

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -622,4 +628,21 @@ def file_access_symbol
...
+      module_name = module_name.constantize
+      rev_info    = module_name::Revision_info
+      module_information.push("#{rev_info.basename} rev. #{rev_info.short_commit} #{rev_info.time} (author: #{rev_info.author})");
```
prioux, line 642: Why the semicolon?

BrainPortal/app/models/boutiques_cluster_task.rb
```diff
@@ -76,6 +76,8 @@ def setup #:nodoc:
     descriptor = self.descriptor_for_setup
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
 
+
```
prioux, line 79: new blank lines? why?

prioux (changes requested):
Lots of changes to do.

BrainPortal/app/models/boutiques_portal_task.rb
prioux: Also check your semicolons.... you add them by mistake regularly

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -622,4 +628,24 @@ def file_access_symbol
...
 
+  private
+
+  # Prepare an array with revision information of
+  # all the Boutiques integrator modules used by the
+  # tools.
+  def boutiques_module_information #:nodoc:
+    descriptor = self.descriptor_for_final_task_list
+
+    integrator_modules = (descriptor['custom'] && descriptor['custom']['cbrain:integrator_modules']) || {}
```
prioux, line 639: This is clumsy... just do

```ruby
integrator_modules = descriptor.custom['cbrain:integrator_modules'] || {}
```

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -622,4 +628,24 @@ def file_access_symbol
...
+
+    module_information = []
```
prioux, line 641: This variable is initialized for no reason, it's reset in the map just the line below.

prioux: And then there is no point in even having the variable, since the map is the last statement of the method.

The method shoudl just end with

```ruby
integrator_modules.map do |module_name, value|
  blha blah
end # block end
end # method end
```

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -264,6 +264,12 @@ def final_task_list #:nodoc:
     descriptor = self.descriptor_for_final_task_list
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
 
+    # add information about Boutiques module
+    module_information = boutiques_module_information()
```
prioux, line 268: Remove the variable. just do

```ruby
boutiques_module_information.each { |log_line| self.addlog(log_ling) }
```

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -622,4 +627,21 @@ def file_access_symbol
...
+
+  # Prepare an array with revision information of
+  # all the Boutiques integrator modules used by the
+  # tools.
+  def boutiques_module_information #:nodoc:
+    descriptor = self.descriptor_for_final_task_list
+
+    integrator_modules = descriptor.custom['cbrain:integrator_modules'] || {}
+
+    return module_information = integrator_modules.map do |module_name, value|
```
prioux, line 640: 1. The `return module_information =` is unnecessary, since you're returning the value of the map anyway.
2. the 'value' variable is never used, so replace it with `_`

prioux (changes requested):
See comments

prioux:
The code need to be rebased. There are new methods that are in the way.

It should be easy, if you move the 'private' keyword and the method to the end of the class. Careful because `private` affects all the methods that follow.

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -267,6 +267,12 @@ def final_task_list #:nodoc:
     self.addlog(descriptor.file_revision_info.format("%f rev. %s %a %d"))
     valid_input_keys = descriptor.inputs.map(&:id)
 
+    # Add information about Boutiques module
+    module_information = boutiques_module_information()
```
prioux, line 271: Can we simplify this even further now? Why assign the information to a variable?

prioux (commented):
See comment.

PR #1277, opened 2022-10-11
Load description when use preset of a task

### LINKED ISSUE #1257: Load preset of a task do not populate description.
When a task_preset is saved with a description, the description is saved on the preset (I verified in the console). 
But when the task_preset is reload the description is not populated with the value present in the task_preset.

natacha-beck:
Close #1257. 
Now when a user load preset of a task the description will be filled.

### DIFF
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -1503,7 +1503,7 @@ class TasksController < ApplicationController
         old_params = @task.params.clone
         @task.params         = preset.params
         @task.description    = @task.description || ""
-        @task.description   += @task.description.blank? ? "#{preset.description}" : "\n\nDescription from preset configuration:\n\n#{preset.description}"
+        @task.description   += "\n\nDescription from preset configuration:\n\n#{preset.description}" if preset.description.present?
         @task.restore_untouchable_attributes(old_params, :include_unpresetable => true)
         if preset.group && preset.group.can_be_accessed_by?(current_user)
           @task.group = preset.group
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -1502,6 +1502,7 @@ class TasksController < ApplicationController
       if (! preset_id.blank?) && preset = CbrainTask.where(:id => preset_id, :status => [ 'Preset', 'SitePreset' ]).first
         old_params = @task.params.clone
         @task.params         = preset.params
+        @task.description    = @task.description || ""
         @task.description   += @task.description.blank? ? "#{preset.description}" : "\n\nDescription from preset configuration:\n\n#{preset.description}"
         @task.restore_untouchable_attributes(old_params, :include_unpresetable => true)
         if preset.group && preset.group.can_be_accessed_by?(current_user)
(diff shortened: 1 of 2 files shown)


MontrealSergiy (approved):
seems ok. Not sure of any need to removal all existing module logs, as they carry additional info, e.g  exactly module was used

prioux:
This PR includes the code of another PR.

Please be careful when you create your local branches!

prioux (changes requested):
Adjust teo remove code of other PR

MontrealSergiy (approved):
seems ok. I would probably not copy a nil descriptor, though not quite sure

BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -1502,6 +1502,7 @@ def handle_preset_actions #:nodoc:
       if (! preset_id.blank?) && preset = CbrainTask.where(:id => preset_id, :status => [ 'Preset', 'SitePreset' ]).first
         old_params = @task.params.clone
         @task.params         = preset.params
+        @task.description    = preset.description
```
prioux, line 1505: Can we get a description merge maybe? This will just replace outright whatever other description the user has already entered in the task, or whatever description was generated automatically by the task (e.g. a CIVET task puts the name of the mnc file in the description by default, and that's useful for user, but if the user loads a preset now, they lose that information)

prioux: I'd add the preset's description AFTER the original description, and with a newline added, but only of the original description is non-blank. Shoudl be possible in one line.

MontrealSergiy: might be useful, yet even config name is overwritten during merge

natacha-beck: @prioux, regarding the following: 

>e.g. a CIVET task puts the name of the mnc file in the description by default, and that's useful for user, but if the user loads a preset now, they lose that information

--> Civet include this information only if the description is empty, no ?

natacha-beck: > might be useful, yet even config name is overwritten during merge. For greater consistency maybe concatenate the config names (versions) too?

@MontrealSergiy Which config name ?

MontrealSergiy (changes requested):
I think that when the original description is empty, this will create description starting from two line break, which against Pierre's recommendations. `join` or other hack could help keep logic in one line.

Also can descriptor be nil or whitespace etc?

BrainPortal/app/controllers/tasks_controller.rb
MontrealSergiy: sorry, Natacha, that was not well thought through, forget that comment

BrainPortal/app/controllers/tasks_controller.rb
```diff
@@ -1502,6 +1502,7 @@ def handle_preset_actions #:nodoc:
       if (! preset_id.blank?) && preset = CbrainTask.where(:id => preset_id, :status => [ 'Preset', 'SitePreset' ]).first
         old_params = @task.params.clone
         @task.params         = preset.params
+        @task.description   += @task.description.blank? ? "#{preset.description}" : "\n\nDescription from preset configuration:\n\n#{preset.description}"
```
prioux, line 1505: This will crash if @task.description is nil

prioux: This is ok except in the case where the preset doesn't have a description. Then the result is that we get just 'Description from preset configuration'

prioux:
I can't believe it took 7 commits for that.

PR #1283, opened 2022-11-15
CBRAIN public checkbox 1244 

MontrealSergiy:
Adds public checkbox to  groups

### DIFF
--- a/BrainPortal/app/views/groups/new.html.erb
+++ b/BrainPortal/app/views/groups/new.html.erb
@@ -54,18 +54,17 @@
       <%= f.check_box :track_usage %>
     <% end %>
 
-    <% Make project public, so its files are visible to all the users %>
-    <%= f.check_box :public %>
-    <p>
-    A public project makes all its files visible to all the users. Be careful with this option!
-    You can always make project to public latter on.
-
     <p>
     Normal members will not be able to assign files or other resources
     to this project (but editors are always allowed to do so):
     <%= f.check_box :not_assignable %>
 
-    <% if current_user.has_role?(:normal_user) %>
+    <p>
+    Make public, so all user can access it. Be <b> careful </b> with this option! You can always make project to public latter on:
+    <%= f.check_box :public %>
+    <p>
+
+      <% if current_user.has_role?(:normal_user) %>
       <%= hidden_field_tag("group[user_ids][]", current_user.id.to_s) %>
     <% else %>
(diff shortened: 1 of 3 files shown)


BrainPortal/app/views/groups/new.html.erb
```diff
@@ -54,6 +54,12 @@
       <%= f.check_box :track_usage %>
     <% end %>
 
+    <% Make project public, so its files are visible to all the users %>
```
prioux, line 57: I've not looked at how the form is rendered, but is this layout correct? In the original form, we had the explanations before the inputs, and you added your explanation after your input.

prioux (changes requested):
See comment

BrainPortal/app/views/groups/new.html.erb
MontrealSergiy: corrected

BrainPortal/app/views/groups/new.html.erb
```diff
@@ -59,7 +59,12 @@
     to this project (but editors are always allowed to do so):
     <%= f.check_box :not_assignable %>
 
-    <% if current_user.has_role?(:normal_user) %>
+    <p>
+    Make public, so all user can access it. Be <b> careful </b> with this option! You can always make project to public latter on:
```
prioux, line 63: Fix text, you're missing articles: `Make the project public, so that all users can access the files. Be careful with this option! You can always make the project public later on.`

prioux (changes requested):
Text change

BrainPortal/app/views/groups/new.html.erb
MontrealSergiy: `all _the_ user` I guess ? are any other articles or grammar issues missed?

prioux:
Sergiy, you actually inserted an article where there shouldn't be one, and you did not modify the rest of the sentence like I suggested. Can you please review my comment again? There were several changes to make.

MontrealSergiy:
Ok, done. @prioux do you suggest to end the sentence with the full stop or semicolon, as in the rest of the page?

prioux:
Ah don't mind all that much about the colon or periods. At some point we'll go over the entire interface and just adjust all the description sentences to make them all consistent, once we know what we want. Thanks for the fix.

prioux:
At this point the code is fine, I'll test it and merge soon.

PR #1284, opened 2022-11-17
[ADD] improve nh_loris_hook and add extended CBCsv file

natacha-beck:


BrainPortal/app/controllers/nh_loris_hooks_controller.rb
```diff
@@ -64,21 +64,70 @@ def file_list_maker
...
+
+      parent_dir
+    end.uniq
+
     userfiles = base.where(:name => source_basenames)
     userfiles = userfiles.where(:data_provider_id => s_dp.id)    if s_dp
     userfiles = userfiles.where(:group_id         => s_group.id) if s_group
 
+    # Add setter and getter for json_params in Userfile class
+    Userfile.class_eval do
```
prioux, line 97: You shouldn't do that. You're modifying the Userfile class for ALL classes and subclasses.

What you want is defined object methods JUSt on the objects that are you pass along to the ExtendedCbrainFileList constructor.

BrainPortal/app/controllers/nh_loris_hooks_controller.rb
```diff
@@ -64,21 +64,70 @@ def file_list_maker
...
-    exp_count  = source_basenames.size
+    file_count  = found_names.size
+    exp_count   = source_basenames.size
     if (file_count == 0) || (strict_match && (file_count != exp_count))
       cb_error "Could not find an exact match for the files. Found #{file_count} of #{exp_count} files"
     end
 
     # Create CbrainFileList content and save it to DP
-    cblist_content = CbrainFileList.create_csv_file_from_userfiles(userfiles)
+    extra_params_by_parent_is_empty = true
```
prioux, line 122: What is this for?

prioux: Use the array method `any?()` instead!

prioux:
```ruby
[1,2,3,4,5].any? { |x| x == 3 } # returns tue or false, and also stops as soon as it can detect a true condition
```

BrainPortal/app/controllers/nh_loris_hooks_controller.rb
```diff
@@ -64,24 +64,70 @@ def file_list_maker
     s_group = Group.where_id_or_name(source_group_id).first if source_group_id
     base = Userfile.where(nil) # As seen in userfiles_controller
     base = Userfile.restrict_access_on_query(current_user, base, :access_requested => :read)
+
```
prioux, line 67: This entire block should be moved into a new private method.

It should only return the bare information about the paths as a hash: e.g. { 'sub-1234' => [ 'anat/blah1_t1.nii.gz', 'anat/blah2_t1.nii.gz' ], 'sub-2345' => [], 'sub-33221' => [...] }

BrainPortal/app/controllers/nh_loris_hooks_controller.rb
prioux: Use singleton methods

prioux: The creation of the method shoudl be performed by a utility method in the ExtendedCbrainFileList class:

```ruby
  # in ExtendedCbrainFileList
  def self.extend_userfile_list(userfiles, extra_info = {}) ; extra info is { basename_or_uid? => RUBYTHING }
     #add singleton method :json_params to return RUBYTHING
  end

File: BrainPortal/app/controllers/nh_loris_hooks_controller.rb [hunk 4]
```diff
@@ -64,24 +64,70 @@ def file_list_maker
...
+        @json_params = arg.to_s
+        Userfile.columns_hash['json_params'] = OpenStruct.new({type: :hash})
+      end
+
+      define_method :json_params do
+        @json_params || nil
+      end
+    end
+
+    is_extended = false
```
prioux, line 108: Again, why not simply use `any?()` instead?

natacha-beck:
I will close this PR since it is replaces by this [one](https://github.com/aces/cbrain/pull/1286).

PR #1286, opened 2022-11-22
ExtendedCbrainFileList and Boutiques portal use extra fields in ExtendedCbCSV file.

natacha-beck:
Handle extra parameter in CbrainFileList. 
Porpagate extra parameter to individual task whan ExtendedCbrainFileList is used to launch multiple task.

### DIFF
--- a/BrainPortal/app/models/boutiques_portal_task.rb
+++ b/BrainPortal/app/models/boutiques_portal_task.rb
@@ -322,9 +322,18 @@ class BoutiquesPortalTask < PortalTask
 
     # Array with the actual userfiles corresponding to the cbcsv
     mapCbcsvToUserfiles = cbcsvs.map { |f| f[1].ordered_raw_ids.map { |i| (i==0) ? nil : i } }
+    # Array with the actual params as Hash corresponding to the cbcsv
+    mapCbcsvToParams    =  cbcsvs.map do |f|
+                             cbcsv = f[1]
+                             cbcsv.is_a?(ExtendedCbrainFileList) ?
+                              cbcsv.ordered_params : [nil]
+                            end
+
     # Task list to fill and total number of tasks to output
-    tasklist = []
-    nTasks   = mapCbcsvToUserfiles[0].length
+    tasklist   = []
+    nTasks     = mapCbcsvToUserfiles[0].length
+    input_keys = descriptor.inputs.map{|input| input["id"]}
+
     # Iterate over each task that needs to be generated
     for i in 0..(nTasks - 1)
       # Clone this task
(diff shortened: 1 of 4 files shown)


BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -335,6 +344,8 @@ def final_task_list #:nodoc:
         #currTask.params[:interface_userfile_ids] << mapCbcsvToUserfiles unless currId.nil?
         currTask.invoke_params[cinput.id] = currId # If id = 0 or nil, currId = nil
         currTask.invoke_params.delete(cinput.id) if currId.nil?
+        params     = mapCbcsvToParams[j][i] || {}
```
prioux, line 347: don't call this local variable 'params'.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/extended_cbrain_file_list/extended_cbrain_file_list.rb
```diff
@@ -0,0 +1,112 @@
...
+  #
+  # cached_csv_array, ordered_raw_ids, userfiles_accessible_by_user and extra_params!
+  def flush_internal_caches
+    @ids_with_zeros_and_nils = nil
+    @userfiles               = nil
+    @rows                    = nil
+    @extra_params            = nil
+  end
+
+  # "a/b/c" -> {"a" => ["a/b/c", "a/d/f"]}
```
prioux, line 84: adjust comment, it's still not showing off that the argument is an array and where the d/f part comes from

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -293,11 +295,24 @@ def final_task_list #:nodoc:
...
           fillTask.( f, task )
-
         else # One task per userfile in the CbrainFileList
-          ufiles = f.userfiles_accessible_by_user!( self.user, nil, nil, file_access_symbol() )
-          # Skip files that are purposefully nil (e.g. given id 0 by the user)
-          subtasks = ufiles.select { |u| ! u.nil? }.map { |u| fillTask.( u, self.dup ) }
+          ufiles               = f.userfiles_accessible_by_user!( self.user, nil, nil, file_access_symbol() )
+          ordered_extra_params = f.is_a?(ExtendedCbrainFileList) ? f.ordered_params : []
+
+          # Remove files that are purposefully nil (e.g. given id 0 by the user)
```
prioux, line 302: Line 302 to 308 seems to be code that identify the null entries in ufiles and remove the entries in the ordered_extra_params accordingly.

But why is this necessary? Can we just have a "next" statement in the loop that's at line 312 instead just to skip over these entries?

PR #1291, opened 2023-01-05
Clean bids

natacha-beck:
Add module to clean BIDS subject when some extra files need to be deleted before running a tool.


natacha-beck:
Not relevant correspond to: PR https://github.com/aces/cbrain/pull/1286

PR #1298, opened 2023-01-20
[ADD] launching task arrays by selecting multiple files in New Boutiques

### LINKED ISSUE #1201: Re-implement launching task arrays by selecting multiple files in New Boutiques
The old boutiques integrator allowed a user to launch a set of tasks by selecting a bunch of files and leaving one file input blank; multiple tasks were launched and each of them got one file in that input.

The new boutiques integrator forces the user to create a CbrainFileList. I did it like that because I was re writing the launch code form scratch, but I think getting the old behavior to work too would be useful.

If a task has several file inputs, only one can be left blank, of course.

Make sure the 'preview' button's code says the right thing too.

natacha-beck:
Closed #1201


prioux (changes requested):
Double check all his, I really am not sure it works.

natacha-beck:
> Double check all his, I really am not sure it works.

It was tested and seems to work properly when I did the PR. 
I renamed some variable and remove the `next`. I do not see the point of the `next` neither. 

I tested with multiple files selected (3 files selected as inputs --> Produce 3 tasks), I also verify that using a CbrainFileList work ok too.

prioux:
The original 'next nil' statement was meant to replace in-situ a bad userfile ID in a File input.

Your code no longer does that.

Also, instead of looking up each and every file with a `find_accessible_by_user`, why don't you get them all in one shot? That way if you have 1000 file entries you're not going top make 1000 requests to the database.

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -640,4 +642,10 @@ def file_access_symbol
     @_file_access ||= (self.class.properties[:readonly_input_files].present? || self.tool_config.try(:inputs_readonly) ? :read : :write)
   end
 
+  private
+
+  def with_single_file_input
+    self.descriptor_for_form["inputs"].select{|x| x["type"] == "File"}.size == 1
```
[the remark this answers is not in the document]
prioux: Also, name it with a trailing `?` maybe?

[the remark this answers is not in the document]
prioux: Also, don't use the hash syntax for accessing the descriptor's inputs, it's a proper object with properly named access methods (e.g. `.inputs`)

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -640,4 +643,10 @@ def file_access_symbol
     @_file_access ||= (self.class.properties[:readonly_input_files].present? || self.tool_config.try(:inputs_readonly) ? :read : :write)
   end
 
+  private
+
+  def single_file_input?()
+    @single_file_input ||= self.descriptor_for_form.inputs.select{|x| x["type"] == "File"}.size == 1
```
prioux, line 649: 1. why do you declare the method with `()` on the `def` line ?
2. You cannot cache a true/false value with a `||=` because in the case of false, you will recompute the expression every time.
3. Use the methods for the inputs! `x.type` instead of `x["type"]` 
4. You use select() to create a temporary array of all the inputs of type File, and then all you do is check if the array has size `1`. Then the array is discarded. That's a waste of memory. Arrays have a count() method that you can pass a block to:
``ruby
desc.inputs.count { |x| x.type == 'File' } == 1
```

File: BrainPortal/app/models/boutiques_portal_task.rb [hunk 3]
```diff
@@ -522,21 +523,27 @@ def sanitize_param(input)
       # Make sure the file ID is valid, accessible, not already used and
       # of the correct type.
       when :file
-        unless (Integer(value) rescue nil)
+        if !(Integer(value) rescue nil) && !single_file_input?
```
prioux, line 526: I don't understand why the fact we have a single file input is important when validating that a file's value is an integer? It shoudl always be true, no?

prioux (changes requested):
I think this entire change set is wrong. The code modified here is mostly in the method sanitize_params(), which is a validation method, not a method supposed to actually perform behavioral actions like assigning input files. It's just supposed to verify them. 

Please reconsider the whole thing, it's sanitize_params is not the right place to implement this feaure.

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -758,8 +758,16 @@ def boutiques_module_information #:nodoc:
     integrator_modules.map do |module_name, _|
       module_name = module_name.constantize
       rev_info    = module_name::Revision_info
-      "#{rev_info.basename} rev. #{rev_info.short_commit} #{rev_info.time} (author: #{rev_info.author})"
+      rev_info.format("%f rev. %s %a %d")
```
prioux, line 761: As far as I know this change is already in the codebase, maybe you included commits that are not related to this PR?

BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_boutiques_input.html.erb
```diff
@@ -78,16 +78,21 @@
...
-            { :id        => input.cb_invoke_html_id(false), # false disable list input's '[]'
-              :name      => input.cb_invoke_html_name(false), # false disable list input's '[]'
-              :options   => interface_files.map { |userfile|  [ userfile.id, userfile.browse_name ] },
-              :optional  => opt,
-              :default   => interface_files.count == 1 ? interface_files.first.id.to_s : nil,
-              :value     => value,
-              :none_msg  => '(Nothing to select)', # not used
-            }
-      %>
+      <%- if @descriptor.inputs.count { |x| x.type == 'File'            } == 1 &&
```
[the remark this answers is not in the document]
prioux: Actually the entire condition is

```
IF only_one_input AND that_input_is_not_a_list
```

So you don't have to iterate at all... `@descriptor.file_input.first.list` will be true or false already.

prioux (changes requested):
Review comments

prioux:
Please review my comments.

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -72,6 +72,11 @@ def descriptor_for_show_params
     self.boutiques_descriptor
   end
 
+  # Return true if the descriptor have only one mandatory input
+  def have_a_single_mandatory_input_file?(descriptor)
+    descriptor.file_inputs.select{|i| i.optional == false }.size == 1
```
prioux, line 77: Don't check with an explicit comparison `something == false` because that fails for nil. Try it in the console. There are several better ways of doing your check:

```ruby
descriptor.file_inputs.select { |i| ! i.optional }.size == 1
descriptor.file_inputs.reject { |i|   i.optional }.size == 1
```

and the best, using a block for count():

```ruby
descriptor.file_inputs.count { |i| ! i.optional } == 1
```

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -113,14 +118,19 @@ def before_form
...
+    if (num_in_files < num_needed_inputs)
       message = "This task requires #{num_needed_inputs} mandatory file(s) and #{num_opt_inputs} optional file(s)\n" +
         input_infos
       cb_error message
     end
 
-    #return "Warning: you selected more files than this task requires, so you won't be able to assign them all." if
-    #  num_needed_inputs > num_needed_inputs+num_opt_inputs
+    # Verify the number of files selected in case of a single mandatory file input or descriptor without a file input list
+    if ((!have_a_single_mandatory_input_file?(descriptor) && !have_file_input_list) && num_in_files > num_needed_inputs + num_opt_inputs)
```
prioux, line 129: I don't understand, where does have_file_input_list come from? Is this a method, a variable? I don't know where it's defined?

prioux (changes requested):
Review not completed, problems identified.

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -773,6 +808,26 @@ def generate_placeholder_descriptor
 
   private
 
+  # Return id of the uniq mandatory file
+  def uniq_mandatory_input_file_id
```
prioux, line 812: Many of these methods should be added to the BoutiquesSupport::BoutiquesDescriptor class. And like instead of returning true/false or an ID, they shoudl return that thing that was (searched for)/(detected) or nil. E.g. `have_a_file_input_list` should be

```ruby
# in class BoutiquesSupport::Descriptor
def mandatory_file_inputs # note: plural
  file_inputs.reject { |input| input.optional }
end

def sole_mandatory_file_input  # note: singular!
  all = mandatory_file_inputs
  return all[0] if all.size == 1
end

def has_one_mandatory_file_input? # note: can have multiple other file inputs
   sole_mandatory_file_input.present?
end

def has_one_file_input_and_it_is_mandatory
   file_inputs.size == 1 && sole_mandatory_file_input # returns it too
end

def file_inputs_that_are_lists
  file_inputs.select { |input| input.list }
end
```

By returning the THINGS you dcan both check the situation, and use them in the main code.

prioux:
Overall I think the modification are still way too complex and the small utility methods are not named properly, and not defined in the proper classes (see my comments).

prioux (changes requested):
See review comments

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -773,6 +788,15 @@ def generate_placeholder_descriptor
 
   private
 
+  # Return remaining file ids for the unique mandatory file descriptor input
+  def ids_for_uniq_mandatory_file(descriptor)
+    used_file_ids   = []
+    descriptor.file_inputs.select{|x| x.optional == true }.map(&:id).each do |input_id|
```
prioux, line 794: We have a method `optional_file_inputs` in the descriptor model.

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -304,8 +310,17 @@ def final_task_list #:nodoc:
         tsk
       end
 
-      original_userfiles_ids = self.params[:interface_userfile_ids].dup
+
```
prioux, line 313: This entire block of changes (lines 313-320) is REALLY hard to read. You use the long ruby expression

```ruby
self.invoke_params[descriptor.sole_mandatory_file_input.id]
```

three times, why don't you extract it in a variable first? Also, why not leave the original line (line 307 in the original code)

```ruby
  original_userfiles_ids = self.params[:interface_userfile_ids].dup
```

exactly as it was, and THEN crush the value of original_userfiles_ids if needed? That would simplify the if/else block

BrainPortal/app/models/boutiques_portal_task.rb
prioux: Also, in general in Ruby, if you wrute code that has the overall structure

```ruby
variable = [] # or {}
some_loop.each do |xyz|
  variable << something
end
```

then there is probably a better way to do this. The Ruby iterators can return values. In the current case, you could have written more simply as

```
user_file_ids = descriptor.optional_file_inputs.map do |input|
  Array(self.invoke_params[input.id])
end.flatten
```

Also, last note about your code (this is made obsolete by my suggestion above), you did a ".flatten" in your last statement but in your original code, in your loop, you could have avoided it by using `used_file_ids |= blahblah` instead of `+=`

prioux (changes requested):
Minor modifications. Adjustements to code

Issue #1281, opened 2022-11-03
Create new BTQ integrator option to force a full copy of input files


prioux:
By default, CBRAIN tasks create symbolic links to the CBRAIN DP cache for all the input files that the task need; that way inputs can be shared by multiple task without too much data duplication.

The symbolic links are created by the make_available() utility method in class ClusterTask.

Some stupid tasks though MUTATE their own input data, so creating a symlink doesn't work anymore, the input data of one task would perturb the input data of another one directly in the cache area.

I'd like a new option for make_available() where instead of creating a symlink, the method would make a full copy of the input data directly in the task workdir.

I'd also like this option to be invoked when a Boutiques descriptor for the new integrator contains a statement about making a full copy. This shouldn't be implemented as a module, but as an option in the `custom` section of the descriptor.

natacha-beck:
Could be closed par part of the PR #1452

prioux:
@natacha-beck Indeed. But then I'm not sure about #1299 ?

prioux:
Done from PR #1452

PR #1299, opened 2023-01-27
flag to copy input data to work folder

MontrealSergiy:
This to reduce corruption of input data and race conditions with input mutating scientific tools, see #1281 for description

### DIFF
--- a/BrainPortal/app/models/boutiques_cluster_task.rb
+++ b/BrainPortal/app/models/boutiques_cluster_task.rb
@@ -83,7 +83,8 @@ class BoutiquesClusterTask < ClusterTask
 
       # Most common situation
       if ! input.list || ! userfile.is_a?(CbrainFileList)
-        make_available(userfile, userfile.name)
+        copy_file = descriptor.custom.dig('cbrain:full-copy-input-files')&.include?(input.id)
+        make_available(userfile, userfile.name, nil, nil, copy_file)
         next
       end
 
@@ -91,7 +92,8 @@ class BoutiquesClusterTask < ClusterTask
       userfile.sync_to_cache
       userfile_list = userfile.userfiles_accessible_by_user!(user, nil, nil, file_access_symbol)
       userfile_list.compact.each do |subfile|
-        make_available(subfile, subfile.name)
+        copy_file = descriptor.custom.dig('cbrain:full-copy-input-files')&.include?(input.id)
+        make_available(userfile, userfile.name, nil, nil, copy_file)
       end
     end
 
--- a/BrainPortal/app/models/boutiques_cluster_task.rb
+++ b/BrainPortal/app/models/boutiques_cluster_task.rb
@@ -81,9 +81,11 @@ class BoutiquesClusterTask < ClusterTask
       next if userfile_id.blank? # that happens when it's an optional file
       userfile    = Userfile.find(userfile_id)
 
+      # this flag indicates should one
+      copy_file = descriptor.custom.dig('cbrain:full-copy-input-files')&.include?(input.id)
+
       # Most common situation
       if ! input.list || ! userfile.is_a?(CbrainFileList)
-        copy_file = descriptor.custom.dig('cbrain:full-copy-input-files')&.include?(input.id)
         make_available(userfile, userfile.name, nil, nil, copy_file)
         next
       end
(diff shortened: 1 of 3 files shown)


MontrealSergiy:
test boutiques at https://github.com/MontrealSergiy/cbrain-plugins-test/blob/main/boutiques_descriptors/copy_flag_demo.json

prioux:
I don't really want a new attribute in the tool config structure, rather I want a simple implementation that just copies files when make_available is called with a new option:

```ruby
def make_available(userfile, file_path, userfile_sub_path = nil, start_dir = nil, copy_files = false)
```

On the Boutiques descriptor side, all I want is support for:

```json
    "custom": {
        "cbrain:full-copy-input-files": [ "inputid1", "inputid2", "inputid3" ],
```

prioux (changes requested):
See my previous comment.

* Remove all tool config changes
* Add copy_files option to make_available()
* Adjust BTQ integrator to set copy_files to true when calling make_available

MontrealSergiy:
> * BTQ integrator to set copy_files

Note that I see several uses of make_available, besides the BTQ integrator e.g. i
such as boutiques_input_subdir_maker, bids_app_maker

Should I address them too? Would dev continue to use it in modules ?

probably, it safer to pull property automatically, maybe yet another parameter copy_input_as_in_boutiques?


should I add copy_input there too?

MontrealSergiy:
Looks like rebase gone wrong still working

BrainPortal/app/models/boutiques_cluster_task.rb
```diff
@@ -83,15 +83,17 @@ def setup #:nodoc:
 
       # Most common situation
       if ! input.list || ! userfile.is_a?(CbrainFileList)
-        make_available(userfile, userfile.name)
+        copy_file = descriptor.custom.dig('cbrain:full-copy-input-files')&.include?(input.id)
```
prioux, line 86: You are extracting this option twice, once at line 86 and once at line 95, and at line 95 you are also doing it within an internal loop. This flag doesn't change over time, extract it just once for the input, at around line 83!

BrainPortal/app/models/cluster_task.rb
```diff
@@ -485,13 +488,21 @@ def make_available(userfile, file_path, userfile_sub_path = nil, start_dir = nil
...
-      # catch the error in the situation.
+
+      # copying code
+      return userfile.cache_copy_to_local_file(full_path) if copy_file
+      # presently we do not save what (user)file is copied so no exception is raised
+      # if file or link exists already
+
+      # symlinking code
       unless File.symlink?(file_path.to_s) && File.readlink(file_path.to_s) == target.to_s
+        # Do nothing is symlink already exists with proper value.
```
prioux, line 502: You moved the comment block that explains the logic of the `unless` statement UNDER the unless statement, why? That makes the code more difficult to understand.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -485,13 +488,21 @@ def make_available(userfile, file_path, userfile_sub_path = nil, start_dir = nil
...
+
+    # Create the symlink or copy input data
     Dir.chdir(self.full_cluster_workdir) do
-      # Do nothing is symlink already exists with proper value.
-      # If there is something not a symlink in the way, or a symlink with a different
-      # value, the symlink() method will crash, which is what we want to
-      # catch the error in the situation.
+
+      # copying code
+      return userfile.cache_copy_to_local_file(full_path) if copy_file
```
prioux, line 496: This doesn't work if the caller of make_available() has specified a `userfile_sub_path` in argument. You're only supporting the case where we need a full copy of a FileCollection, but the not case where we need a copy of a partial subtree/subfile.

prioux:
Please also look at supporting this option in the original Boutiques integrator. This code only supports the new integrator. If it's too complicated, let me know. We don't use the old integrator anymore, so maybe it's not worth it.

prioux (changes requested):
Bugs and adjustements.

MontrealSergiy:
rebased

prioux:
I am closing this PR without merging because anyway we now have a modules that does this sort of thing.

PR #1300, opened 2023-01-31
Ext3 capturing of files in cbrain tasks

prioux:
Adds a new feature. Admins can configure that a subdirectory in a task's work directory will be bound to a mounted ext3 filesystem-in-a-file. In the ToolConfig object, in the "overlays" section, this is done with directives like:

ext3capture:work=40G

which will create an empty 40 Gb file `.capt_work.ext3` in the work directory, format it as ext3, and then when the singularity container is started, the directory `{WORKDIR}/work` will be used as a mountpoint.

As an additional feature, without modifying the ToolConfig, a designer of a boutiques descriptor can add a similar specification in the 'custom' section (see BoutiquesExt3Capturer in this PR).

### DIFF
--- a/BrainPortal/app/models/cluster_task.rb
+++ b/BrainPortal/app/models/cluster_task.rb
@@ -689,7 +689,7 @@ class ClusterTask < CbrainTask
       self.make_cluster_workdir
       self.apply_tool_config_environment do
         Dir.chdir(self.full_cluster_workdir) do
-          setup_success   = self.meta[:submit_without_setup] || self.setup
+          setup_success = self.meta[:submit_without_setup] || self.setup
           if ! setup_success  # as defined by subclass
             self.addlog("Failed to setup: 'false' returned by setup().")
             self.status_transition(self.status, "Failed To Setup")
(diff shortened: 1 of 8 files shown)


natacha-beck (approved):
Tested on my end with my local Bourreau works fine.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2330,17 +2330,27 @@ def singularity_commands(command_script)
...
-    cache_dir     = self.bourreau.dp_cache_dir
+    # (2) The root of the DataProvider cache
+    cache_dir      = self.bourreau.dp_cache_dir
+    gridshare_dir  = self.bourreau.cms_shared_dir # not mounted explicitely
+    cache_dir_syml = "#{gridshare_dir}/DP_Cache"
 
-    # (6) The path to the task's work directory
-    task_workdir  = self.full_cluster_workdir
+    # (3) Ext3 capture mounts, if any.
+    # These will look like "-B .capt_abcd.ext2:/path/workdir/abcd:image-src=/"
```
MontrealSergiy, line 2345: are you sure it is ext2 not ext3? (Not like I hold a grudge toward ext2 or ext4, who knowns, they might work as well, it is just everywhere else you say ext3

prioux: It's an error in the comment. It's ext3. ext2 is not supported by Singularity anyway.

PR #1304, opened 2023-02-13
Updating the list of users in a workgroup (Bug1+Bug2)

natacha-beck:
I was annoyed by this issue and couldn't ignore it.

The fix concerns bug 1 for the moment:

> the users already members of the project don't have their checkboxes pre-selected in table A

**Note**: the `select_all_box` is also used in `bourreaux/rr_disk_usage`. 
Seems to be a little bit buggy too in rr_disk_usage when you: 
- select_all a row then select_all a column (see screenshot 1)
- then unselect the row did not unable the all column (see screenshot 2).

image

Then

image

### DIFF
--- a/BrainPortal/app/helpers/dynamic_form_helper.rb
+++ b/BrainPortal/app/helpers/dynamic_form_helper.rb
@@ -49,9 +49,12 @@ module DynamicFormHelper
   # Create a checkbox that will select or deselect all checkboxes on the page
   # of class +checkbox_class+.
   # Most +options+ are just treated as HTML attributes.
-  # Except +options[:persistant_name]+; if provided,
-  # an additional hidden input will be added to track
-  # the state of the select_all checkbox.
+  #
+  # Except:
+  #   - +options[:persistant_name]+: if provided, an additional hidden
+  # input will be added to track the state of the select_all checkbox.
+  #   - +options["data-noload"]+: if true, the select_all checkbox will
+  # not be updated when the page is loaded.
   def select_all_checkbox(checkbox_class, options = {})
     options[:class] ||= ""
     options[:class]  +=  " select_all"
--- a/BrainPortal/app/helpers/dynamic_form_helper.rb
+++ b/BrainPortal/app/helpers/dynamic_form_helper.rb
@@ -58,7 +58,6 @@ module DynamicFormHelper
 
     options["data-checkbox-class"] = checkbox_class
 
-
     options["data-noload"] = "true" if options["data-noload"]
     atts  = options.reject{|x| x.to_s === "persistant_name"}.to_html_attributes
 
--- a/BrainPortal/app/helpers/dynamic_form_helper.rb
+++ b/BrainPortal/app/helpers/dynamic_form_helper.rb
@@ -49,7 +49,7 @@ module DynamicFormHelper
   # Create a checkbox that will select or deselect all checkboxes on the page
   # of class +checkbox_class+.
   # Most +options+ are just treated as HTML attributes.
-  # Except +options[:persistant_name]+; if provided, 
+  # Except +options[:persistant_name]+; if provided,
   # an additional hidden input will be added to track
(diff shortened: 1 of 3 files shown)


natacha-beck:
@MontrealSergiy please have a look to the following fix and test it.

MontrealSergiy:
@natacha-beck Hi Natacha, will do could you please explain why do you need emulate load event at all? I fail to follow, is it for some special cases of "persistent" check-boxes?

prioux:
Not only that, this new option is secret, it's not even documented in the usage statement of

```ruby
def select_all_checkbox
```

MontrealSergiy:
On bright side Bug1 addressed - current users shown after the click on the edit (though select_all boxes are not)

MontrealSergiy:
for new issue you mentioned a quick fix could be use on change even instead on click. Yet I bet recomputing everything might be faster in some corner cases

MontrealSergiy (approved):
seems to work (the issue addressed) not sure is solution optimal

natacha-beck:
The fact that we called click_select_all on load made the checkbox of the active user (table) dependant of the status of WorkGroup/SiteGroup selection. 
That's was the reason why no user was selected (previous member of the group was ignored), and why I introduce the noload option. 

Seems to work properly I will have a closer look, the logic here seems strange when it apply to the project group. Maybe due to a mix between user information and WorkGroup/SiteGroup information.  That confirm @prioux suspicion about: 
> I suspect there is a confusion about the IDs and classes associated with the normal checkboxes in table A and the master select checkboxes in table B.

MontrealSergiy:
@natacha-beck I was just wondering if you must at all, why do you, on load, define subordinate from master and not vise versa? Are there any special use cases? Perhaps related to persistence?

MontrealSergiy:
I do not think Bug 2 reported Pierre is strange or some id confusion. It follow the logic you implemented (masters mirror the state of subordinates). Well, the issue you mentioned in the beginning of PR is most likely also present on group though less anoying than on that rr_disk_usage table

MontrealSergiy:
A side note with your changes master-slave terminology probably does not apply any more. Maybe Leader/Group Lead checkbox? Representative checkbox? Delegate checkbox? Elder checkbox? Head? Principal? Chieftain? Foreman? :)

prioux:
There is too much talking here. I just want a solution. Make it all work like before the bug occurred. No programming shenanigans. Keep things simple. The select_all_checkbox() helper was committed by Tarek in Nov 2011, I don't want programmers to no longer be able to use it as it was designed.

MontrealSergiy:
Natacha groups Bug 1 seem addressed, not sure how macaques should work never tried it before. Will install and test tomorrow. The "bid app handler" seems suspicious, not sure how other checkboxes might be affected.  Please mention in method comment/docs that checkboxes will be reinitialized to master value. Maybe keep copy of old branch we do not which would be preferred

natacha-beck:
It fix both bug.

MontrealSergiy:
Confirm Bug 1 and so-called Bug 2 addressed.

MontrealSergiy:
the difference in behaviour for persistent case might be documented better ( the fact that for persistant case you reinit subordinate checkboxes to select all box value, even if they already have checked property still suprizing, best to mention if not explain)

MontrealSergiy:
you mean Bug1 Bug2 or something else'

MontrealSergiy:
do you use safari or chrome?

prioux:
I've not even gotten to bug 2 yet. As soon as I open any workproject in CBRAIN, the users currently in the project are not pre-selected. Why did you guys say this was fixed?

prioux:
Firefox.

MontrealSergiy:
refresh? Ctrl F5

prioux:
I think Firefox caches Javascript libraries much more aggressively than other browsers

prioux:
My sincere apologies!

Issue #1306, opened 2023-02-27
Index table filters trigger the first form in the page when hitting 'return' in the search box


prioux:
A POST to a controller is performed when it shouldn't happen.

To reproduce:

1) Go the an index page such as `/tasks`
2) Click on the small magnifying glass for one of the columns (e.g. the Task Type column)
3) Enter dummy text in the search box for the column (e.g. 'blah')
4) Hit return

In the case of the `/tasks` page, a request is posted to `TasksController#update_multiple` using the first hijacker submit button in the 'Update Attributes' button/menu.

See screenshots:

![Screen Shot 2023-02-27 at 13 56 12](https://user-images.githubusercontent.com/777588/221658094-32cd0516-be45-4cf9-a7da-8d28dc9c8a15.png)

![Screen Shot 2023-02-27 at 13 57 17](https://user-images.githubusercontent.com/777588/221658123-e2c4ae85-0309-42e3-b772-36b33d7a80bc.png)

prioux:
Solution: completely disable submit on `Enter` in these little search boxes, the return key is not supposed to do anything.

prioux:
The Rails server logfile show the request that was sent:

```
Processing by TasksController#update_multiple as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"[FILTERED]", "cbrain_task"=>{"group_id"=>"", "results_data_provider_id"=>"27", "tool_config_id"=>""}, "commit"=>"Update", "dup_bourreau_id"=>"21", "archive
_dp_id"=>"", "per_page"=>"25", "update_group_id"=>"1"}
```

MontrealSergiy:
temporary withdrow

prioux:
Fixed by #1307

PR #1307, opened 2023-02-28
fixes bug when enter in table find causes  post fixes #1304

MontrealSergiy:
for details see Index table filters trigger the first form in the page when hitting 'return' in the search box #1306


MontrealSergiy:
as usually needs refresh

MontrealSergiy:
Note if browser cached need to expire automatically 1) "auto-versioning" hack can be used https://stackoverflow.com/questions/118884/how-to-force-browsers-to-reload-cached-css-and-js-files either on rails app side or with a server module/plugin 2) looking forward cache expiration can be reduced for few hours

prioux:
I really don't mind if browser caching JS script cause an issue, that solves itself after some time anyway.

If this JS change fixes the problem, I'm satisfied!

prioux:
To test: what happens if a user hits SHIFT-RETURN, or CTRL-RETURN, or other special return keymaps?

MontrealSergiy:
I noted though that some people block code 10 (linefeed), I can block it just in the case. I am not sure 100% for touchscreen/ trackpad, but few people use cbrain on mobile and I hope jquery handles non-standard input devices intelligently.

prioux:
Yes, I'd add a check for LF (10) too.

Have a bit of a loo, try things yourself, until you're pretty satisfied you can't trigger the bug anymore no matter how much you try. The fix is pretty simple, at this point, so when you tell me you're confident enough about it I'll merge it.

MontrealSergiy:
A slightly more elegant to test for `enter`, `return` key and friends without explicit enumeration . See https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

To be sure that there are no spuriously hijacking elsewhere something else could be done. Seems good enough for merging to me though.

natacha-beck (approved):
Tested on my side, works fine.

MontrealSergiy:
For example, to guaranty absence of hijacking throughout the cbrain one can indicate clearly which form  each input belong using appropriate attributes see https://www.w3schools.com/html/html_form_attributes_form.asp

If needed sink (that is 'fake') forms can be created. Additionally to above js.

PR #1312, opened 2023-03-20
[ADD] Use data_link helper for qc_panel

natacha-beck:
Use the Subviewer mechanism in order to display files in QC_panel if a viewer is available.

### DIFF
--- a/BrainPortal/app/helpers/userfiles_helper.rb
+++ b/BrainPortal/app/helpers/userfiles_helper.rb
@@ -79,7 +79,7 @@ module UserfilesHelper
 
   # Generates links to pretty file content for files inside FileCollections.
   # Generates a download link if no viewer code can be found for the files.
-  def data_link(file_name, userfile)
+  def data_link(file_name, userfile, replace_div_id="sub_viewer_filecollection_cbrain")
     full_path_name  = Pathname.new(userfile.cache_full_path.dirname + file_name)
 
     display_name  = full_path_name.basename.to_s
@@ -87,9 +87,10 @@ module UserfilesHelper
 
     file_lstat = full_path_name.lstat  # lstat doesn't follow symlinks, so we can tell if it is one
 
-    return h(display_name) unless file_lstat.file?
+    return h(display_name) if !file_lstat.file? && !userfile.is_a?(SingleFile)
 
-    matched_class = SingleFile.descendants.unshift(SingleFile).find { |c| file_name =~ c.file_name_pattern }
+    matched_class = SingleFile.descendants.unshift(SingleFile).find { |c| file_name =~ c.file_name_pattern } ||
+                    userfile.class
     viewer        = matched_class.class_viewers.first.partial rescue nil
 
     if matched_class && viewer
@@ -100,10 +101,10 @@ module UserfilesHelper
                                              :viewer                => viewer,
                                              :viewer_userfile_class => matched_class
                                              ),
-            :replace => "sub_viewer_filecollection_cbrain",
+            :replace => replace_div_id,
           }
         ) do
-          ("<span class=\"sub_viewable_link\">"+display_name+"</span>").html_safe
+          ("<span class=\"sub_viewable_link "+replace_div_id+"\">"+display_name+"</span>").html_safe
       end
     elsif display_name =~ /\.html$/i # TODO: this will never happen if we ever create a HtmlFile model with at least one viewer
       link_to "#{display_name}",
--- a/BrainPortal/app/helpers/userfiles_helper.rb
+++ b/BrainPortal/app/helpers/userfiles_helper.rb
@@ -87,10 +87,12 @@ module UserfilesHelper
 
     file_lstat = full_path_name.lstat  # lstat doesn't follow symlinks, so we can tell if it is one
 
-    return h(display_name) if !file_lstat.file? && !userfile.is_a?(SingleFile)
+    # return if userfile class is a FileCollection and file is not a file (i.e. a directory)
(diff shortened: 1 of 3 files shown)


BrainPortal/app/helpers/userfiles_helper.rb
```diff
@@ -100,10 +103,10 @@ def data_link(file_name, userfile)
                                              :viewer                => viewer,
                                              :viewer_userfile_class => matched_class
                                              ),
-            :replace => "sub_viewer_filecollection_cbrain",
+            :replace => replace_div_id,
           }
         ) do
-          ("<span class=\"sub_viewable_link\">"+display_name+"</span>").html_safe
+          ("<span class=\"sub_viewable_link "+replace_div_id+"\">"+display_name+"</span>").html_safe
```
prioux, line 109: ID vs CLASS to resolve.

natacha-beck: Should be ok now.

BrainPortal/app/views/userfiles/_default_qc_panel.html.erb
```diff
@@ -22,15 +22,29 @@
...
+      </span>
+    <% end %>
   <% end %>
-<% end %>
+</div>
+
+<script type="application/javascript">
+  // Find the link inside the div and click it
+  var subViewerDiv  = $("#<%= div_id_to_replace %>");
+  var subViewerLink = subViewerDiv.find(".ajax_onclick_show_element");
```
prioux, line 46: What is this class? How is it added to the HTML elements? If it's done later after the page is loaded, could this script be executed too early? Maybe the script tag shoudl use the attribute `defer` ?

prioux: Try this in a bran new browser in incognito and no cached javascript

natacha-beck: Done, it works fine in an incognito browser.

BrainPortal/app/views/userfiles/_default_qc_panel.html.erb
```diff
@@ -22,15 +22,29 @@
...
+  <% if ! @userfile.is_locally_synced? %>
+    This file must be synced locally to view QC data.
   <% else %>
-    <span class="warning">
-    This file does not seem to contain any QC data, or no template is available to QC this file type.
-    </span>
+    <% link = data_link(@userfile.name, @userfile, div_id_to_replace) %>
+    <% if link =~ /data-url=/%>
+        <%= link %>
+      </div>
```
prioux, line 34: Your DIV structure is imbalanced!!!

BrainPortal/app/views/userfiles/_quality_control_panel.html.erb
```diff
@@ -36,10 +36,12 @@
     <legend>Navigation: file <%= @current_index + 1 %> out of <%= @filelist.size %></legend>
     <%= link_to_userfile_if_accessible(@userfile, current_user, :html_options => { :target => '_blank' }) %>
     <p>
-    <% if @current_index > 0 %>
+    <% first_file = @current_index > 0 %>
```
prioux, line 39: I don't get it. The variable 'first_file' is TRUE when this is NOT the first file ?!?!? What kind of naming is this?

prioux (changes requested):
Fix bugs please

prioux (changes requested):
See comments

BrainPortal/app/views/userfiles/_default_qc_panel.html.erb
```diff
@@ -23,14 +23,30 @@
...
     <span class="warning">
-    This file does not seem to contain any QC data, or no template is available to QC this file type.
+      This file does not seem to contain any QC data, or no template is available to QC this file type.
     </span>
   <% end %>
 <% end %>
 
+<script type="application/javascript" defer>
+  // Find the link inside the div and click it
+  <% if div_id_to_replace %>
```
prioux, line 45: I feel uncomfortable about the scope of the variable `id_to_replace` though. It feels it is a local variable inside a code block well above in the HTML page, so maybe move its initialization out to be at the level of the entire page, or (better) move the entire script tag inside the `if link =~ /data-url/`

BrainPortal/app/views/userfiles/_default_qc_panel.html.erb
```diff
@@ -23,14 +23,30 @@
...
-  This file must be synced locally to view QC data.
+  <span class="warning">
+    This file must be synced locally to view QC data.
+  </span>
 <% else %>
-  <% if @userfile.is_a?(ImageFile) %>
-    <%= image_tag url_for(:action  => :content, :id => @userfile.id) %><br>
+  <% div_id_to_replace = "sub_viewer_#{@userfile.id}" %>
+  <% link = data_link(@userfile.name, @userfile, div_id_to_replace) %>
+  <% if link =~ /data-url=/%>
```
prioux, line 32: Please add a space before the final `%>`

prioux: Also, here would go the `<script>` tag

BrainPortal/app/views/userfiles/_default_qc_panel.html.erb
prioux: My original comment before I suggested moving the script tag was that the condition `if div_id_to_replace` shoudl go AROUND the entire script tag, but that is moot if you move the script tag to the HTML above.

prioux (changes requested):
Re-entered some comments

PR #1316, opened 2023-03-28
[FIX] BTQ integrator outputfile type setter

### LINKED ISSUE #1315: BTQ integration: error save result with *.mat folder.
I have a tool ([mcflirt](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/MCFLIRT))  that produce a folder named *_mcf.mat. 
When running the tool I get: 

> CbrainError: Error: incompatible directory 'D99_template_pd_mcf.mat' given for a SingleFile.

My first try to fix the bug was to set `BoutiquesOutputFileTypeSetter` for the output to `FileCollection` but it still failed because it is ignore in `name_and_type_for_output_file` of `boutiques_cluster_task.rb`.

natacha-beck:
Fix #1315

### DIFF
--- a/BrainPortal/app/models/boutiques_cluster_task.rb
+++ b/BrainPortal/app/models/boutiques_cluster_task.rb
@@ -331,7 +331,8 @@ class BoutiquesClusterTask < ClusterTask
 
     # Get basename, use it to guess the class
     name = File.basename(pathname)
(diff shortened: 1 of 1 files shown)


prioux:
Rejected. You can't refer to custom external module in the base framework like that

PR #1326, opened 2023-04-26
Fixes race conditions when tracking file syncing

prioux:
Adds logic to try twice if a sync object gets newly created and picked up by multiple clients on the same server. Twice is enough because the second time the code will receive a completion status from whatever other process did permorm the operation.

This commit should get rid forever of all exceptions that appeared as:

CbrainTransitionException: Sync status was changed before lock was acquired.

### DIFF
--- a/BrainPortal/app/models/sync_status.rb
+++ b/BrainPortal/app/models/sync_status.rb
@@ -126,42 +126,51 @@ class SyncStatus < ApplicationRecord
     unless userfile_id
       return yield
     end
+    prettyfile = "'#{userfile.name}' (##{userfile_id})" # for messages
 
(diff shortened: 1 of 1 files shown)


prioux:
Note for reviewers: the main difference is that in 4 methods, I inserted a new block:

```ruby
2 times do
  (OLD indented by two more spaces)
end

if status check not expected
   raise exception
end
```

The code in the `2.times` block is pretty much unchanged except at the end, instead of a `status_transition!(blah)` we have a `break if status_transition(blah)` (note the absence of the `!` in the method name).

prioux:
Consider reviewing this with spacing not significant.

PR #1331, opened 2023-05-04
fix error message duplication for duplicate filename ( fixes #1309 )

### LINKED ISSUE #1309: Trying to upload a file that already exists shows bad error message
When trying to upload a file that already exists, users get to see the message "422 Unprocessable entity"; that show up until a new page is loaded and THEN the proper message "name has already been taken" is shown.

MontrealSergiy:
Fixes confusing error messages when trying to upload a file that already exists shows bad error message

I did not investigate messages for any other errors.

https://github.com/aces/cbrain/issues/1309

### DIFF
--- a/BrainPortal/app/controllers/userfiles_controller.rb
+++ b/BrainPortal/app/controllers/userfiles_controller.rb
@@ -549,8 +549,9 @@ class UserfilesController < ApplicationController
         flash[:error]  += "It might help to rename the file.\n" if flash[:error].include? "Name is already"
         respond_to do |format|
           format.html { redirect_to redirect_path }
-          format.json { render :json => { :notice => flash[:error] }, :status => :unprocessable_entity
-                        flash.discard
+          format.json {
+                          render :json => { :notice => flash[:error] }, :status => :unprocessable_entity
+                          flash.discard # no need to repeat error message
                       }
         end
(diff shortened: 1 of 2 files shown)


prioux:
I doesn't seem to work. I tried uploading twice the same file and the second time I still get the ugly 422 Unprocesseable Entity message.

prioux:
Never mind, it works, my Firefox was still caching the old javascript.

BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -546,9 +546,13 @@ def create #:nodoc:
         userfile.errors.each do |field, error|
           flash[:error] += "#{field.to_s.capitalize} #{error}.\n"
         end
+        flash[:error]  += "It might help to rename the file.\n" if flash[:error].include? "Name is already"
```
prioux, line 549: The error message is normally `Name has already been taken.` which doesn't match `Name is already`

MontrealSergiy: corrected as requested, but feel free to drop that line if still in doubts

prioux:
I verified that it works and I will merge it, but I will adjust the message a little afterwards.

PR #1333, opened 2023-05-15
[ADD] ability to add note at the top of form when creating a task

### LINKED ISSUE #1025: Add support for notes-for-users in the custom section of Boutiques descriptor
I'd like the Boutiques integrator to show some notes in bullet-point format at the top of a form when launching a task.

The notes would come from the `custom` section of the descriptor:

```json
"custom": {
  "cbrain:input_notes": [
    "Please leave the field number of threads (--threads) blank, it is not used in CBRAIN",
    "Do not enter a maximum memory (-MAX_MEN)"
  ]
}
```

which would simply render as these notes at the top of the form, something like this:

-----------------------------------

**CBRAIN integration notices:**

* Please leave the field number of threads (--threads) blank, it is not used in CBRAIN
* Do not enter a maximum memory (-MAX_MEN)

-----------------------------------

natacha-beck:
Close #1025. 

To test if please add (or not) the following on a Boutiques descriptor (in custom section): 

```
        "cbrain:input_notes": [
            "Note 1: this is the first note",
            "Note 2: this is the  note"
        ],
```

It works for bot Boutiques integrato old and new.

### DIFF
--- a/BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_task_params.html.erb
+++ b/BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_task_params.html.erb
@@ -27,52 +27,78 @@
 <% end %>
 
 <%
-  @descriptor = @task.descriptor_for_form
-  inputs      = @descriptor.inputs          || []
-  required    = @descriptor.required_inputs || []
-  optional    = @descriptor.optional_inputs || []
-  groups      = @descriptor.groups          || []
+  @descriptor      = @task.descriptor_for_form
+  inputs           = @descriptor.inputs          || []
+  required         = @descriptor.required_inputs || []
+  optional         = @descriptor.optional_inputs || []
+  groups           = @descriptor.groups          || []
   params_in_groups = groups.map(&:members).flatten.uniq # list of param IDs
   interface_files  = Userfile.where(:id => @task.params[:interface_userfile_ids])
+  input_notes      = @descriptor.custom['cbrain:input_notes'] || []
 %>
 
-<div class="task-params">
-  <%- if inputs.empty? -%>
-    <%= @task.name %> has no parameters.
-  <%- else -%>
-  <ul>
-    <%# Write info for ungrouped parameters %>
-    <% (required + optional).each do |input| %>
-       <% next if params_in_groups.include? input.id %>
-       <%= render :partial => task_partial(:boutiques_input), :locals =>
-           { :input           => input,
-             :isGroupMember   => false,
-             :interface_files => interface_files,
-           }
-       %>
-    <% end %>
-    <%# Write info for grouped parameters %>
-    <%- groups.each do |group| %>
-       <%= render :partial => task_partial(:boutiques_group), :locals =>
-           { :group           => group,
-             :interface_files => interface_files,
-           }
-       %>
-    <% end %>
-  </ul>
+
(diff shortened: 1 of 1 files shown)


MontrealSergiy:
For some on few occasions Pierre insisted on not changing other places of CBRAIN when working with modules, and doing everything from the module in the library.

Workaround to avoid changing template could be to use flash notices, or fake inputs (called Note)

Maybe check with him

MontrealSergiy:
works on my local install (new boutiques)

Screenshot from 2023-05-31 14-27-46

BrainPortal/app/views/tasks/_params.html.erb
```diff
@@ -44,11 +44,36 @@
...
+  <% if raw_partial.(:task_params) %>
+    <%= render :inline => raw_partial.(:task_params), :locals => locals %>
+  <% else %>
+    <%= render :partial => task_partial('task_params'), :locals => locals %>
+  <% end %>
 </div>
+
+<% if input_notes.any? %>
+  </div>
+<% end %>
```
MontrealSergiy, line 79: not sure why errors is subdiv of notes div

MontrealSergiy (approved):
I am not totally sure of code but seems works as expected on my local install

BrainPortal/app/views/tasks/_params.html.erb
prioux: This is wrong. Sergiy is right to be suspicious. If you have to put a `</div>` inside an `if` it certainly means the layout structure is badly done.

Also, this file here is a template for the general params page of all tasks, it cannot possibly have any sort of knownlede or dependencies on things like `@task.descriptor_for_before_form` which are compltely internal to the new BTQ integrator.

prioux (changes requested):
Re-think the whole thing, see my comments

prioux (changes requested):
Please adjust the layout of the HTML and make sure strings are properly escaped. There shouldn't be if/else statements are opening and closing `div` elements, it's a sign that the structure of the page is not sound.

Also, revert the changes to the old integrator; we won't use it anymore and testing/debugging that old code is wasted time.

BrainPortal/app/views/tasks/_params.html.erb
```diff
@@ -44,11 +44,36 @@
...
 
-<div class="generalbox">
-<%= error_messages_for(@task) %>
-<% if raw_partial.(:task_params) %>
-  <%= render :inline => raw_partial.(:task_params), :locals => locals %>
-<% else %>
-  <%= render :partial => task_partial('task_params'), :locals => locals %>
+<%
+input_notes =
+  @task.class.superclass == PortalTask ?
```
prioux, line 49: This is really not right. The current view file, `_params.html.erb`, does not and should not know anything about the internals of the new Boutiques integrator. It is not correct for it to invoke the method `descriptor_for_before_form` because that is not a method of the core CbrainTask class.

What you have to do is add a PortalTask method called, maybe "input_notes" that by default returns `[]` with a comment that says subclasses can provide an array of short helpful comments. And in the BoutiquesPortalTask class, you override it.

prioux:
Instead of adding the notes in the file

```
BrainPortal/app/views/tasks/_params.html.erb
```

which is the main wrapper, consider adding the notes more specifically in the wrapper for boutiques tasks, in

```
BrainPortal/cbrain_plugins/cbrain-plugins-base/cbrain_task/boutiques_task/views/_task_params.html.erb
```

prioux:
Why does this PR suddenly contain the commits for the overlay PR ?

Issue #1292, opened 2023-01-13
The 'My account' page should have a section with disk quotas


prioux:
The "my Account" page should have a small report about all the quotas that apply to the user.

Also, as an addition, we have a general disk quotas index table that show (to the admin) all the quota records, and (to a user) all the quota records that apply to them. That table doesn't help an admin figure out which user is near their quotas. Maybe we need a new table for that? We could also have an option to show the table 'as user XYZ' ?

MontrealSergiy:
What is near quota? Say 95 % ?

prioux:
95% seems like a fine idea.

Note that the DiskQuota models don't have any methods to do things like that. We'd have to add such methods first.

PR #1344, opened 2023-08-22
user quota section, near quote report, view quotes as user dropdown #1292

MontrealSergiy:
see preliminary discussion and motivation at #1292


MontrealSergiy:
CI db failure, mariadb version is missing 

```manifest for mariadb/server:latest not found: manifest unknown: manifest unknown```

Issue #1345, opened 2023-08-25
Add a 'copy to clipboard' button in the API token page


prioux:
CBRAIN has a page that allows users to generate an API token.

When the token is generated, it is shown int he page and users are expected to write it down/copy it somewhere.

I suggest that just besides the token we had a nice 'copy to clipboard' button.

MontrealSergiy:
a snippet for that https://www.w3schools.com/howto/howto_js_copy_clipboard.asp

MontrealSergiy:
a dedicated library, with an old backend https://clipboardjs.com/ with cool button and 'your text is copied' message

https://www.golinuxcloud.com/javascript-copy-to-clipboard/ recommends it as a polyfill/fall-back solution for old browsers.

PR #1346, opened 2023-08-29
a button to copy api token to clipboard #1345

MontrealSergiy:
see #1345


natacha-beck:
I have multiple comment: 

- The code did not work as you expected, when you copy the API key you have space (before and after the API key you even have a `\n`. You should trim it. 
- I do not see the necessity of having the class `copiable` instead you can use an ID. 

I think this can do the job: 

```html
We have just generated a new API token for you (click to copy).
<p>
<div>
    <button onclick="copy_text_to_clipboard()" id="api_token"><%= @new_token %></button>
</div>
<p>
```

and: 

```javascript
function copy_text_to_clipboard() {
    navigator.clipboard.writeText($("#api_token").text().trim());
}
```

It gave: 

image

Also you need to correct the date in the Copyright section.

MontrealSergiy:
@natacha-beck do you actually observed the whitespace in copied string with your browser (safari I guess?), or trimming is 'just in case'?)

MontrealSergiy:
I can adjust Copyrights, but fyi I do not think it is no longer part of cbrain development policies. Darcy had a  discussion with Pierre regarding relevance of Copyright date updating, and I think he convinced everybody that copyright update is not necessary.

natacha-beck:
> I can adjust Copyrights, but fyi I think it is no longer part of cbrain development policies. Darcy had a discussion with Pierre regarding relevance of Copyright date updating, and I think he convinced everybody that copyright update is not necessary.

So do not update it, it's fine.

natacha-beck:
> @natacha-beck do you actually observed the whitespace in copied string with your browser (safari I guess?), or trimming is 'just in case'?)

I use chrome and yes I observe it by copying it, then when debugging and print in the console I was able to confirm that it copy some beginning white space and the a `\n`

MontrealSergiy:
Ok then. Trimming never hurts, I do not mind to delete all the styling.

Your solution is elegant, yet ticket  requested to add a new button besides token, rather than placing token into button. I do mind strongly but if you insist I have to consult with team or at the least the requester of the feature. And even in this case I suggest to add small copying icon aside in svg, and, perhaps a tooltip.

MontrealSergiy:
While your token inside button suggestion has elegance, it might go again what many developers do in unix-like system - selecting word with a double click and then copying it with wheel click, without need for any clipboard

natacha-beck:
You can discard my comment about the API button if you want, result is the same. 

Your 1st solution with the trimming is good, but you can consider to remove the `copiable` class that is not needed.

MontrealSergiy:
Ok, please, check the changes. I can rebase if you like

natacha-beck (approved):

BrainPortal/public/javascripts/cbrain/users/clipboard.js
```diff
@@ -0,0 +1,27 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+*/
+
+function copy_text_to_clipboard() {
```
prioux, line 25: This file is adding a function in the top-level javascript namespace?!? This is wrong, it becomes a global identifier.

Please add the function to cbrain.js were we isolate all the functions properly. Use a class to bind it to your element.

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -790,6 +790,56 @@
...
+        // Keep line breaks
+        let text = $this.html()
+            .replace(/<br\s*\/?>/g, "\n")
+            .replace(/<\/p>\s*<p>/g, "\n")
+            .replace(/<\/div>\s*<div>/g, "\n")
+            .replace(/<\/?[^>]+(>|$)/g, "")
+            .trim();
+
+        if ($this.find(".copy-button").length === 0) {
+          let copyBtn = $(`
```
prioux, line 809: I don't understand why you add the button and SVG image in the javasript code. Shouldn't this be just in the HTML template already? The JS code should only take care of the event triggered by a click, and make the copy to the clipboard. Everything visible in the page should already be in the view template, one way or another.

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -790,6 +790,56 @@
...
+          $this.css("position", "relative");
+          $this.append(tooltip);
+
+          // Attach the button
+          $this.append(copyBtn);
+
+          copyBtn.on("click", function (event) {
+            event.stopPropagation(); // Allow normal text selection
+            navigator.clipboard.writeText(text).then(() => {
+              console.log("Copied");
```
prioux, line 833: Don't console.log() in a PR. Use it if you want to debug, but make you not to leave these around when you submit a PR

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -790,6 +790,56 @@
...
 
+
+    // Adds copy to buffer button to all the element of class 'copy-text'.
+
+    if (navigator.clipboard) {
+      $(".copy-text").each(function () {
+        let $this = $(this);
+
+        // Keep line breaks
+        let text = $this.html()
```
prioux, line 801: Why is this even necessary?

I think you're trying to extract the token? Why not extract it from a data-token attribute in HTML instead of the user facing HTML ? e.g.

```HTML
<div data-token="1836482537455723637374">
  <pretty>
    <something>
       1836482537455723637374
    </something>
   </pretty>
</div>
```

Basically, you fetch the token value from the attribute, not from the pretty HTML. That way you don't have to do any parsing, escaping etc.

prioux (changes requested):
Restructure code

BrainPortal/public/javascripts/cbrain.js
MontrealSergiy: The goal of replacing `<br>` and `<p>` was to address multiline snippets, but at the moment it is not particularly needed and simple `.test()` worked fine in the previous version. The reliance on text rather than `data-` was to make the copy feature to easy to replicate without unnecessary duplication of code. Believe or not for a while I thought to add `data-` support as well, as an alternative option.

BrainPortal/public/javascripts/cbrain.js
prioux: But if you want to make it reusable, create a helper method! That's what helpers are for. The folder BrainPortal/app/helpers is full of them. Right now if someone wants to change the layout of the copy-to-clipboard button, we have to go into the javascript, that doesn't make any sense. The javascript code should do only the binding and event processing for copying to the clipboard, all the button and layouts should be performed by a helper. Something like:

```erb
Your token is <%= @token %> <%= copy_to_clipboard_button @token %>
```

which would generate the HTML, and put the token inside an attribute "data-clipboard" within the button. No need to fetch the text anywhere else, or parse it. Just a button.

MontrealSergiy: ok

MontrealSergiy:
Ok, a helper added

BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -336,6 +336,40 @@ def button_with_dropdown_menu(title, options={}, &block)
...
+      '<svg
+        width="12" height="12" viewBox="0 0 24 24" fill="none"
+        xmlns="http://www.w3.org/2000/svg" style="vertical-align: middle;">
+        <rect x="4" y="4" width="14" height="14" rx="2" ry="2"
+        fill="#d1d5da" stroke="#6a737d" stroke-width="1.25"/>
+        <rect x="7" y="7" width="14" height="14" rx="2" ry="2"
+        fill="#d1d5da" stroke="#6a737d" stroke-width="1.25"/>
+      </svg>'
+
+    options['class']         = options['class'].to_s + " copy_button"
```
prioux, line 362: You're mutating the content of options['class'] here, so that the caller has now a different options hash after calling your helper. Use an intermediate variable or dup options

BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -336,6 +336,40 @@ def button_with_dropdown_menu(title, options={}, &block)
...
+    options['data-clipboard'] = txt
+
+    tooltip = '<span class="copy-tooltip">' + ok_msg + '</span>'
+
+    atts = options.to_html_attributes
```
prioux, line 367: Since you already put the 'class' attribute in options it will generate it in to_html_attribute, so why do you add the class again explicitely at line 368? You're going to end up with 

```html
<button class="copy-button" class=" copy_button" ...>
```

Did you run this helper in the console to test what it creates?

prioux (changes requested):
Aside from my two comments about the helper, the overall structure of this pull request is good. Isn't it much cleaner than the original design?

BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -336,6 +336,40 @@ def button_with_dropdown_menu(title, options={}, &block)
...
+  #
+  # Options:
+  # [:label] a button label. Defaults to 'copy' followed by a copy icon
+  # [:message] feedback message. Default to 'Copied!'.  #
+  # All other options are treated as HTML attributes of the button
+  def copy_to_clipboard_button(txt, options={})
+    ok_msg   = options.delete(:message)
+    ok_msg ||= 'Copied!'
+    label    = options.delete(:label)
+    label  ||= 'copy' +
```
prioux, line 352: I'd like the caller to be able to change the text label and still get the icon. Right now if I change the label, the icon disappear.

Also the default label should be the full "Copy to clipboard", because just the word "copy" is not clear enough.

prioux:
I'm not sure why you 'asked for my review' given nothing has changed since I last reviewed this PR and you have not addressed my comments about your helper.

MontrealSergiy:
I might jumped the gun here. I was mainly trying to draw attention to my comment, which is mostly about label wording but could also impact other aspects. Reading it now, I realize it's a bit long and might be better discussed in person. In the meantime, I'll try to improve the helper. I'm still uncertain: (1) whether the icon or text should come first, and (2) whether a clipboard icon (or a Unicode symbol) would work better. But I'll do my best.

MontrealSergiy:
1. I got feeling you originally leaned to have separate helper params for the icon, so I added one. 

If I misunderstand you it is should be pretty easy to delete.

2. Also I am still on fence  is clipboard icon better than copy icon, added both icon helpers just in the case.

axif0:
Just curious, Don't we also need to add clipboard in [nh_users/new_token.html.erb](https://github.com/aces/cbrain/blob/69e05f6b35677788177aa0bb4d04e3d221c1c7d0/BrainPortal/app/views/nh_users/new_token.html.erb#L40)?

prioux:
Please merge this patch to your PR:

```diff
diff --git a/BrainPortal/app/assets/stylesheets/cbrain.css.erb b/BrainPortal/app/assets/stylesheets/cbrain.css.erb
index 1b6a94954..24357e95b 100644
--- a/BrainPortal/app/assets/stylesheets/cbrain.css.erb
+++ b/BrainPortal/app/assets/stylesheets/cbrain.css.erb
@@ -2800,17 +2800,12 @@ img {
 /* % ######################################################### */
 
 .copy-button {
-    margin: 1px;
-    margin-bottom: 15px;
-    margin-left: 8px;
-    vertical-align: middle;
     color: #484848;
-    font-size: 0.8em;
     border-radius: 4px;
     border: 2px solid #6a737d;
     background-color: #e0e0e0;
+    padding: 0.6em;
     position: relative;
-    user-select: none;
 }
 
 .copy-text {
@@ -2819,9 +2814,7 @@ img {
     border-style: solid;
     word-break: break-all;
     overflow-wrap: break-word;
-    text-align: left;
-    padding: 6px;
-    padding-top: 12px;
+    padding: 1em;
     border-radius: 5px;
     background-color: #f5f5f5;
 }
@@ -2832,10 +2825,10 @@ img {
     right: 1px;
     background: #333;
     color: white;
-    padding: 3px;
+    padding: 0.5em;
     border-radius: 4px;
     opacity: 0;
-    transition: opacity 0.3s;
+    transition: opacity 1.0s;
     pointer-events: none;
 }
 
diff --git a/BrainPortal/app/views/users/new_token.html.erb b/BrainPortal/app/views/users/new_token.html.erb
index 76cb2ff69..bb8ca6ace 100644
--- a/BrainPortal/app/views/users/new_token.html.erb
+++ b/BrainPortal/app/views/users/new_token.html.erb
@@ -28,11 +28,12 @@
 
 We have just generated a new API token for you.
 <p>
-  <code class="copy-text"><%= @new_token %><%= copy_to_clipboard_button(
-                                                 @new_token,
-                                                 label: "copy",
-                                                 title: "Click this button to copy the API token to the clipboard")
-  %></code>
+  <code class="copy-text"><%= @new_token %></code>
+  <%= copy_to_clipboard_button(
+     @new_token,
+     label: "copy",
+     title: "Click this button to copy the API token to the clipboard")
+  %>
 <p>
 
 If you are a developer and want to automate working
```

Issue #1343, opened 2023-08-10
Create a "mkdir" utility module for Boutiques integrator


prioux:
Some tools integrated with boutiques receive as an input a string which will be provided on the command line, and that string is supposed to point to an existing subdirectory. E.g.

```json
"command-line":  "supertool -out [OUTDIR] [OPT] [OPT]..."
```

 When integrating such tools in CBRAIN, they often expect the OUTDIR to be pre-existing, and the tool crash at run time.

Historically, we modified some of these descriptor within CBRAIN so the the command was changed to

```json
"command-line": "mkdir -p [OUTDIR] ; supertool -out [OUTDIR] [OPT] [OPT]..."
```

But it's kind of ugly and it means the descriptor is no longer really as generic (usable outside CBRAIN as well as within).

I suggest we implement a new utility module where a CBRAIN dev can just leave the command-line intact, and specify which input argument to run mkdir on when setting up the task.

Something like:

```json
"BoutiquesSubdirMaker": [ "output_dir", "someinputid2" ]
```

prioux:
Maybe the values for the configuration could instead be templates, with values subsituted from any inputs. e.g.

```json
"BoutiquesSubdirMaker": [ "[OUTDIR]", "[OUTDIR]/[THRESHOLD]_res" ]
```

That woudl make the module much more flexible.

PR #1348, opened 2023-09-08
utility module make folders #1343

MontrealSergiy:
see #1343 for details

### DIFF
--- a/BrainPortal/lib/boutiques_dir_maker.rb
+++ b/BrainPortal/lib/boutiques_dir_maker.rb
@@ -73,9 +73,10 @@ module BoutiquesDirMaker
         self.addlog("BoutiquesDirMaker skips '#{pattern}', dir '#{path}', absolute paths are not supported")
         next
       end
-      path.gsub!(/[^0-9A-Za-z.\/\-_]+|(\.\.+)/, '_')
+      path.gsub!(/[^0-9A-Za-z.\/\-_]+|(\.\.+)/, '_')  #
     end
-    FileUtils.mkdir_p paths
+    self.addlog("The templates are evaluated as #{paths}")
+    FileUtils.mkdir_p paths.compact
     true
   end
 end
--- a/BrainPortal/lib/boutiques_dir_maker.rb
+++ b/BrainPortal/lib/boutiques_dir_maker.rb
@@ -33,7 +33,7 @@ require 'fileutils'
 #          [ "[OUTDIR]", "[OUTDIR]/[THRESHOLD]_res", "tmp" ]
 #
 # Please avoid special characters save underscore and hyphen, and
-# use relative paths. Boutiques templates (hereafter called patterns) are supported.
+# use relative paths. Boutiques templates (hereafter called patterns) are supported
 #
 module BoutiquesDirMaker
 
@@ -58,8 +58,9 @@ module BoutiquesDirMaker
     descriptor = self.descriptor_for_setup
     patterns   = descriptor.custom_module_info('BoutiquesDirMaker')
 
-    # invoke main setup
-    return false if ! super
+    # invoke overridden method
+    commands = super
+    return false if ! commands # early return
 
     substitutions_by_token  = descriptor.build_substitutions_by_tokens_hash(
       JSON.parse(File.read(self.invoke_json_basename))
@@ -69,14 +70,19 @@ module BoutiquesDirMaker
     paths = patterns.map do |pattern|
       # Replace tokens
       path = descriptor.apply_substitutions(pattern, substitutions_by_token)
+
       if Pathname(path).absolute?
         self.addlog("BoutiquesDirMaker skips '#{pattern}', dir '#{path}', absolute paths are not supported")
         next
       end
-      path.gsub!(/[^0-9A-Za-z.\/\-_]+|(\.\.+)/, '_')  #
+
+      # replacing weird and special characters
+      if path.gsub!(/[^0-9A-Za-z.\/\-_ ]+|(\.\.+)/, '_')
+        self.addlog("Note, special symbols are encountered in directory name which is replace with '#{path}'")
+      end
+      path
     end
-    self.addlog("The templates are evaluated as #{paths}")
     FileUtils.mkdir_p paths.compact
-    true
+    commands
   end
 end
--- a/BrainPortal/lib/boutiques_dir_maker.rb
+++ b/BrainPortal/lib/boutiques_dir_maker.rb
@@ -20,21 +20,20 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+require 'fileutils'
+
 # Some tools expect that a directory e.g. results, etc exists in the working directory
 #
 # While traditionally we add to the boutiques command line prefix akin to 'mkdir -p [OUTPUT];'
 # external collaborators might dislike polluting the command line with technical staff
 # or need a clean boutiques descriptor which does not create any new folders
-# For example :
-#
-#
-# and in the `cbrain:integrator_modules` section look like:
+# The the `cbrain:integrator_modules` section look like:
 #
 #     "BoutiquesDirMaker":
 #          [ "[OUTDIR]", "[OUTDIR]/[THRESHOLD]_res", "tmp" ]
 #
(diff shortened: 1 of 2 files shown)


prioux:
The code to make the directories should not be performed in a method that is meant to return a descriptor to the framework. It should be done in "setup", that's the proper place where everything related to preparing a work directory is supposed to happen.

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,73 @@
...
+    basename = Revision_info.basename
+    commit   = Revision_info.short_commit
+    self.addlog("Creating directories with BoutiquesDirMaker.")
+    self.addlog("#{basename} rev. #{commit}")
+
+    dir_names.each do |x|
+      if Pathname(x).absolute?
+         raise CbrainError("#{x} is absolute paths, which are presently not supported")
+      end
+      x.gsub!(/[^0-9A-Za-z.\/\-_\[\]]|(\.\.+)/, '_')  # silent sanitizing and quotes
```
prioux, line 65: The regex is not bad but

1. don't allow square brackets  (remove `\[\]`)
2. match multiple times the bad characters (add `+` after the character range)

so

```ruby
   x.gsub!(/[^0-9A-Za-z.\/\-_]+|(\.\.+)/, '_')
```

Also don't call a variable x when a better name like `relpath` is clearer

prioux: Also see my separate comment on the PR's discussion: the code to mkdir shoudl be done in setup, as performed using the pattern substitution methods we already have within CBRAIN. See the BoutiquesPostProcessingCleaner which has the ability to substitute `[TOKENS]` within strings.

prioux (changes requested):
Two comments in code and one comment in main discussion

BrainPortal/lib/boutiques_dir_maker.rb
MontrealSergiy: Yes, thanks, BoutiquesPostProcessingCleaner, yet not in early stages, only when invocation file is completely ready. 

So to apply that procedure at setup, as you suggest, I would need copy or factor-out the invocation massaging (file ids => filenames, bools etc). Any preference?

prioux: Invoke super() at the beginning of your overridden setup(), then read back the invocation struct.

MontrealSergiy: it is the invocation struct need massaging

prioux: You don't need to filter out anything. The struct is just exactly what you need and nothing more or less. What do you think is not right about it?

prioux:
Actually I double checked, the on-disk invoke struct is created during 'cluster_commands()' so that's the method you need to override instead of setup().

MontrealSergiy:
So do I keep old version (with minor improvements in  variable names and plus of course), or should I avoid altering the command line

prioux:
Don't modify the command line. Just writ the ruby code that run the mkdir. That's it.

MontrealSergiy:
ok. Another possibility is to factor out invocation struct massaging into separate method, so setup can be used

MontrealSergiy:
Done. 

(adjusted to create subfolders asap)

MontrealSergiy:
As agreed in person with Pierre,  the method cluster_commands() is modified (but in a more complex way than at the first attempt)

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,88 @@
...
+    basename = Revision_info.basename
+    commit   = Revision_info.short_commit
+    self.addlog("#{basename} rev. #{commit}")
+
+    descriptor = self.descriptor_for_setup
+    patterns   = descriptor.custom_module_info('BoutiquesDirMaker')
+
+    # invoke overridden method
+    commands = super
+    return false if ! commands # early return
```
prioux, line 63: remove line 63: we will return commands at the end no matter what its value is, and a value of nil is allowed in the framework anyway.

prioux: Also, `command=super` should be the first thing in the method; that way if it crashes the logs of the task won't be misleading because you printed "Created directories" just before the crash (which would confuse someone unto thinking that the module is responsible for the crash)

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,88 @@
...
+    substitutions_by_token  = descriptor.build_substitutions_by_tokens_hash(
+      JSON.parse(File.read(self.invoke_json_basename))
+    )
+
+    # Process each directory pattern
+    paths = patterns.map do |pattern|
+      # Replace tokens
+      path = descriptor.apply_substitutions(pattern, substitutions_by_token)
+
+      if Pathname(path).absolute?
```
prioux, line 74: Make another check in case path starts with "..". That would pass the test "is not absolute" but one could then enter "../../../../../../../../usr/local/lib/etc/etc" to point to any path on the filesystem

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,88 @@
...
+        next
+      end
+
+      # replacing weird and special characters
+      if path.gsub!(/[^0-9A-Za-z.\/\-_ ]+|(\.\.+)/, '_')
+        self.addlog("Note that in '#{path}', underscore is used instead of special symbol(s)")
+      end
+      path
+    end
+    FileUtils.mkdir_p paths.compact
```
prioux, line 85: Consider calling the ClusterTask method 'safe_mkdir' which I added to the framework explicitly for creating subdirectories in the work directory of a task, with some checks.

prioux (changes requested):
Minor changes, one of them a security check

BrainPortal/lib/boutiques_dir_maker.rb
MontrealSergiy: ok

BrainPortal/lib/boutiques_dir_maker.rb
MontrealSergiy: that was checked below

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,82 @@
...
+    commands = super
+
+    # Log revision information
+    self.addlog("Creating additional directories with BoutiquesDirMaker.")
+    basename = Revision_info.basename
+    commit   = Revision_info.short_commit
+    self.addlog("#{basename} rev. #{commit}")
+
+    descriptor = self.descriptor_for_setup
+    patterns   = descriptor.custom_module_info('BoutiquesDirMaker')
```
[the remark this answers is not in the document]
prioux: I would also move the lines 65-66 ABOVE the lines that log the version, so that you can immediately do a `return commands if patterns.blank?` and not get a log message for a module that has done nothing at all.

BrainPortal/lib/boutiques_dir_maker.rb
```diff
@@ -0,0 +1,82 @@
...
+      JSON.parse(File.read(self.invoke_json_basename))
+    )
+
+    # Process each directory pattern
+    patterns.each do |pattern|
+      path = descriptor.apply_substitutions(pattern, substitutions_by_token)
+      path = path.strip  # whitespaces in dir names are not supported
+      path = Pathname.new(path.strip).cleanpath  # normalizing: removing unnecessary dots ...
+      cb_error "BoutiquesDirMaker cannot create path '#{path}' (pattern '#{pattern}')." if path.to_s.start_with?('.')
+      safe_mkdir(path)
```
prioux, line 78: Maybe add a addlog() line somewhere in the loop to log each folder created?

BrainPortal/lib/boutiques_dir_maker.rb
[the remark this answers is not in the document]
MontrealSergiy: so with blankness check, maybe then no need for `|| []`  ?

prioux:
@MontrealSergiy This looks all good to me now. I will make a final review tomorrow morning but I think it's fine.

PR #1350, opened 2023-09-15
Changed the main SSH key from RSA to ed25519

prioux:
Backwards compatibility is kept by adding the old RSA key to the agent, if it's found. Eventually, operators will want to only use the ed25519 key, they can do so by removing the files.


MontrealSergiy:
seem ok. Note that agent might be configured to expire id soon, e.g. with a variable for Ubuntu

```SSHAGENTARGS="-t 1h"```

MontrealSergiy (approved):
seems to work with either key.

PR #1354, opened 2023-10-11
HTML model and viewer plugin

MontrealSergiy:
an html file viewer with a iframe sandbox wrap.

Resubmitted PR after debasing to master as old was failing ci


BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb
```diff
@@ -0,0 +1,41 @@
...
+<%= link_to h("Open in a separate tab"), url_for(:controller  => :userfiles, :id  => @userfile.id, :action  => :display, :viewer => :html, :content_viewer => "on", :file_name => @userfile.name), :target => "_blank" if params[:action] == "show" %>
+
+<iframe sandbox
+    src="<%= stream_userfile_path(@userfile, :disposition => 'inline')%>"
+    class="bigiframe"
+    style="width:100%; height: 100%;"
+ >
+    HTML Viewer
+</iframe>
+HTML was sandboxed and filtered for greter safety
```
prioux, line 36: "greter" ?

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb
```diff
@@ -0,0 +1,41 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+%>
+
+
+
+<%= link_to h("Open in a separate tab"), url_for(:controller  => :userfiles, :id  => @userfile.id, :action  => :display, :viewer => :html, :content_viewer => "on", :file_name => @userfile.name), :target => "_blank" if params[:action] == "show" %>
```
prioux, line 27: Is this link just opening the same current page in a new tab, or something different from the current page?

In which situations would "params[:action]" not be "show" ?

prioux (changes requested):

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb
MontrealSergiy: That link opens other userfile in iframe, without the meta info. 

The condition lets me hide this link, assuming there is no need to show.

The action in that case is `display`

That said can delete this condition or any line if you are not comfortable.

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb
prioux: Why did you resolve this comment? I was pointing out a typo. "greter" is misspelled.

prioux: The sentence could be made better:

```
This HTML document was sandboxed and filtered for greater safety. If you trust the content of the file, you can <%= link_to "Show it in a separate tab, without sandboxing" ... %>.
```
```

File: BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb (hunk 2 above)
Comment (prioux) [reply BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb:27]: OK

**Review by prioux — CHANGES_REQUESTED** [review]
Small changes in sentences

File: BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb (hunk 1 above)
Comment (MontrealSergiy) [reply BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb:36]: I likely meant to revisit the wording day later but forgot to push the update, sorry. 

The suggested phrasing is great, though I'd put 'you can open' instead of 'we can display'—it feels more direct and empowering.

**Comment by MontrealSergiy** [comment]
I skipped the comma in the link label, seems redundant

**Comment by prioux** [comment]
Remember to test your helper in the console, to make sure it does what you expect.

File: BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+%>
+
+
+<%= link_to(
```
prioux, line 26: Can you please restructure this the 'standard' way? e.g.

```erb
  <% if page_is?('#show') %>
    <%= link_to ... %>
  <% end %>
```

The reason is, your big `<%=` block is reduced to nil of the condition is false, but we have to read all the way until the end to realize that it can print nothing.

MontrealSergiy:
Maybe drop "open without sandboxing", seems open for varios fishing and cross-site forgery? Or limit to file author

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb
```diff
@@ -0,0 +1,52 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+%>
+
+<% if page_is?('#show') %>
+  <%= link_to(
+      h("Open in a separate tab"),
```
prioux, line 27: You know you don't have to h() the link name in link_to(), it will do it automatically? And in any case, your constant string doesn't need any HTML escaping, it's just "Open in a separate tab" which stays exactly the same...

prioux (commented):
I made a comment but no need to change things. I will run the viewer on my dev env now.

prioux:
The 'separate tab with sandboxing' mechanism doesn't work right. I get a tiny region with scrollbars and underneat and to the right, your message's layout is not properly on a separate line. See screenshot. Can you please make sure you add `<p>` and `<br>`  between your messages?

Screenshot 2025-03-24 at 6 14 03 PM

For this test I save a CBRAIN page as HTML and re-uploaded it.

In the new tab, an inspect reveal the iframe has these attributes:

```
<iframe class="full-frame" sandbox="" src=...
```

I'm not sure of sandbox is supposed to have an empty string?

prioux:
I'm sending you two HTML files on Slack, please make sure your viewer works with both, even when changing back between TEXT mode and HTML mode. I can't get your PR to work right, myself. Also, adjust the layouts!

MontrealSergiy:
> I'm not sure of sandbox is supposed to have an empty string?
as far I got it empty sandbox means stricter sandbox, but if you would like to allow

- same origin
- scripts
- forms
- top navigation

we add those as attribute values. Please indicate if you like to enable any of above

MontrealSergiy:
I rolled back size of tab iframe by cost of using hardcoded html style. Tell me if It worth to put proper css.

Text file behavior was not changed. Files ended with html are not shown directly (only via link)
So would you prefer

- change the default behavior
- remove text viewer
- add 'html source' viewer
...

BrainPortal/cbrain_plugins/cbrain-plugins-base/userfiles/html_file/views/_html.html.erb
```diff
@@ -0,0 +1,52 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+%>
+
+<% if page_is?('#show') %>
+  <%= link_to(
+      "Open in a separate tab",
+            url_for(
```
prioux, line 28: I'm not sure why you are using such a low-level URL builder when we have higher level route helpers. Your URL should be built with:

```ruby
display_userfile_url(@userfile,
  :viewer         => :html,
  :content_viewer => "on",
  :file_name      => @userfile.name
)
```

This helper knows already about the action, the controller, and where to put the ID of the userfile inside it.

MontrealSergiy:
1. I am not yet sure if it is a good idea to grant any permissions to an HTML file opened without warning the user. However, based on our recent discussions, I have slightly relaxed sand-boxing and removed the second "open in the new tab as is, without sand-boxing link.

I hope it is closer to what you wanted, though it might be not.

2. Not sure maybe If would prefer to drop sandboxing completely, we put some confirmation button before rendering HTML below the metadata. 

3. The bug where links disappeared after switching to text mode and back has been fixed using an auxiliary query parameter, page helper does not work here for me.

prioux:
I think this code is now fine. The decisions you made regarding sandboxing are also fine. We can revisit the viewer later if necessary, but this works correctly.

PR #1355, opened 2023-10-11
quota improvements (rebased)

MontrealSergiy:
see preliminary discussion and motivation at https://github.com/aces/cbrain/issues/1292


natacha-beck:
@MontrealSergiy Please review it I was trying to test it but it cannot be merge in the current version of CBRAIN.

PR #1356, opened 2023-10-18
logging walltime and other qsub details

MontrealSergiy:


BrainPortal/app/models/cluster_task.rb
```diff
@@ -1956,7 +1956,8 @@ def submit_cluster_job
         jobid              = scir_session.run(job)
         self.cluster_jobid = jobid
         self.status_transition(self.status, "Queued")
-        self.addlog("Queued as job ID '#{jobid}'.")
+        self.addlog("Queued as job ID '#{jobid}'" +
+            "with walltime #{job.walltime}s and extra args are '#{job.tc_extra_qsub_args.presence || 'not provided'}'." )
```
prioux, line 1960: 1. add a space before the word 'with'
2. change 'not provided' to '(none provided)'

MontrealSergiy:
corrected

MontrealSergiy:
I do not think we support timeout/walltime for generic Unix workers, if you like I can add comment 'ignored' or 'unsupported by xxx server' for that case. Perhaps with `lfs` too, whatever it stands for.

prioux:
Nah no worries. If there is no walltime, we just get no log entry, which is fine with me.

prioux:
Sure, but maybe one day we will implement "timeout" for unix jobs, so the message can just stay for the moment.

prioux:
man "timeout", it seems to be the same on MacOS as in Linux, so it should be easy to implement.

prioux:
Oh, the reason timeout is the same is because I have brew.

PR #1359, opened 2023-10-25
NeuroHub - component categorical grid

MontrealSergiy:


MontrealSergiy:
while few menu items are removed, bulk of code and pages still exists, as some useful features, like editors are not transferred.

PR #1364, opened 2023-11-24
stay on myaccount page when ssh push fail fixes #1363

### LINKED ISSUE #1363: When a user pushes a key to a Bourreau, if the push fails, they get redirected to the users index page
Users can push their special personal key to the Bourreaux. It's done in the My Account page.

If the Bourreau is disconnected/invalid, an error occurs and the users is redirected to `/users` which is a route not allowed for normal users.

Instead, an error message should be shown and the users should be redirected to the My Account page again.

Note: Handle gracefully the case when the users pushes to several Bourreaux at once.

MontrealSergiy:
see #1363

### DIFF
--- a/BrainPortal/app/controllers/users_controller.rb
+++ b/BrainPortal/app/controllers/users_controller.rb
@@ -408,11 +408,13 @@ class UsersController < ApplicationController
                                            :ssh_key_pub       => pub_key,
                                            :ssh_key_priv      => priv_key,
                                           )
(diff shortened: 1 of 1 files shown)


MontrealSergiy:
added additional 'we are looking into it' message to calm down user, exception log

prioux:
This PR's code fix (which was ok) has already been handled by code from Natacha when she worked on something else.

PR #1365, opened 2023-12-08
auto-refresh config's boutiques descriptor  #1174

### LINKED ISSUE #1174: Add ability to force a reload of a descriptor in the new integrator
This is just a feature, but given that the new integrator stores descriptors in memory at boot time, maybe having the ability to force a reload of all of them (just replacing their content) would be nice. This would allow the admin to edit a descriptor in the installation, click the reload button and then the new descriptor takes effect. Maybe an admin feature in the interface somewhere.

Notes:

1) Maybe a new Control message could be sent to a Bourreau do perform the same reload
2) Descriptors loaded by the integrator have `from_file()` method that is useful
3) The admin must realize that eloading won't affect existing ToolConfigs and Tools (e.g. their container config attributes)

MontrealSergiy:
Auto-refresh tool config boutiques descriptor by checking the json file timestap.

Not sure how well scales, perhaps create race conditions (task started with old boutiques, then switched to a new one).

resolves #1174

### DIFF
--- a/BrainPortal/app/models/tool_config.rb
+++ b/BrainPortal/app/models/tool_config.rb
@@ -593,15 +593,21 @@ class ToolConfig < ApplicationRecord
   def self.registered_boutiques_descriptor(tool_name, tool_version) #:nodoc:
     @_descriptors_ ||= {}
     key = [ tool_name, tool_version ] # two strings
+    @_descriptors_[key] = @_descriptors_[key]&.reload_if_file_timestamp_changed
     @_descriptors_[key]
   end
 
   def boutiques_descriptor
+    if @_descriptor_
+      @_descriptor_       = @_descriptor_.reload_if_file_timestamp_changed
+      key                 = [ self.tool.name, self.version_name ] # two strings
+      @_descriptors_[key] = @_descriptor_
+      return @_descriptor_
+    end
     path = boutiques_descriptor_path.presence
     if ! path
       return self.class.registered_boutiques_descriptor(self.tool.name, self.version_name)
     end
-    return @_descriptor_ if @_descriptor_
     path = Pathname.new(path)
     path = Pathname.new(CBRAIN::BoutiquesDescriptorsPlugins_Dir) + path if path.relative?
     @_descriptor_ = BoutiquesSupport::BoutiquesDescriptor.new_from_file(path)
--- a/BrainPortal/lib/boutiques_support.rb
+++ b/BrainPortal/lib/boutiques_support.rb
@@ -52,8 +52,9 @@ module BoutiquesSupport
   # initialize above (top_prop_names etc).
   BoutiquesDescriptor = Class.new(RestrictedHash) do |klass|
 
-    allowed_keys top_prop_names # 'name', 'author' etc
-    attr_accessor :from_file    # not a hash attribute; a file name, for info
+    allowed_keys top_prop_names  # 'name', 'author' etc
+    attr_accessor :from_file     # not a hash attribute; a file name, for info
+    attr_accessor :mtime_of_file # not a hash attribute, a file timestamp, for caching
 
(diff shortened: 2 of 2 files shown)


BrainPortal/app/models/tool_config.rb
```diff
@@ -593,15 +593,21 @@ def self.register_descriptor(descriptor, tool_name, tool_version) #:nodoc:
   def self.registered_boutiques_descriptor(tool_name, tool_version) #:nodoc:
     @_descriptors_ ||= {}
     key = [ tool_name, tool_version ] # two strings
+    @_descriptors_[key] = @_descriptors_[key]&.reload_if_file_timestamp_changed
     @_descriptors_[key]
   end
 
   def boutiques_descriptor
+    if @_descriptor_
```
prioux, line 601: These 6 lines (601 to 606) should be at the same place you removed the line 604 in the original code. Otherwise you'te changing the priority logic

BrainPortal/lib/boutiques_support.rb
```diff
@@ -89,18 +90,26 @@ def self.new_from_string(text)
 
     def self.new_from_file(path)
       obj = self.new_from_string(File.read(path))
-      obj.from_file = path
+      obj.from_file     = path
+      obj.mtime_of_file = File.mtime(path)
       obj
     end
 
+    def reload_if_file_timestamp_changed()
```
prioux, line 98: Remove parens that are unnecessary

BrainPortal/lib/boutiques_support.rb
```diff
@@ -89,18 +90,26 @@ def self.new_from_string(text)
...
+      filepath = self.from_file
+      return self if filepath.blank? || File.mtime(filepath) == self.mtime_of_file
```
prioux, line 100: You cannot truste that you can compare timestamps with `==` ; it is posisble the internal time and the filesystem times do not have the same granularity. Instead compare with something like `(time1 - time2).abs < 1.second`

prioux (changes requested):
Minor

prioux:
Please review my comments.

BrainPortal/lib/boutiques_support.rb
MontrealSergiy: not sure I understand how this can happen. Do you mean ruby rounds up the system file's timestamp differently time from time?

MontrealSergiy: It worked on my machine, and in worth case it is just a refresh, so I did not pay much attention to that.

MontrealSergiy:
I got the comment, but not sure will it work, or at least not causing issues with old style boutiques

BrainPortal/lib/boutiques_support.rb
prioux: OK keep the == comparison.

prioux:
Please address my comments that were made three weeks ago!

BrainPortal/lib/boutiques_support.rb
MontrealSergiy: well I added the comparison,seems work fine. Indeed time may be tricky better be on safe side here

prioux (changes requested):
Minor. I want to remove a feature.

BrainPortal/app/models/tool_config.rb
```diff
@@ -633,7 +634,15 @@ def boutiques_descriptor
     if ! path
       return self.class.registered_boutiques_descriptor(self.tool.name, self.version_name)
     end
-    return @_descriptor_ if @_descriptor_
+
```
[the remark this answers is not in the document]
MontrealSergiy: admin-set paths is seldom used, but seems potentially could be used in some cases, e.g to create tool configs more dynamically, perhaps just with different docker version, qsub or variables. Anyways, removed as requested

PR #1367, opened 2023-12-20
touch cache, its link, md5 and id,  resolves #1282

### LINKED ISSUE #1282: At boot time, a CBRAIN app should touch() some filesystem config dirs and files
Some Bourreaux systems are configured with disk allocations where files older than N days are erased automatically.

To prevent such system from deleting the top-level directories for the DP_Cache, and some cbrain-specific admin files, I suggest that part of the boot process should touch them to reset their timestamps.

On a portal or bourreau:

* the DataProvider cache dir
* the `DP_Cache_Key.md5`  and `DP_Cache_Rev.id` located in that cache dir

On a bourreau:

* the gridshare dir
* the `DP_Cache` symbolic link located in it. WARNING: make sure to update the timestamp of the symlink itself!

MontrealSergiy:
resolves #1282

It maybe safer to recreate the symlink than touch, but I guess should work for us


Bourreau/lib/bourreau_system_checks.rb
```diff
@@ -484,6 +485,44 @@ def self.a100_ensure_dp_cache_symlink_exists #:nodoc:
 
 
 
+  # touch to avoid deletion by cluster bimonthly sctratch cleanup : 1) DataProvider cache dir, 2) DP_Cache_Key.md5 3) DP_Cache_Rev.id
+  # 4) gridshare dir, and 5) DP_Cache symbolic link located in it
+  # gridshare and cache_dir are typically updated often, but we touch them just in the case
+  def self.a105_ensure_dp_cache_and_symlink_will_exists #:nodoc:
```
prioux, line 491: 1. I think this new checker will prevent a new CBRAIN installation from booting, given that it probably doesn't have a DP cache directory already configured. We allow a CBRAIN portal to start even without one, because we expect the admin to configure it with the interface.
2. most of the code in here should be moved to the DataProvider class, alongside other methods that generate/install these files; the checker code should be made way simpler (just report if yes/no the update was performed).

MontrealSergiy:
I believe I added the requested changes

prioux:
For the GridShare dir and the DP cache symlink, I'd rather both 'touch' commands to be run in the bourreau system check in a100_ensure_dp_cache_symlink_exists because this is the ideal place.

For the remaining three DataProvider cache files, the class method in DataProvider is fine as it is (minus the code moved to the place I talked about at the beginning of this comment).

BrainPortal/lib/cbrain_system_checks.rb
```diff
@@ -334,6 +334,24 @@ def self.a050_check_data_provider_cache_wipe #:nodoc:
     end
   end
 
+  # prevents archiving/delete of cbrain system files and top directories, related to
+  # cache and gridshare
+  def self.a060_ensure_system_files_will_not_be_deleted #:nodoc:
+
+    #-----------------------------------------------------------------------------
+    puts "C> Updating timestamp for important system files and directories"
```
prioux, line 342: Change message, because 'important system files and directories' is much too general. Say what is happening for real:

```
Updating DataProvider cache administrative files
```

BrainPortal/app/models/data_provider.rb
```diff
@@ -1353,7 +1353,43 @@ def self.cleanup_leftover_cache_files(do_it=false, options={})
...
+    cache_dir    = myself.dp_cache_dir
+    dp_cache_id  = File.join cache_dir, DataProvider::DP_CACHE_ID_FILE
+    dp_cache_md5 = File.join cache_dir, DataProvider::DP_CACHE_MD5_FILE
+
+    FileUtils.touch [cache_dir, dp_cache_id, dp_cache_md5], verbose: true, nocreate: true
 
+    # touch only cache for Portal, for Bourreau touch gridshare
+    return true unless myself.is_a? Bourreau
+
+    gridshare_dir = myself.cms_shared_dir
```
prioux, line 1384: Please move the logic of lines 1384-1391 to the file `Bourreau/lib/bourreau_system_checks.rb` within the method `self.a100_ensure_dp_cache_symlink_exists` like I asked in my previous comments.

prioux (changes requested):
See my two comments of today

Bourreau/lib/bourreau_system_checks.rb
```diff
@@ -289,15 +289,22 @@ def self.a100_ensure_dp_cache_symlink_exists #:nodoc:
...
 
     return unless Dir.exists?(gridshare_dir) && Dir.exists?(cache_dir)
 
     #----------------------------------------------------------------------------
     puts "C> Making sure the grid share directory has a symlink to the data provider cache..."
     #----------------------------------------------------------------------------
 
-    sym_path = "#{gridshare_dir}/#{DataProvider::DP_CACHE_SYML}"
-    return if File.symlink?(sym_path) && File.realpath(sym_path) == File.realpath(cache_dir)
+    # update the work directory timestamp to counter cluster deletion policies
```
prioux, line 300: Why is this so complicated?!??

The original code's logic was:

```
return if symlink_is_ok
create symlink
```

The new code should simply be instead:

```
if symlink_is_ok
  touch symlink
  return
end
create symlink
```

But you did a complex set of conditions including checking the return status of system():

```
touch_but_not_create symlink
ret=system(touch symlink)
return if ret is 0 and symlink_is_ok
create_symlink
```

Why is make it so complicated?

BrainPortal/app/models/data_provider.rb
```diff
@@ -1360,6 +1360,22 @@ def self.rsync_ignore_patterns #:nodoc:
...
   end
 
+  # Some Bourreaux systems are configured with disk allocations where files older than N days are erased automatically.
+  # To prevent such system from deleting the top-level directories for the DP_Cache, and some cbrain-specific files,
+  # the system can reset their timestamps using the touch command.
+  #
+  # - the +DataProvider+ cache dir
+  # - the +DP_Cache_Key.md5+ and
+  # - +DP_Cache_Rev.id+ located in that cache dir
+  def self.system_touch
```
prioux, line 1370: The method name `system_touch` is way too general. The method touches two special config files related to the cache subsystem. In general, in the DataProvider class, all methods that deal with the cache subsystem have "cache_" in their names. So please rename this

```ruby
def self.touch_cache_md5
```

and place it after `create_cache_md5` where it belongs.

prioux (changes requested):
See comments

Bourreau/lib/bourreau_system_checks.rb
```diff
@@ -289,15 +289,22 @@ def self.a100_ensure_dp_cache_symlink_exists #:nodoc:
...
+    FileUtils.touch(gridshare_dir, verbose: true, nocreate: true)
+
+    # update timestamp for the symbolic link rather than the folder it points to
+    # --no-dereference works on most major os, otherwise link recreated below
+    link_updated = system("touch", "--no-dereference", sym_path)
```
prioux, line 305: Adendum, `--no0derefence` doesn't work on Darwin

Bourreau/lib/bourreau_system_checks.rb
MontrealSergiy: To make code work even if touch fails. But ok, can delete.

prioux:
I am closing this PR without merging because it's taking too long with too many back and forth.

I replaced it with my own pull request, #1487 which shows how it's supposed to be done.

Issue #969, opened 2020-07-23
Add new consistency check at boot time: Userfile and CbrainTask classes verification


prioux:
Add a new boot-time validation method to check that all the 'type' columns of our single table inheritance models represent ruby classes that are properly loaded.

For tasks, that would be:

```ruby
not_loaded = CbrainTask.group(:type).pluck(:type).select do |name|
    (name.constantize rescue nil) ? nil : name
end
```

MontrealSergiy:
@prioux  Actually there is a benefit in **not** having that feature. I one works on several usefile/task features in the same time, it is not necessary to wipe up to database to work on each feature branch.

Well, if still needed we can add, please confirm

MontrealSergiy:
@prioux can you clarify 

> "ID matching the current app
rows which have an ID matching the current app"

Do you suggest restrict by  `bourreau_id == current_bourreau.id` ?

> 
> Since not all CbrainTask classes are deplpyed on the Bourreaux, the check for the types for the cbrain_tasks table should be restricted to those rows which have an ID matching the current app.
>

prioux:
Yes exactly.

PR #1369, opened 2024-02-01
new consistency check at boot time: Userfile and CbrainTask classes STI  #969

MontrealSergiy:
see #969

### DIFF
--- a/BrainPortal/lib/cbrain_checker.rb
+++ b/BrainPortal/lib/cbrain_checker.rb
@@ -82,16 +82,11 @@ class CbrainChecker
   # already loaded objects; on mild issues (forgot run rake install etc...)
   # in developer env gives warning but fails execution in production or test environment
   def self.validate_sti(query)
-
     table = query&.table_name || 'a table'
-
-    #----------------------------------------------------------------------------
-    puts "C> Checking STI column of #{table} ..."
-    #----------------------------------------------------------------------------
-
-    level    = Rails.env.development? ? 'Warning' : 'Error'
+    level = Rails.env.development? ? 'Warning' : 'Error'
+    puts  "C> \t - Checking table #{table}"
     messages = query.distinct.pluck(:type).map do |name|
-      "C> \t- #{level}: Invalid STI type: " + name unless name.safe_constantize
+      "C> \t\t #{level}: Invalid STI type: #{name || '(blank value)'}" unless name.present? && name.safe_constantize
     end.compact
     puts messages
     raise "Invalid types in the STI column of #{table}" unless messages.empty? || level == "Warning"
--- a/BrainPortal/lib/cbrain_checker.rb
+++ b/BrainPortal/lib/cbrain_checker.rb
@@ -76,5 +76,26 @@ class CbrainChecker
 
   end
 
+  protected
+
+  # helper method to runs a query and checks that type column contains
+  # already loaded objects; on mild issues (forgot run rake install etc...)
+  # in developer env gives warning but fails execution in production or test environment
+  def self.validate_sti(query)
+
+    table = query&.table_name || 'a table'
+
+    #----------------------------------------------------------------------------
+    puts "C> Checking STI column of #{table} ..."
+    #----------------------------------------------------------------------------
+
+    level    = Rails.env.development? ? 'Warning' : 'Error'
+    messages = query.distinct.pluck(:type).map do |name|
+      "C> \t- #{level}: Invalid STI type: " + name unless name.safe_constantize
+    end.compact
+    puts messages
+    raise "Invalid types in the STI column of #{table}" unless messages.empty? || level == "Warning"
+  end
+
 end
 
--- a/BrainPortal/lib/portal_system_checks.rb
+++ b/BrainPortal/lib/portal_system_checks.rb
@@ -290,17 +290,15 @@ class PortalSystemChecks < CbrainChecker #:nodoc:
   def self.z020_ensure_task_sti_enabled #:nodoc:
 
     #----------------------------------------------------------------------------
-    puts "C> Checking CbrainTask STI column"
+    puts "C> Checking CbrainTask STI column ..."
     #----------------------------------------------------------------------------
 
     level = Rails.env.development? ? 'Warning' : 'Error'
-    not_loaded = CbrainTask.distinct.pluck(:type).select do |name|
-      name.safe_constantize.nil?
-    end.each do |name|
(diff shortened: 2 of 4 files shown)


MontrealSergiy:
I think I can further shorten code but factoring out common elements of two checks into another method. Also if changes will be needed (e.g. rollback to using rescue requested) I would need less change  ... so if it is ok wait a sec

MontrealSergiy:
added a generic STI checking methods and a task check for Bourreau

prioux:
I think this PR is fine. I would add more check on more tables however:

* groups
* custom_filters
* users
* remote_resources
* data_providers
* background_activities
* resource_usage
* messages
* disk_quotas

(All of these, only during the portal system checks)

prioux:
I would also modify the traces. The banned

```
#----------------------------------------------------------------------------
puts "C> Checking STI column of #{table} ..."
#----------------------------------------------------------------------------
```

shoudl appear only once in the script, and print only a single line for all STI checks, then be followed by a single log lone per tabel checks (each optionally followed by their warning messages if any). Something like

```
C> Checking Single Table Inheritance classes ...
C>    - Checking table 'userfiles'
C>    - Checking table 'cbrain_tasks'
C>       ERROR: type 'CbrainTask::StupidTool' is not defined!
C>    - Checking table 'blahblah'
```

prioux:
Still waiting for updates.

MontrealSergiy:
disk_quota is not yet sti.

prioux:
I think overall the code is ok, I'm just a bit not sure for the overall structure of where the code is run.

1. Why is the Userfile table checked separately? We'll get twice the exact same boot message "C> Checking Single Table Inheritance classes"
2. Why not do all the checks at the end of the common "cbrain_system_checks.rb", so that both Bourreaux and Portal verify everything? That way we don't modify at all the portal_system_checks, nor the bourreau_system_checks. I understand we wanted the Bourreau code to check only the rows assigned to itself, but that can be done in the cbrain_stystem_checks with something like

```ruby
   (myself.is_a?(Bourreau) ? CbrainTask.where(:bourreau_id => myself.id) : CbrainTask)
```

in the list of models to pass to validate_sti()

BrainPortal/lib/cbrain_checker.rb
```diff
@@ -76,5 +76,21 @@ def self.check(checks_to_run, options={})
 
   end
 
+  protected
+
+  # helper method to runs a query and checks that type column contains
+  # already loaded objects; on mild issues (forgot run rake install etc...)
+  # in developer env gives warning but fails execution in production or test environment
+  def self.validate_sti(query)
+    table = query&.table_name || 'a table'
```
prioux, line 85: table should never be blank, and .table_name should never be invoked on a nil so that line should be made straighforwardly:

```ruby
  table = query.table_name
```

BrainPortal/lib/portal_system_checks.rb
```diff
@@ -285,5 +285,26 @@ def self.z020_start_background_activity_workers #:nodoc:
 
   end
 
+
+
+  def self.z040_ensure_sti_enabled #:nodoc:
```
prioux, line 290: replace "enabled" with "valid" because it's not like STI is something that can be enabled or disabled

prioux (changes requested):
See comments

prioux:
So I just verified that the class check code cannot go into the common cbrain_system_checks because the models aren't yet fully loaded at that time.

prioux:
This is what I cam up with instead:

```diff
diff --git a/BrainPortal/lib/portal_system_checks.rb b/BrainPortal/lib/portal_system_checks.rb
index 857dc4bdb..63b4d3325 100644
--- a/BrainPortal/lib/portal_system_checks.rb
+++ b/BrainPortal/lib/portal_system_checks.rb
@@ -95,6 +95,40 @@ class PortalSystemChecks < CbrainChecker #:nodoc:
 
 
 
+  def self.a025_check_single_table_inheritance_types
+
+    #-----------------------------------------------------------------------------
+    puts "C> Checking 'type' columns for single table inheritance..."
+    #-----------------------------------------------------------------------------
+
+    myself      = RemoteResource.current_resource
+    err_or_warn = Rails.env == 'production' ? 'Error' : 'Warning'
+    errcount    = 0
+
+    [
+      Userfile,
+      (myself.is_a?(Bourreau)? myself.cbrain_tasks : CbrainTask),
+      CustomFilter,
+      User,
+      RemoteResource,
+      DataProvider,
+      BackgroundActivity,
+      ResourceUsage,
+      Invitation,
+    ].each do |query|
+      badtypes = query.distinct(:type).pluck(:type).compact
+        .reject { |type| type.safe_constantize }
+      next if badtypes.empty?
+      puts "C> \t- #{err_or_warn}: Bad type values in table '#{query.table_name}': #{badtypes.join(', ')}"
+      errcount += 1
+    end
+
+    raise "Single table inheritance check failed for #{errcount} tables" if Rails.env == 'production' && errcount > 0
+
+  end
+
+
+
   def self.a030_ensure_we_can_load_oidc_config
 
       #----------------------------------------------------------------------------
```

prioux:
About my patch: the check method should work fine at a025 for both portals and bourreaux; the checking code could be extracted but I don't mind repeating the check in both files too.

prioux:
About my patch 2: it worked on the Bourreau side too, except a) there is no Invitation model on the bourreau side (by design) and 2) my fancy `myself.cbrain_tasks` needs to be replaced with a `CbrainTask.where(:bourreau_id => myself.id)`

prioux:
Given I've got a solution, I'll just commit it and close this PR

BrainPortal/lib/cbrain_checker.rb
```diff
@@ -76,5 +76,21 @@ def self.check(checks_to_run, options={})
...
+    level = Rails.env.development? ? 'Warning' : 'Error'
+    puts  "C> \t - Checking table #{table}"
+    messages = query.distinct.pluck(:type).map do |name|
+      "C> \t\t #{level}: Invalid STI type: #{name || '(blank value)'}" unless name.present? && name.safe_constantize
```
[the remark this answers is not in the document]
MontrealSergiy: Agree using both `name.present?` and `blank value` overkill. But I would keep at the least blank value.
It might happens with some, perhaps buggy tables, such as `messages`

Please confirm

prioux:
All implemented by me in https://github.com/aces/cbrain/commit/93d0473fa1c8b6480025e1f30366b5905e97e3f0

PR #1371, opened 2024-02-08
[ADD] DataProvider API endpoint

natacha-beck:


prioux:
Rejected after discussions

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -98,7 +98,7 @@ def create  #:nodoc:
       respond_to do |format|
         format.html { redirect_to :action => :index, :format => :html}
         format.xml  { render :xml   => @provider }
-        format.json { render :json  => @provider }
+        format.json { render :json  => @provider.for_api }
```
prioux, line 101: Can you add the `.for_api` for the XML request too please? it's just that you made me realize it needs to be there also!

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -115,22 +115,44 @@ def new_personal #:nodoc:
...
+                                  :cloud_storage_region,
                                  )
+
+    authorized_type = [UserkeyFlatDirSshDataProvider, S3DataProvider, S3FlatDataProvider, S3MultiLevelDataProvider]
+    dp_type         = normal_params[:type] || "UserkeyFlatDirSshDataProvider"
+
+    # Check if the type is an authorized class
+    if ! authorized_type.include?(dp_type.constantize)
+      respond_to do |format|
+        format.html { render :action => :new_personal}
```
prioux, line 145: If the type is wrong, you just redirect to the NEW page without any message. Add a message to :base please

BrainPortal/public/swagger/cbrain-6.2.0.1-swagger.json
```diff
@@ -517,6 +517,30 @@
             "description" : "No session created yet."
           }
         }
+      },
```
prioux, line 520: Please undo all changes to this JSON file. Instead, provide a NEW pair of YAML and JSON file with a new version number (e.g. 6.2.0.2). We also need to publish this on SwaggerHub

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -115,22 +115,44 @@ def new_personal #:nodoc:
...
+                                  :type,
+                                  :cloud_storage_client_identifier,
+                                  :cloud_storage_client_token,
+                                  :cloud_storage_client_bucket_name,
+                                  :cloud_storage_client_path_start,
+                                  :cloud_storage_endpoint,
+                                  :cloud_storage_region,
                                  )
+
+    authorized_type = [UserkeyFlatDirSshDataProvider, S3DataProvider, S3FlatDataProvider, S3MultiLevelDataProvider]
```
prioux, line 139: Do NOT support the S3DataProvider, this is a very old implementation that may not even work anymore.

prioux (changes requested):
Minor changes.

BrainPortal/app/controllers/data_providers_controller.rb
prioux: You still haven't added an active record error to :base

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -115,22 +115,45 @@ def new_personal #:nodoc:
...
+        format.html { render :action => :new_personal}
+        format.json { render :json   => { :error => "DataProvider type is not authorized" },  :status => :unprocessable_entity }
+      end
+    end
+
     group_id = normal_params[:group_id]
     current_user.assignable_group_ids.find(group_id) # ensure assignable, not sure need check visibility etc more
-    @provider         = UserkeyFlatDirSshDataProvider.new(normal_params)
-    @provider.user_id = current_user.id # prevent creation of dp on behalf of other users
+    @provider          = dp_type.constantize.new(normal_params)
```
prioux, line 153: This code will set a bunch of attributes that are meaningless depending on the type. E.g. for a UserKey DP, this will allow a user to set the attributes for a cloud storage DP, and vice versa. There should be two params authorization methods, one for each type of DP, each with their own appropriate white list of attributes

prioux (changes requested):
Controller params need better validation based on DP type

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -115,22 +115,45 @@ def new_personal #:nodoc:
...
+    if ! authorized_type.include?(dp_type)
+      flash[:error] = "DataProvider type is not authorized"
+      respond_to do |format|
+        format.html { render :action => :new_personal}
+        format.json { render :json   => { :error => "DataProvider type is not authorized" },  :status => :unprocessable_entity }
+      end
+    end
+
+    # Whitelist params
+    whitelist_params = dp_type == UserkeyFlatDirSshDataProvider ?
```
prioux, line 138: Rewrite the next 12 lines to call a separete method to white list the params, one for UserKey DPs and one for S3 DPs.

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -97,8 +97,8 @@ def create  #:nodoc:
       flash[:notice] = "Provider successfully created."
       respond_to do |format|
         format.html { redirect_to :action => :index, :format => :html}
-        format.xml  { render :xml   => @provider }
-        format.json { render :json  => @provider }
+        format.xml  { render :xml   => @provider.for_api }
+        format.json { render :json  => @provider.for_api }
```
prioux, line 101: While we're invoking .for_api() on these objects, shouldn't we configure the attributes that we would liked returned back to the client, based on DP type? E.g. for a UserKey DP, return the remote host, user, and path? For a S3 DP, the path, bucket name etc? (Never return any tokens of course)

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -115,22 +115,41 @@ def new_personal #:nodoc:
...
-    current_user.assignable_group_ids.find(group_id) # ensure assignable, not sure need check visibility etc more
-    @provider         = UserkeyFlatDirSshDataProvider.new(normal_params)
-    @provider.user_id = current_user.id # prevent creation of dp on behalf of other users
+    dataprovider_params = params[:data_provider]
+
+    # Check if the type is an authorized class
+    dp_type             = (dataprovider_params[:type] || "UserkeyFlatDirSshDataProvider").constantize
+    authorized_type     = [UserkeyFlatDirSshDataProvider, S3FlatDataProvider, S3MultiLevelDataProvider]
+
+    if ! authorized_type.include?(dp_type)
```
prioux, line 132: Make a check on the type here means that when the type is wrong, and the form is rendered again, the user will lose all the values that they entered in the form. It's not much of a problem because of course the form normally doesn't allow the user to change the type to something wrong, but it's something to think about. The normal workflow for a create action in rails is 1) create the object 2) validate 3) re-render the form if validation fails 4) save the object in database if validation succeeds

prioux:
Line 161:

```ruby
format.json { render      :json   => @provider }
```

There is no .to_api() method invoked, which means the returned objects contains sensitive information (e.g. tokens) that should never be returned back on the API channel.

prioux (changes requested):
Still improvements to make

BrainPortal/app/models/data_provider.rb
```diff
@@ -1055,14 +1055,34 @@ def alive_cached_valid?
   #     is_fast_syncing                      => true/false
   #     content_storage_shared_between_users => true/false
   def for_api
+    extra_info = {};
```
prioux, line 1058: Don't do that.

Overriding for_api() is only necessary when we need to provdie information that is not directly an attribute. E.g. in the original case, this was here because `is_browsable` is not an actual attribute of the model, so we had to invoke the method at merge it to the hash returned by the standard for_api().

For normal attributes, just add the attributes to `api_attr_visilble` in each class. This class method builds lists of attributes to return and also inherits automatically attributes defined in superclasses (so you don't have to repeat user_id for exmaple)

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -60,6 +60,8 @@ class S3FlatDataProvider < DataProvider
 
   before_save :canonify_path_start
 
+  api_attr_visible :cloud_storage_client_identifier, :cloud_storage_client_bucket_name, :cloud_storage_client_path_start, :cloud_storage_endpoint, :cloud_storage_region
```
prioux, line 63: Do not return the client_identifier !!!!

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -115,25 +115,31 @@ def new_personal #:nodoc:
...
-                                  :remote_user, :remote_host,
-                                  :remote_port, :remote_dir
-                                 )
-    group_id = normal_params[:group_id]
-    current_user.assignable_group_ids.find(group_id) # ensure assignable, not sure need check visibility etc more
-    @provider         = UserkeyFlatDirSshDataProvider.new(normal_params)
-    @provider.user_id = current_user.id # prevent creation of dp on behalf of other users
+    dp_type   = (params[:data_provider][:type] || "UserkeyFlatDirSshDataProvider").constantize
+
+    @provider = dp_type.new(base_provider_params)
```
prioux, line 128: Why extract the DP type and not let the STI code handle the tpying directly from the params?

```ruby
@provider = DataProvider.new(base_provider_params)
```

would work even better given `:type` is already one of the attributes returned by `common_params_list()`

BrainPortal/app/controllers/data_providers_controller.rb
```diff
@@ -1065,4 +1071,36 @@ def generate_register_response #:nodoc:
...
+
+  def userkey_provider_params #:nodoc:
+    params.require_as_params(:data_provider).permit(userkey_params_list)
+  end
+
+  def s3_provider_params #:nodoc:
+    params.require_as_params(:data_provider).permit(s3_params_list)
+  end
+
+  def filtered_create_personal_params(params,type) #:nodoc:
```
prioux, line 1086: What is this method used for?

prioux (changes requested):
Handlign of dp_type is still clumsy, see my suggestion

PR #1377, opened 2024-02-20
[ADD] switch user via API

natacha-beck:


BrainPortal/app/controllers/users_controller.rb
```diff
@@ -346,7 +372,10 @@ def switch #:nodoc:
     # ... but we must adjust the CBRAIN session object too
     cbrain_session.user_id = @user.id
 
-    redirect_to start_page_path
+    respond_to do |format|
```
prioux, line 375: Why is this changed if 'switch' is not even an available API call ?

BrainPortal/app/controllers/users_controller.rb
```diff
@@ -328,6 +329,31 @@ def destroy #:nodoc:
...
+  # Only use via API and accessible only to admin users
+  def create_user_session #:nodoc:
+    for_user_id = params[:id]
+    for_user    = User.find(for_user_id)
+
+    new_user_session = LargeSessionInfo.new(
+      user_id:    for_user_id,
+      session_id: SecureRandom.hex,
+      active: true,
+      data: {
```
prioux, line 341: It's a bit weird that you are mixing the style

```ruby
:symbol => value
```

and

```ruby
symbol: value
```

prioux (changes requested):
Minor changes.

MontrealSergiy:
code seems ok, though not clear why the API is needed

MontrealSergiy:
should I test it?

BrainPortal/app/controllers/users_controller.rb
```diff
@@ -328,6 +329,31 @@ def destroy #:nodoc:
     end
   end
 
+  # Only use via API and accessible only to admin users
```
MontrealSergiy, line 332: I would rephrase in polite  "Please, use only via API ..." 
or non-prescriptive manner "API-only action ..."

Issue #1128, opened 2021-06-09
Document somewhere the special cbrain flag for Boutiques descriptors


prioux:
The CBRAIN Boutiques integrator supports a set of special flags in the `"custom"` field of the descriptor.

Document them all somewhere?

At the top of my head:

```json
"custom": {
  "cbrain:author": "descriptor author",
  "cbrain:inherits-from-class": "ClassName",
  "cbrain:can-submit-new-tasks: truefalse,
  "cbrain:ignore-exit-status": truefalse,
  "cbrain:ignore_outputs": [id, id, id],
  "cbrain:save_back_inputs": [id, id, id],
  "cbrain:readonly-input-files": truefalse,
  "cbrain:alters-input-files": truefalse
}
```

MontrealSergiy:
I do not find `cbrain:alters-input-files`

Yet see

- `cbrain:authors`
- `cbrain:boutiques_bosh_exec_mode`
- `cbrain:walltime-estimate`
- `cbrain:inherits-from-class`
- also multiple `cbrain:integrator_modules

Could go somewhere into developer's wiki, readme of cbrain-neuro-plugins or cbrain.ca. Also I believe @natacha-beck had a wiki on boutiques integration, not sure why it did not get into the main wiki

natacha-beck:
@MontrealSergiy here is the page you are refering: https://github.com/natacha-beck/cbrain-wiki/wiki/Tasks-Integration-With-Boutiques 

It was document about old Boutiques integration. We should at least have documentation about both way to integrate Boutiques (in cbrain_task_descriptors and in boutiques_descriptors).

MontrealSergiy:
actually there is a designated CBRAIN Boutiques wiki section now, is it a good place for it?
https://github.com/aces/cbrain/wiki/Tool-integration-with-Boutiques-or-Scientific-tool-integration-in-CBRAIN

MontrealSergiy:
The current draft documentation is at https://docs.google.com/spreadsheets/d/1GdetgwsxfX9dXDXpqlwi_CU2_Shvn1u1L5yVcZU31g0

prioux:
I want this documentation to be part of the codebase.

I am not sure which file is more appropriate, but given we have a library for loading and accessing Boutiques descriptor files, maybe we should put it there.

A larger comment block should enumerate them all in a table. Then we could have a nice new set of accessor methods for each of the special flags, something like

```ruby
# Return the CBRAIN author of the descripto from the custom field
[...]

PR #1378, opened 2024-02-20
Document cbrain specific boutiques properties (extensions) #1128

MontrealSergiy:
see #1128

### DIFF
--- a/BrainPortal/public/doc/boutiques_custom_properties/cbrain_properties.html
+++ b/BrainPortal/public/doc/boutiques_custom_properties/cbrain_properties.html
@@ -1,0 +1,250 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>TABLE OF CBRAIN SPECIFIC BOUTIQUES CUSTOM PROPERTIES</title>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <style>
+        table {
+            border-collapse: collapse;
+            width: 100%;
+        }
+
+        th, td {
+            border: 1px solid #dddddd;
+            text-align: left;
+            padding: 8px;
+        }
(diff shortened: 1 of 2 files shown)


prioux (changes requested):
This doc is good but it has indentations problems (e.g. line 228) and overall it should match the conventions for RDOC blocks. Try to put your new large comment block at the beginning of a method and generate the rdoc with the rake tasks, and inspect the results with yoru browser. Make sure it looks good!

I would prefer all your new methods that end with "_custom" in their names ot START with "custom_" instead. E.g. instead of `allow_empty_strings_custom`, use `custom_allow_empty_strings`.

MontrealSergiy:
I'll add a few directives to facilitate automated generation of documentation by rdoc, and, following @natacha-beck suggestion, move the property table into a separate file

MontrealSergiy:
Tested with rdoc 6.6.2, please check now  @prioux @natacha-beck

prioux:
So I understand that adding comments in the BoutiquesSupport module was made difficult because many classes, instead of being declared like the standard:

```ruby
class BoutiquesDescriptor < RestrictedHash
end
```

were instead declared with meta programming, as in:

```ruby
BoutiquesDescriptor = Class.new(RestrictedHash) do
end
```

So the RDOC parser couldn't really identify what methods were in which class.

Well, I solved that problem by modifying boutiques_support.rb and re-coding all the declarations
in the standard way!

The commit for my change are in master and can be seen in

https://github.com/aces/cbrain/commit/bde084416df4819517aecb3bb7fc158caf89aace

Please rebase to master, take the documentation in this PR and just move it to their proper methods, in
the standard RDOC way. It could be easier to just start with the master version of the code and
re-introduce the comment blocks one by one, because otherwise there will be tons of merge conflicts.

PR #1379, opened 2024-02-26
[FIX] Copying file issue

### LINKED ISSUE #1374: Cannot copy a file in a subfolder from a S3MultiLevelDataProvider to a UserkeyFlatDirSshDataProvider
When trying to copy a file that was register from a S3MultiLevelDataProvider in a subfolder: 

`sub123/abc.txt`

to a UserkeyFlatDirSshDataProvider I have a strange error message: 

```
[2024-02-13 18:21:10 UTC] For 1 file, Did not get a proper numeric ID? Got: 'nil'.:
abc.txt
```

and the copy is not performed.

natacha-beck:
[FIX] issue when copying file from DP with browse capablity to one th…at do not have it

Fixed #1374

### DIFF
--- a/BrainPortal/app/models/data_provider.rb
+++ b/BrainPortal/app/models/data_provider.rb
@@ -860,7 +860,7 @@ class DataProvider < ApplicationRecord
     newfile.updated_at       = Time.now
     newfile.immutable        = false
     # Check if DP has_browse_path_capabilities? if not, set browse_path to nil
-    newfile.browse_path     = nil if !DataProvider.find(otherprovider.id).has_browse_path_capabilities?
+    newfile.browse_path     = nil if !otherprovider.has_browse_path_capabilities?
     newfile.save
 
     # Trigger callbacks for tracking size changes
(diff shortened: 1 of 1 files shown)


BrainPortal/app/models/data_provider.rb
```diff
@@ -859,6 +859,8 @@ def provider_copy_to_otherprovider(userfile, otherprovider, options = {})
     newfile.created_at       = Time.now      unless target_exists
     newfile.updated_at       = Time.now
     newfile.immutable        = false
+    # Check if DP has_browse_path_capabilities? if not, set browse_path to nil
+    newfile.browse_path     = nil if !DataProvider.find(otherprovider.id).has_browse_path_capabilities?
```
prioux, line 863: Why do `DataProvider.find(otherprovider.id)` ?!? Isn't that ridiculous?

natacha-beck: Indeed it is.

PR #1380, opened 2024-02-26
[ADD] ssh key to /user/id

natacha-beck:


BrainPortal/app/controllers/users_controller.rb
```diff
@@ -99,7 +99,8 @@ def show #:nodoc:
         render :xml  => @user.for_api_xml
       end
       format.json do
-        render :json => @user.for_api
+        # Append the SSH key to the JSON response
+        render :json => @user.for_api.merge(public_key: @ssh_key.public_key)
```
prioux, line 103: This will crash is @ssh_key is nil

prioux:
Please fix the tests.

Issue #1373, opened 2024-02-12
The attribute `time_of_death` is no longer used in DataProvider model.


prioux:
Remove it from schema.

MontrealSergiy:
Note, if you running the neurohub grid branch on port 4000 the update might require debase of that brunch

PR #1381, opened 2024-02-27
Remove time_of_death attribute from Data Provider 

MontrealSergiy:
see #1373


prioux:
You schema update changes the collation character set for the tables, please don't do that. Ensure that your DB is using the proper collation keywords before you run your migration so that it doesnt' change (You'll probably have to update  manually your tables with a `alter table` command in your SQL client)

prioux (changes requested):
Collation are changed and they shouldn't.

MontrealSergiy:
the collation is corrected

prioux:
This PR is fine.

PR #1382, opened 2024-03-04
Copy file endpoint

natacha-beck:


BrainPortal/app/controllers/bourreaux_controller.rb
```diff
@@ -610,6 +610,29 @@ def rr_access_dp
 
   end
 
+  # API method to copy files from one DP to another via a bourreau
+  def file_copy #:nodoc:
```
prioux, line 614: Rejected for security. This action has no validation or verifications. It will copy any files to anywhere. Which means that if we deployed this, a normal user could copy any file, including those that belongs to the admins or other users, to their own data provider.

prioux:
Rejected. The role of a controller action is to make sure all parameters are valid for the current session!

prioux:
Check legitimate access for

1. the DP
2. the Bourreau
3. the userfiles (for these, there is a loop, replace this with a mass accessible check instead of going to the database for each and every file ID)

prioux:
For point 3 of my previous comment, there is an excellent example already in the controllers, e.g. in `userfiles_controller` for the action `manage_compression`:

```ruby
    userfiles = Userfile
      .find_all_accessible_by_user(current_user, :access_requested => :write)
      .where(:id => file_ids)
```

BrainPortal/app/controllers/bourreaux_controller.rb
```diff
@@ -622,14 +622,13 @@ def file_copy #:nodoc:
...
-      userfile = Userfile.find(userfile_id)
-      userfile.can_be_accessed_by?(current_user)
-    end
-
+    # Filter out userfile_ids that are not readable by the user
+    userfile_ids = Userfile.find_all_accessible_by_user(current_user, :access_requested => :read)
+                        .where(:id => userfile_ids).pluck(:id)
+    
     bourreau.send_command_copy_files(userfile_ids, data_provider_id, current_user.id)
     render :json => { :status => "ok" }
```
prioux, line 633: While you are at it, when not return some useful information in the JSON reply, like the number of files that were actually selected for the move? (`userfile_ids.size`) ?

prioux:
Looks good to me. Before I merge, I want to discuss if this action is really a 'bourreau' action, or if it would be more appropriate in a different controller. We are selecting file to be copied to a data provider, so maybe it's userfile action, or a data provider action?

PR #1385, opened 2024-04-15
Background activity

prioux:


BrainPortal/app/models/background_activity.rb
```diff
@@ -0,0 +1,608 @@
...
+  ###########################################
+
+  # Crushes the items attribute and replaces it
+  # with a special unique value to represent that
+  # the items list will be filled later, dynamically
+  def configure_for_dynamic_items!
+    self.items = [ DYNAMIC_TOKEN ]
+  end
+
+  def is_configured_for_dynamic_items?
```
MontrealSergiy, line 471: why no doc or nodoc?

BrainPortal/app/models/background_activity/archive_task_workdir.rb
```diff
@@ -0,0 +1,47 @@
...
+# Archive a CBRAIN task.
+#
+# Options: archive_data_provider_id.
+#
+# Must be run on a Bourreau only.
+class BackgroundActivity::ArchiveTaskWorkdir < BackgroundActivity::TerminateTask
+
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+  def process(item)
```
MontrealSergiy, line 32: often no docs

BrainPortal/app/controllers/application_controller.rb
```diff
@@ -297,6 +298,21 @@ def disable_cookies_for_api
   # CBRAIN Messaging System Filters
   ########################################################################
 
+  # Count the number of active BackgroundActivities
+  def count_background_activities #:nodoc:
+    return unless current_user
+    return if     request.format.blank? || request.xhr?
```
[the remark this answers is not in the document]
prioux: Not exactly. The next line would crash if `request.format` was nil, because `nil.to_sym` doesn't work. For sure we could rewrite the next line as `return unless request.format&.to_sym == :html` but the original code this was copied from came from older ruby that didn't have `&.`

BrainPortal/app/models/background_activity/archive_task_workdir.rb
prioux: Yes, I plan to add this later when I finish the full integration on the user-side

BrainPortal/app/controllers/application_controller.rb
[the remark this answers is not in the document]
MontrealSergiy: fair enough, even though I am a bit surprised that nil format exists, I thought that bug was fixed [long time ago](https://github.com/rails/rails/pull/8084).  Also I just realized that `request.xhr?` might be or become  different `request.format.xhr?`. Anyways I would merge two lines into 

```return if  request.format.blank? || ! request.format.html? || request.xhr?```

which would keep blank check near html check. and would avoid unless, which in English usually used account for abnormal situations. (So I do not mind

File: BrainPortal/app/views/background_activities/new.html.erb [hunk 4]
```diff
@@ -0,0 +1,258 @@
...
+    <%= user_select("background_activity[user_id]", :selector => @bac.user_id.to_s) %>
+
+    <p><%= f.label :remote_resource_id, "Portal or Execution Server" %><br>
+    <%= bourreau_select("background_activity[remote_resource_id]", {
+       :bourreaux => RemoteResource.all.order(:name).to_a,
+       :selector => @bac.remote_resource_id.to_s } ) %>
+
+    <p><%= f.label :start_date, "Initial Start Date" %><br>
+    <%= text_field_tag :start_date, @start_date, :class => "datepicker" %>
+
```
MontrealSergiy, line 49: Is it mandatory? Can you mark it somehow? Or, maybe default to current day. Time could be 5 min ahead of the current or so

prioux: When scheduling, yes, start_at is mandatory

prioux: The calendar picker will already default to today.

PR #1395, opened 2024-05-23
Added support for subset copying

prioux:
Includes a new BAC for copying with unregister at the end.
Subset copying works only for two DP types, SSH and S3,
and is only available through the dispatcher API.

Just read the code, point out any problem that is visible by reading it.

### DIFF
--- a/BrainPortal/app/models/data_provider.rb
+++ b/BrainPortal/app/models/data_provider.rb
@@ -392,11 +392,11 @@ class DataProvider < ApplicationRecord
     cb_error "Error: provider #{self.name} is not syncable."   if     self.not_syncable?
     cb_error "Error: file #{userfile.name} is immutable."      if     userfile.immutable?
     rr_allowed_syncing!("synchronize content to")
-    if alternate_source_path.present? && ! self.can_upload_directly_from_path?
+    if alternate_source_path.to_s.present? && ! self.can_upload_directly_from_path?
       cb_error "Provider #{self.name} does not support direct uploads."
     end
     final_status = 'InSync'
-    final_status = 'ProvNewer' if alternate_source_path.present? && self.can_upload_directly_from_path?
+    final_status = 'ProvNewer' if alternate_source_path.to_s.present? && self.can_upload_directly_from_path?
     SyncStatus.ready_to_copy_to_dp(userfile, final_status) do
       final_status == 'ProvNewer' ? # same test as the 'if' two lines above
         impl_sync_to_provider(userfile, alternate_source_path) :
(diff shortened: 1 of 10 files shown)


BrainPortal/app/models/data_provider.rb
```diff
@@ -635,6 +635,7 @@ def cache_collection_index(userfile, directory = :all, allowed_types = :regular)
       if userfile.is_a? FileCollection
         if directory == :all
           entries = Dir.glob(userfile.name + "/**/*")
+          #entries = Dir.glob(userfile.name + "/**/*", File::FNM_DOTMATCH).reject { |p| p.ends_with?("/.") || p.ends_with?("/..")}
```
natacha-beck, line 638: Any reason to have this line commented ?

BrainPortal/app/models/s3_flat_data_provider.rb
```diff
@@ -448,6 +456,28 @@ def rsync_emulation(src_fileinfos,dst_fileinfos) #:nodoc:
...
+    return fileinfos if patterns.blank?
+    patterns = Array(patterns).map(&:to_s)  # e.g. "dir1/dir2/*"
+    fileinfos.select do |fi|
+      finame  = Pathname.new(fi.name)       # "Collection/dir1/dir2/file"
+      fifirst = finame.each_filename.first  # "Collection"
+      next true  if finame.to_s == colname  # "Collection" == "Collection"
+      next false if fifirst     != colname  # should never happen
+      rest = finame.relative_path_from(colname).to_s  # "dir1/dir2/file"
+      selected = patterns.any? { |pat| File.fnmatch(pat, rest, (File::FNM_PATHNAME | File::FNM_DOTMATCH) ) }
+      #if selected
```
natacha-beck, line 472: Do you want to keep the debug code ?

BrainPortal/app/models/ssh_data_provider.rb
```diff
@@ -202,6 +210,7 @@ def impl_provider_collection_index(userfile, directory = :all, allowed_types = :
       if userfile.is_a? FileCollection
         if directory == :all
           entries = sftp.dir.glob(provider_full_path(userfile).to_s, "**/*")
+          #entries = sftp.dir.glob(provider_full_path(userfile).to_s, "**/*", File::FNM_DOTMATCH)
```
natacha-beck, line 213: Should it be in remove ?

natacha-beck (approved):
Alll, just look to the comment about commented code.

BrainPortal/app/models/data_provider.rb
prioux: While I was testing all this I discovered a behavior of CBRAIN I might want to change in the future. But I am not sure. And that line is how the behavior would change.

BrainPortal/app/models/s3_flat_data_provider.rb
prioux: I will remove the debug code once we verified in works correctly.

BrainPortal/app/models/ssh_data_provider.rb
prioux: Same thing about the behavior I was saying earlier.

PR #1396, opened 2024-05-27
OIDC

natacha-beck:
Generalisation of GlobusHelpers to be compliant with another OIDC (Keycloak)


prioux:
This is also obsolete.

PR #1401, opened 2024-07-02
Optimise task binding of local provider folders 

MontrealSergiy:


BrainPortal/app/models/cluster_task.rb
```diff
@@ -112,6 +112,16 @@ def record_cbraintask_revs(caller_level=1) #:nodoc:
     self.addlog("#{subrev.basename} rev. #{subrev.short_commit}",   :caller_level => caller_level + 1)
   end
 
+  # This determines if the task expects to only read its input files,
```
prioux, line 115: Can you do me a favor? The method `file_access_symbol` is defined exactly the same way in two other files, boutiques_portal_task.rb and boutiques_cluster_task.rb, and now you're adding it to cluster_task.rb. I think we should just have one copy directly in cbrain_task.rb, the superclass of everything.

I woudl added just below self.properties at line 266 of cbrain_task.rb

BrainPortal/app/models/boutiques_cluster_task.rb
```diff
@@ -98,6 +98,19 @@ def setup #:nodoc:
...
+  def local_dp_storage_paths
+    return super & input_file_dps.pluck(:remote_dir)
+  end
+
+  # lists the dp involved into input files
+  def input_file_dps
+    file_ids = descriptor_for_setup.file_inputs.map do |input|
+      invoke_params[input.id]
+    end
+    return Userfile.where(id: file_ids).map &:data_provider
```
prioux, line 111: You are returning a list with duplicate data providers in it (e.g. when two input files are on the same DP)

prioux (changes requested):
Minor changes

BrainPortal/app/models/boutiques_cluster_task.rb
MontrealSergiy: I never claimed it returns unique dp(s).

BrainPortal/app/models/cluster_task.rb
MontrealSergiy: This will place the method into the section is called  Task Launch API. Are you sure? To me it is more of utility or core method section.

Or maybe create new subsection, perhaps  'access control' ,

BrainPortal/app/models/boutiques_cluster_task.rb
MontrealSergiy: But now it returns unique dp

BrainPortal/app/models/cluster_task.rb
prioux: Well, there isn't a good place in the code for this method, and you're right it is kind of in a category of its own. It's not really a utility either. Let me look at the file and find a good place for it. Give me 10 minutes.

prioux: Right, so I guess you can put it in the `Utility Methods` section, at the end.

Make sure you maintain three blank lines after the method's final `end` before the banner for the next section, the one called `Useful ID Generators`. My convention is three blank lines just before each banner.

MontrealSergiy:
empty lines corrected

PR #1407, opened 2024-07-09
CBRAIN should not change data provider type arbitrary (in presence of missed or incorrect fields)

### LINKED ISSUE #1406: Missing a mandatory field when creating a new Data Provider should not alter the type
To reproduce try to create a Data Provider, say DataLad type, yet miss a mandatory field (such as name or DataLad url) to trigger form validation errors. Using already existing provider name would work too.

Expected: the type persist on the form
Observed: type changes (on the production to Loris provider, on my machine Carmin provider)

Note: few less important field, such as "Files can be browsed only by members of this project (browsable DPs only)" field seems to lost, but this should rather go to another issues

This form is admin/manager only, yet
an admin user would assume that when he entered data, it persist, or at he least does not get altered.

MontrealSergiy:
Makes sure that CBRAIN does not change new data provider type arbitrary due validation errors, such as supplying already existing name, missing it completely etc fixes #1406

### DIFF
--- a/BrainPortal/app/views/data_providers/new.html.erb
+++ b/BrainPortal/app/views/data_providers/new.html.erb
@@ -56,7 +56,7 @@
     <span title="Type of data provider.">
 
       <p><%= f.label :type, "Type" %><br/>
-      <%= f.select :type, grouped_options_for_select(@typelist, selected_key=@provider.type), { :prompt => "Select Provider Type" }, { :disabled => (params[:action].to_sym == :edit) } %><br/>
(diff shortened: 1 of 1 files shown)


prioux:
How does this work? Shouldn't that raise an exception?

prioux:
Have you tested this?

prioux:
Oh, is it a named parameter? Don't use the name, since we're not all using the same versions of Ruby everywhere.

MontrealSergiy:
corrected

PR #1417, opened 2024-07-30
Only ensure a CbrainFileList read access at task launch, resolves #1057

### LINKED ISSUE #1057: The framework should never try to ensure a CbrainFileList is read-write when launching a task
As a followup to #1054 , we can make a further improvement in the task launch interface: if a selected file is a CbrainFileList, the framework doesn't have to check about read-write permissions not matter how the task has been integrated and no matter who the file belongs to.

MontrealSergiy:
see #1057 for details

### DIFF
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -234,14 +234,19 @@ class TasksController < ApplicationController
 
     # Filter list of files as provided by the get request
     file_ids = params[:file_ids] || []
+    cb_file_ids    = CbrainFileList.where(:id => file_ids).pluck(:id)
+    other_file_ids = file_ids.map(&:to_i) - cb_file_ids
     if @tool_config.try(:inputs_readonly) || @task.class.properties[:readonly_input_files]
       access = :read
     else
       access = :write
     end
-    @files   = Userfile.find_accessible_by_user(file_ids, current_user, :access_requested => access) rescue []
-    if @files.count == 0
-      flash[:error] = "You must select at least one file to which you have write access."
+    cb_files     = Userfile.find_accessible_by_user(cb_file_ids, current_user, :access_requested => :read) rescue []
+    other_files  = Userfile.find_accessible_by_user(other_file_ids, current_user, :access_requested => access) rescue []
+    @files = cb_files + other_files
+    if @files.count != file_ids.count
+      flash[:error] = "Select files to which you have #{access} access."
+      flash[:error] = "You must select at least one file." if @files.count == 0
       redirect_to :controller  => :userfiles, :action  => :index
       return
     end


prioux:
Looks good to me.

MontrealSergiy:
Hi, thanks. Indeed technically feature seems working, and task launch renders task submission forms successfully, there could be problems down the road: task submission, validation, whenever file access is double-checked. (Repeat check servers either as anti-hack or access change support I guess). I think rest of task access check should be modified,  case by case ( or, if you prefer keep code dried, by adding `skip_cbcvs` / `skip_lists` etc param to `find_accessible_userfiles` method, to keep it drier ). 

```
--- a/BrainPortal/app/models/boutiques_cluster_task.rb
+++ b/BrainPortal/app/models/boutiques_cluster_task.rb
@@ -85,6 +85,7 @@ class BoutiquesClusterTask < ClusterTask
 
       # Most common situation
       if ! input.list || ! userfile.is_a?(CbrainFileList)
+        # todo check read assess
         make_available(userfile, userfile.name)
         next
       end
diff --git a/BrainPortal/app/models/boutiques_portal_task.rb b/BrainPortal/app/models/boutiques_portal_task.rb
index d360dc148..3a78f4bc1 100644
--- a/BrainPortal/app/models/boutiques_portal_task.rb
+++ b/BrainPortal/app/models/boutiques_portal_task.rb
@@ -307,7 +307,12 @@ class BoutiquesPortalTask < PortalTask
       original_userfiles_ids = self.params[:interface_userfile_ids].dup
       self.params[:interface_userfile_ids] = [] # zap it; we'll re-introduce each userfile.id as needed
       tasklist = original_userfiles_ids.map do |userfile_id|
-        f = Userfile.find_accessible_by_user( userfile_id, self.user, :access_requested => file_access_symbol() )
+      if CbrainFileList.find_by(:id => userfile_id)
+        f = CbrainFileList.find_accessible_by_user( userfile_id, self.user, :access_requested => :read )
+      else
+        f = Userfile.find_accessible_by_user( userfile_id, self.user, :access_requested => :read )
+      end
+
 
         # One task for that file
         if (! f.is_a?( CbrainFileList ) || input.list) # in case of a list input, we *do* assign it the CbFileList
@@ -406,7 +411,7 @@ class BoutiquesPortalTask < PortalTask
         next if isInactive(input)
         userfile_id = invoke_params[input.id]
         next if userfile_id.blank?
-        userfile = Userfile.find_accessible_by_user(userfile_id, self.user, :access_requested => file_access_symbol())
+        userfile = Userfile.find_accessible_by_user(userfile_id, self.user, :access_requested => :read)
         next unless ( userfile.is_a?(CbrainFileList) || (userfile.suggested_file_type || Object) <= CbrainFileList )
         [ input, userfile ]
     end.compact
@@ -481,7 +486,6 @@ class BoutiquesPortalTask < PortalTask
   # Raises an exception for the input parameter name if the parameter's value
   # is not adequate.
   def sanitize_param(input)
-
     name = input.id
     type = input.type.downcase.to_sym # old code convention from previous integrator
 
@@ -542,7 +546,7 @@ class BoutiquesPortalTask < PortalTask
           next nil # remove bad value
         end
 
-        file = Userfile.find_accessible_by_user(value, self.user, :access_requested => file_access_symbol()) rescue nil
+        file = Userfile.find_accessible_by_user(value, self.user, :access_requested => :read) rescue nil
         unless file
           params_errors.add(invokename, ": cannot find userfile (ID #{value})")
           next nil # remove bad value
```

MontrealSergiy:
I am a bit unsure with files that have cbcsv extension.

MontrealSergiy:
@prioux I pushed a few changes to make sure that the task with read-only list can pass the validation

prioux:
As I review this right now, there are three changed files in the pull request.

1. the changes to the tasks_controller.rb seem fine (as I already reviewed them) and I think they are the only necessary changes needed;
2. the changes to boutiques_portal_task.rb seem wrong, because we no longer call the method file_access_symbol at all for any file, which means we could let a user launch a task that modifies files even when they are not supposed to;
3. the changes to the old integrator template BrainPortal/lib/cbrain_task_generators/templates/portal.rb.erb are unnecessary because we no longer support that old integrator code.

prioux (changes requested):
I think the changes to the second file are wrong. Something doesn't seem right

MontrealSergiy:
ok, then I will roll back to previous (approved) version

Issue #1409, opened 2024-07-11
make the list of tools and data sets more noticeable


xmpham:
The` list of tools and data sets available` link on the cbrain portal is not easy to find/locate  ...maybe could we update the font (colour or location)

See attached screenshot 
![image (6)](https://github.com/user-attachments/assets/f2ccc3b9-40f7-49ec-86a6-f98480c99d55)

MontrealSergiy:
@xmpham  I created a pool request, which make the text bolder and also pushed the tool section of dashboard page up. 

Note it would not help users which did not select a project yet. Such users are forwarded to the Project page. If you like to address those Tool menu item can be moved from Resources under cbrain logo. Another solution is put there some slogan like " offering 123 scientific tools and counting!". But I would prefer to hear team opinion before going that far.

xmpham:
Thank you Serge, the full list of tools and data set available under Resources is fine ( to me). But sure, let's ask for the team's opinion.

PR #1420, opened 2024-08-08
Increase visibility of the full tool list link #1409

MontrealSergiy:
see #1409 for details.

### DIFF
--- a/BrainPortal/app/controllers/portal_controller.rb
+++ b/BrainPortal/app/controllers/portal_controller.rb
@@ -53,7 +53,8 @@ class PortalController < ApplicationController
       .where('tool_configs.bourreau_id > 0')
       .where('tool_configs.version_name IS NOT NULL')
       .distinct
-      .pluck 'name'
+      .pluck('name')
+      .sort
 
     @dashboard_messages = Message
       .where(:message_type => 'cbrain_dashboard')
--- a/BrainPortal/app/controllers/portal_controller.rb
+++ b/BrainPortal/app/controllers/portal_controller.rb
@@ -46,14 +46,14 @@ class PortalController < ApplicationController
     @default_data_provider  = DataProvider.find_by_id(current_user.meta["pref_data_provider_id"])
     @default_bourreau       = Bourreau.find_by_id(current_user.meta["pref_bourreau_id"])
 
-    @tool_count =  Tool
+    @tool_names =  Tool
       .joins(:tool_configs)
       .where.not(category: 'background')
       .where('tool_configs.bourreau_id IS NOT NULL')
       .where('tool_configs.bourreau_id > 0')
       .where('tool_configs.version_name IS NOT NULL')
       .distinct
-      .count
+      .pluck 'name'
 
     @dashboard_messages = Message
       .where(:message_type => 'cbrain_dashboard')
--- a/BrainPortal/app/controllers/portal_controller.rb
+++ b/BrainPortal/app/controllers/portal_controller.rb
@@ -46,6 +46,20 @@ class PortalController < ApplicationController
     @default_data_provider  = DataProvider.find_by_id(current_user.meta["pref_data_provider_id"])
     @default_bourreau       = Bourreau.find_by_id(current_user.meta["pref_bourreau_id"])
 
+    @tools = Tool
+               .all
+               .order(:name)
+               .to_a
+               .reject { |t| t.category == 'background' }
+               .select { |t| t.tool_configs.to_a.any? { |tc|
+                 tc.bourreau_id.present?  &&
+                   tc.bourreau_id > 0       &&
+                   tc.version_name.present?
+               }
+               }
+
+    @avaiable_tools = @tools
+
     @dashboard_messages = Message
       .where(:message_type => 'cbrain_dashboard')
       .order("created_at desc")
--- a/BrainPortal/app/controllers/portal_controller.rb
+++ b/BrainPortal/app/controllers/portal_controller.rb
@@ -46,19 +46,14 @@ class PortalController < ApplicationController
     @default_data_provider  = DataProvider.find_by_id(current_user.meta["pref_data_provider_id"])
     @default_bourreau       = Bourreau.find_by_id(current_user.meta["pref_bourreau_id"])
 
-    @tools = Tool
-               .all
-               .order(:name)
-               .to_a
-               .reject { |t| t.category == 'background' }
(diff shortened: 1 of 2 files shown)


BrainPortal/app/controllers/portal_controller.rb
```diff
@@ -46,6 +46,20 @@ def welcome #:nodoc:
     @default_data_provider  = DataProvider.find_by_id(current_user.meta["pref_data_provider_id"])
     @default_bourreau       = Bourreau.find_by_id(current_user.meta["pref_bourreau_id"])
 
+    @tools = Tool
```
prioux, line 49: Use a better query instead of rebuilding all objects

BrainPortal/app/controllers/portal_controller.rb
```diff
@@ -46,6 +46,20 @@ def welcome #:nodoc:
...
+               .to_a
+               .reject { |t| t.category == 'background' }
+               .select { |t| t.tool_configs.to_a.any? { |tc|
+                 tc.bourreau_id.present?  &&
+                   tc.bourreau_id > 0       &&
+                   tc.version_name.present?
+               }
+               }
+
+    @avaiable_tools = @tools
```
prioux, line 61: Why are there two instance variables? Why is this one misspelled?

prioux (changes requested):
See comments

BrainPortal/app/controllers/portal_controller.rb
MontrealSergiy: removing that

MontrealSergiy:
hopefully, better now. If you still see db load issue I can cache.

BrainPortal/app/controllers/portal_controller.rb
```diff
@@ -46,6 +46,15 @@ def welcome #:nodoc:
     @default_data_provider  = DataProvider.find_by_id(current_user.meta["pref_data_provider_id"])
     @default_bourreau       = Bourreau.find_by_id(current_user.meta["pref_bourreau_id"])
 
+    @tool_count =  Tool
```
prioux, line 49: This expression returns a completely different number of tools compared to `current_user.available_tools` (which is used in the page to generate the table). In fact, right now my page says "out of 50" and yet the table lists 161 tools.

I think this should be changed to "@tool_names =" to generate a list of names, not just a count, and then the array_to_table in the page should use that array too.

prioux (changes requested):
See comments

BrainPortal/app/views/portal/welcome.html.erb
```diff
@@ -167,14 +167,14 @@
       <strong>Your current time:</strong> <%= to_localtime(Time.now, :datetime) %><BR>
     </p>
     <p>
-       <strong>Projects you belong to:</strong>  <%= array_to_table(@groups, :table_class => 'simple', :cols => 4) { |g,r,c| link_to_group_if_accessible(g) } %>
+      <strong>Tools available to you (out of <%= link_to @tool_names.count , available_path %>):</strong><br/>
```
prioux, line 170: WHy do we have the same link twice? (line 180 and 174). The one at line 170 is not necessary, it could be be the number.

prioux (changes requested):
Minor

BrainPortal/app/views/portal/welcome.html.erb
MontrealSergiy: That was the point of the PR - To increase visibility of the full tool list. For an average user the link to the full list on line 180 is hard to find, not everyone has a habbit of scrolling welcome page. The one on 170 is much higher and a greater chance to be discovered by users. 

See the issue #1409 for a snapshot and discussion. 

In my opinion repeating important info several time is a good UX practice which we can use more often. But if you uncomfortable with it, maybe it is better just raise line 180 above 174, perhaps with a minor verbiage change?

BrainPortal/app/views/portal/welcome.html.erb
```diff
@@ -167,14 +167,15 @@
       <strong>Your current time:</strong> <%= to_localtime(Time.now, :datetime) %><BR>
     </p>
     <p>
-       <strong>Projects you belong to:</strong>  <%= array_to_table(@groups, :table_class => 'simple', :cols => 4) { |g,r,c| link_to_group_if_accessible(g) } %>
+    <p>
+      <strong> The complete list of curated datasets and all <%= @tool_names.count %> tools on this portal:</strong> <%= link_to 'here.', available_path %>
```
prioux, line 171: Yuk, you get

```
The complete list of curated datasets and all 1234 tools on this portal: here.
```

Also, why not print this on the same line as "Tools available to you:" ?

```
Tools available to you (34 of 845, [link]full list of all tools[/link])
```

prioux: A good web design convention is to never create links with just the value "here" as the anchor. Use a full descriptive sentence about what the user will get when they click the link.

MontrealSergiy: Totally agree. I tried preserve the original text and colon based. I guess I should not goof around and cut and paste the original sentence, perhaps even without ' is here'.

Well, what you proposed is not so bad, though I do not like particularly 

a)  long overloaded sentence; 
b) long text in brackets; 
c) here is still there, I'd rather drop it, otherwise user might got confused, to me here is rather page not necessarily the portal as used by all the user ;  
d) mentioning of curated datasets are disappeared

Not sure maybe 

`Tools available to you ( 34  of   845 [link] of all the configured tools[/link]):`

`Tools available to you ([ link to versions ] 34 ] of  [link] 845 of all the supported [/link]):`

I'd keep then old link below for sake of dataset/

BrainPortal/app/controllers/portal_controller.rb
```diff
@@ -46,6 +46,15 @@ def welcome #:nodoc:
...
     @default_bourreau       = Bourreau.find_by_id(current_user.meta["pref_bourreau_id"])
 
+    @tool_names = Tool
+      .joins(:tool_configs)
+      .where.not(category: 'background')
+      .where('tool_configs.bourreau_id IS NOT NULL')
+      .where('tool_configs.bourreau_id > 0')
+      .where('tool_configs.version_name IS NOT NULL')
+      .distinct
+      .pluck 'name'
```
prioux, line 56: Last change: can we get this list sorted? The list was sorted in the original page, but this query doesn't get it sorted.

prioux (changes requested):
Minor: add `.sort`

Issue #1427, opened 2024-09-05
Missing info on plugins in "about_us" page


prioux:
The about_us page has a section that show the revision numbers of the plugins installed.

The page seems to look for folders named 'cbrain-plugins-XYZ' to identify these pluguins.

However this convention is not necessary, so the page doesn't show the info for plugins named differently.

MontrealSergiy:
@prioux should it show test plugins?  (named xxx**test**xxxx)

PR #1429, opened 2024-09-06
Incomplete plugins list in About Us page #1427

MontrealSergiy:
see #1427

### DIFF
--- a/BrainPortal/app/views/portal/about_us.html.erb
+++ b/BrainPortal/app/views/portal/about_us.html.erb
@@ -37,9 +37,12 @@
   <br>
   <strong>Portal Up Since:</strong> <%= to_localtime(CBRAIN::Startup_LocalTime,:datetime) %> (for: <%= pretty_elapsed((Time.now.localtime - CBRAIN::Startup_LocalTime).to_i) %>)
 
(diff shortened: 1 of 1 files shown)


natacha-beck (approved):

BrainPortal/app/views/portal/about_us.html.erb
```diff
@@ -37,9 +37,12 @@
   <br>
   <strong>Portal Up Since:</strong> <%= to_localtime(CBRAIN::Startup_LocalTime,:datetime) %> (for: <%= pretty_elapsed((Time.now.localtime - CBRAIN::Startup_LocalTime).to_i) %>)
 
-  <% plugins_basenames = Dir.glob(Pathname.new(CBRAIN::Plugins_Dir) + "cbrain-plugins-*")
-                         .map { |x| Pathname.new(x).basename.to_s }
-                         .reject { |x| x =~ /^cbrain-plugins-base$/ } %>
+  <% plugins_basenames = Pathname(CBRAIN::Plugins_Dir)
```
prioux, line 40: This is perfect, except can we just add `.sort` because the filesystem doesn't return the entries in a deterministic order, and we want to be able to compare the reports from one system to another etc.

prioux:
Minor change requested!

MontrealSergiy:
plugins are sorted now

Issue #1433, opened 2024-09-19
Tool and dataset information disappears when moving pointer outside of the (more) hypertext links


bryancaron:
When viewing the tool and dataset listing page at https://portal.cbrain.mcgill.ca/available/ additional details for each tool or dataset are viewable but only when the pointer is directly over the (more) hypertext link.  As soon as the pointer is moved outside of the range of screen covered by (more) then the information disappears.  This in turn makes it difficult for the user to see the information as a persistent source when they are looking to move back and forth between browser windows, for example.

It would be helpful if the information box displayed would be persistent such that it remains visible even when the the pointer position is changed, either within the same browser window as the listing, or to another browser window, until it is explicitly closed by the user.

![Screenshot 2024-09-19 at 2 29 31 PM](https://github.com/user-attachments/assets/8e6222fe-872a-467f-bd94-6b008612fa3e)

The behaviour as seen when clicking on the "Help" button when Launching a task is more akin to a desired behaviour in that it persists until the user closes that particular information box.

Bryan.

prioux:
Yes, these overlay panels are very old and annoying to no end. I created them. They were supposed to be used only for short quick tiny bits of informations (e.g. when hovering over a username, get their city and last login).

For the available page, we need a better more persistent box. I think we have something like that in CBRAIN where a separate panel opens, but I don't know the name of the helper that does that.

MontrealSergiy:
- on task edit page, parameter help opens a full-window large 'modal' box, is it this panel? helper seems to be called `overlay` smth
- alternatively  we can just expand cell on click, like it is done in message list
- pretty often click more cell or page elements is shown (e.g. when browsing nested folders) in file  collection show page

prioux:
Closed by #1449

PR #1440, opened 2024-10-25
Fix tootip overlay closing.

natacha-beck:
Multiple little feature: 

- Unable the closing functionality when leaving `html_tool_tip`.
- Add a `close` at the bottom of the `html_tool_tip`
- Unable the opening of a `html_tool_tip` when mouseenter only if no other are open. 
- It is now possible to open multiple `html_tool_tip` only if click on the link instead `mouseenter`


MontrealSergiy:
Can you refer the issue(s) it is using by id or in Development section?

natacha-beck:
Fix the https://github.com/aces/cbrain/issues/1433

MontrealSergiy (approved):
I do not like that same control ( underlined "more") behaves differently on different pages.
But the PR address the issue as requested.

Maybe we can rename it into 'info', 'card', 'infocard' or '+' ?

MontrealSergiy:
The request recommends to use logic more similar to task help, but the PR still resolves the issue, even though behaviour of 'more' control diverges from the rest of cbrain.

Perhaps, for the case of multiple windows I would add somewhere a button "close all info-cards"

Screenshot from 2024-10-28 14-41-16

natacha-beck:
> I do not like that same control ( underlined "more") behaves differently on different pages. But the PR address the issue as requested.
> 
> Maybe we can rename it into 'info', 'card', 'infocard' or '+' ?

Can you let me know which pageS you are referring to ?

MontrealSergiy:
for example, https://portal.cbrain.mcgill.ca/tools

natacha-beck:
It have the same behaviour on both page ( i just tested, let me know if it is not the case). 

Also I look in the CBRAIN portal and notice some strange behaviour on the tool page:

image

Will look at it,

MontrealSergiy:
Ok, you right. I guess I just do not like that more element.

BrainPortal/app/controllers/nh_users_controller.rb
```diff
@@ -87,9 +87,9 @@ def update
     attr_to_update.delete(:zenodo_sandbox_token) if attr_to_update[:zenodo_sandbox_token].blank?
     attr_to_update.delete(:zenodo_main_token)    if attr_to_update[:zenodo_main_token].blank?
 
-    # Do not update password if user must use globus
-    if user_must_link_to_oidc?(@user)
-      flash[:error] = "You cannot change the password for your account." if attr_to_update[:password].present?
+    # Do not update password if user must use globus (or other oidc)
+    if user_must_link_to_oidc?(@user) && attr_to_update[:password].present?
+      flash[:error] = "You cannot change the password for your account because you should use OpenID." if attr_to_update[:password].present?
```
MontrealSergiy, line 92: not sure how OpenID relates to tootips

natacha-beck:
Replace by #1449

natacha-beck:
Replace by #1449

Issue #1432, opened 2024-09-18
Support "error-codes" in Boutiques integrator


prioux:
A boutiques descriptor can have an optional section where command return codes can be associated with messages, something like:

```json
{
    "dummy": "rest of descriptor here",
    "error-codes": [
        {
            "code": 1,
            "description": "Crashed"
        },
        {
            "code": 12,
            "description": "The program could not find the square root of 144"
        }
    ]
}
```

When entering the post-processing phase of a task, the Boutiques integrator should compare the returned code with that list, if present, and report the message in the task's log.

PR #1443, opened 2024-11-05
Error codes support in boutiques

MontrealSergiy:
see https://github.com/aces/cbrain/issues/1432

I created few test boutiques https://github.com/MontrealSergiy/cbrain-plugins-test/tree/errorcodes. (might fail due to lack of bash in alpine docker, but just add container-free config).

### DIFF
--- a/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
@@ -438,11 +438,6 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
 %   end
       else # Check exit status value
         exit_status = status_file_content.to_i
-        # provide explanations for exit statuses (if any):
-        <% descriptor['error-codes']&.each do |err| %>
-          self.addlog(<%=err["description"].inspect %>)  if <%= err['code'].to_i %> == exit_status  # with escaping
-        <% end %>
-
         unless SystemExit.new(exit_status).success?
           self.addlog("Command failed, exit status #{exit_status}")
 %   if outputs.empty?
--- a/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
+++ b/BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
@@ -440,7 +440,7 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
         exit_status = status_file_content.to_i
         # provide explanations for exit statuses (if any):
         <% descriptor['error-codes']&.each do |err| %>
-          self.addlog '<%=err["description"].gsub("'", '"').gsub(/\R|\\/, ' ')%>'  if <%= err['code'].to_i %> == exit_status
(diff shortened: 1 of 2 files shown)


BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
```diff
@@ -438,6 +438,11 @@ class CbrainTask::<%= name %> < <%= (descriptor['custom'] || {})['cbrain:inherit
 %   end
       else # Check exit status value
         exit_status = status_file_content.to_i
+        # provide explanations for exit statuses (if any):
+        <% descriptor['error-codes']&.each do |err| %>
```
natacha-beck, line 442: Why not `descriptor.error_codes.each` ? and why the `&`

natacha-beck (changes requested):
Please see comment.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
MontrealSergiy: I do not think `error_codes` method exists. Also, key does not necessary exists hence '&'. '&' is more concise than `descriptor['error-codes'] || [ ] ` hack

MontrealSergiy: I think error_codes and other helper method only available for new style boutiques

prioux: Sergiy is right, error_codes as a method is not implemented in the BoutiquesSupport class. We COULD implement it though. If so, then it would have to be part of this PR right here. I will let Sergiy decides if he wants to do so. If yes, keep the code in boutiques_support.rb clean, and insert the method at the proper place please.

prioux: As an ADDDED note, please revert all code changes to the OLD boutiques integrator, we do not maintain it anymore. This means this file right here as I comment this.

prioux: Also I just realized that this means my comment above ("Sergiy is right...") is moot because I had not realized this file here is from the old integrator, which doesn't use BoutiquesSupport.

prioux (changes requested):
I've added a bunch of comments but the end result is:

1. changes to boutiques_cluster_task.rb are all fine
2. revert all changes to the other file, we don't support it.

When (2) is done, I will merge.

BrainPortal/lib/cbrain_task_generators/templates/bourreau.rb.erb
MontrealSergiy: So should I then change it to 'module' to prevent accidental use in old style boutiques?

prioux: Like I said in my comment on the PR main thread, don't modify the old integrator code at all. This PR shoudl only have changes for the file 'boutiques_cluster_task.rb'

MontrealSergiy: Ok removed then. What I meant that when legacy support is not needed anymore, module-based approach is better for documentation and separation of concerns, but perhaps this custom property is quite straightforward and does not need docs.

PR #1449, opened 2024-11-27
Overlay in available page

natacha-beck:
This PR replace #1440

### DIFF
--- a/BrainPortal/app/helpers/rich_ui_helper.rb
+++ b/BrainPortal/app/helpers/rich_ui_helper.rb
@@ -264,16 +264,16 @@ module RichUiHelper
   # Create a link that will open a dialog box.
   #
   # Options:
-  #   title: title of the overlay
-  #   link_text: text of the link
-  #   description: content of the overlay
-  #   size: used to set the max-width of the overlay
-  def closable_overlay_content_link(title, link_text, description="", options = {size: "50em"})
+  #   link_text:      text of the link
+  #   title:          title of the overlay
+  #   description:    content of the overlay
+  #   options[:size]: used to set the max-width of the overlay
+  def closable_overlay_content_link(link_text, title, description="", options = {size: "50em"})
     return "" if description.blank?
 
     link = overlay_content_link(link_text, :enclosing_element => "span") do
       "<h1>#{h(title)}</h1><hr/>".html_safe +
-      "<p style='max-width: #{options[:size]};'>#{h(description)}</p>".html_safe
+      "<p style=\"max-width: #{options[:size]};\">#{h(description)}</p>".html_safe
     end
 
     return link
--- a/BrainPortal/app/helpers/rich_ui_helper.rb
+++ b/BrainPortal/app/helpers/rich_ui_helper.rb
@@ -263,12 +263,14 @@ module RichUiHelper
 
   # Create a link that will open a dialog box.
   #
+  #   +link_text+: clickable text that appears on the page in order to open the overlay
+  #   +title+: title that will show up as a header at the top of the overlay
+  #   +description+: text of the overlay content
+  #
   # Options:
-  #   title: title of the overlay
-  #   link_text: text of the link
-  #   description: content of the overlay
-  #   size: used to set the max-width of the overlay
-  def closable_overlay_content_link(title, link_text, description="", options = {size: "50em"})
+  #
+  #   +size+: maximum height and width of the overlay
+  def closable_overlay_content_link(link_text, title, description="", options = {size: "50em"})
     return "" if description.blank?
 
     link = overlay_content_link(link_text, :enclosing_element => "span") do
--- a/BrainPortal/app/helpers/rich_ui_helper.rb
+++ b/BrainPortal/app/helpers/rich_ui_helper.rb
@@ -261,6 +261,24 @@ module RichUiHelper
     html.html_safe
   end
 
+  # Create a link that will open a dialog box.
+  #
+  # Options:
+  #   title: title of the overlay
+  #   link_text: text of the link
+  #   description: content of the overlay
+  #   size: used to set the max-width of the overlay
+  def closable_overlay_content_link(title, link_text, description="", options = {size: "50em"})
+    return "" if description.blank?
+
+    link = overlay_content_link(link_text, :enclosing_element => "span") do
+      "<h1>#{h(title)}</h1><hr/>".html_safe +
+      "<p style='max-width: #{options[:size]};'>#{h(description)}</p>".html_safe
+    end
+
+    return link
+  end
+
   # Create a button with a drop down menu
   #
   # Options:
--- a/BrainPortal/app/helpers/rich_ui_helper.rb
+++ b/BrainPortal/app/helpers/rich_ui_helper.rb
@@ -266,17 +266,15 @@ module RichUiHelper
   #   +link_text+: clickable text that appears on the page in order to open the overlay
   #   +title+: title that will show up as a header at the top of the overlay
   #   +description+: text of the overlay content
-  #
-  # Options:
-  #
   #   +size+: maximum height and width of the overlay
-  def closable_overlay_content_link(link_text, title, description="", options = {size: "50em"})
+  #
+  def closable_overlay_content_link(link_text, title, description="", size="50em", options = {})
     return "" if description.blank?
 
     link = overlay_content_link(link_text, :enclosing_element => "span") do
-      html = "<div style='overflow: auto; max-height: #{options[:size]}; max-width: #{options[:size]};'>" +
+      html = "<div style='overflow: auto; max-height: #{size}; max-width: #{size};'>" +
       "<h1>#{h(title)}</h1><hr/>" +
-      "<p'>#{h(description)}</p>" +
+      "<p>#{h(description)}</p>" +
       "</div>"
 
       html.html_safe
--- a/BrainPortal/app/helpers/rich_ui_helper.rb
+++ b/BrainPortal/app/helpers/rich_ui_helper.rb
@@ -264,16 +264,20 @@ module RichUiHelper
   # Create a link that will open a dialog box.
   #
   # Options:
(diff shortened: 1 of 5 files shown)


BrainPortal/app/views/portal/_available_datasets_table.html.erb
```diff
@@ -41,7 +41,14 @@
     end
   %>
   <tr>
-    <td><%= overlay_description((name + "\n" + description).strip) %></td>
+    <td>
```
prioux, line 44: Why is this done with an explicit construction of the H1, HR and P elements ? This is cumbersome and will result in empty overlays when the description is empty.

Look at the original code, it was invoking a HELPER. This what what helpers are for: they can make decision about what to render, make adjustements, and even decide not to render anything at all.

With a help such as closable_overlay_content_link() (name I just invented, adjust it please) you could have instead:

```html
<td><%= closable_overlay_content_link(name, description) %></td>
```

which would make all these decisions and not leave empty P elements if there are no descriptions.

BrainPortal/app/views/portal/_available_tools_table.html.erb
```diff
@@ -53,7 +53,14 @@
     end
   %>
   <tr>
-    <td><%= overlay_description((name + "\n" + description).strip) %></td>
+    <td>
```
prioux, line 56: Same here. Like I says, look at the original code you replaced, and ask yourself why is the new code more complicated? That should be an immediate red flag in your mind.

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -179,7 +179,16 @@
     // Tab Bar, div's of type tabs become tab_bars
     // See TabBar class
     loaded_element.find(".tabs").each( function() {
-      $(this).tabs();
+      $(this).tabs(
```
prioux, line 182: Why is this closing code associated with tags instead of a specific CSS class? Tabs are for the tabbing framework, we're modifying the overlays, use distinct appropriate classes. This code here also means the new overlays cannot be used outside of tabs, for instance.

MontrealSergiy:
would be nice to describe the solution (UI change) in a few words

MontrealSergiy (approved):
Looks nicer.

BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -261,6 +261,24 @@ def overlay_content_link(name, options = {}, &block)
...
+  #   title: title of the overlay
+  #   link_text: text of the link
+  #   description: content of the overlay
+  #   size: used to set the max-width of the overlay
+  def closable_overlay_content_link(title, link_text, description="", options = {size: "50em"})
+    return "" if description.blank?
+
+    link = overlay_content_link(link_text, :enclosing_element => "span") do
+      "<h1>#{h(title)}</h1><hr/>".html_safe +
+      "<p style='max-width: #{options[:size]};'>#{h(description)}</p>".html_safe
```
prioux, line 276: single quotes are not canonical in html attributes!

BrainPortal/app/views/portal/_available_datasets_table.html.erb
```diff
@@ -41,7 +41,10 @@
...
+      <%= name %>
```
prioux, line 45: Why do you have `name` in both places?

BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -261,6 +261,24 @@ def overlay_content_link(name, options = {}, &block)
...
   end
 
+  # Create a link that will open a dialog box.
+  #
+  # Options:
+  #   title: title of the overlay
+  #   link_text: text of the link
+  #   description: content of the overlay
+  #   size: used to set the max-width of the overlay
+  def closable_overlay_content_link(title, link_text, description="", options = {size: "50em"})
```
prioux, line 271: I would re order the arguments: `(link_text, title, description)` makes more sense (look at the HTML code when you use the helper)

BrainPortal/public/javascripts/cbrain.js
```diff
@@ -182,6 +182,19 @@
       $(this).tabs();
     });
 
+    // Find tabs in div with class available_tabs in order to close all overlay dialogs
```
prioux, line 185: This seems awfully complicated to find what to close, no? Why not put a class directly on the elements you want closable and make them easier to find that way?

BrainPortal/app/views/portal/_available_datasets_table.html.erb
natacha-beck: I kept what was in the tooltip previously.

MontrealSergiy (changes requested):

MontrealSergiy:
Thanks for renaming 'more', though I am not sure what "(iz)" stands for

MontrealSergiy (approved):
seems ok now

BrainPortal/app/helpers/rich_ui_helper.rb
prioux: Why was this marked as resolved since the order of the arguments is still wrong?

When I create a web page with the helper, it makes more sense to have LINK_TEXT first, which is the thing we see in the page initially, and theN two other things that show up when the overlay opens.

```ruby
   helper( "Lab Info", "Information about the lab", "We Do stuffff here" )
```

BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -261,6 +261,28 @@ def overlay_content_link(name, options = {}, &block)
     html.html_safe
   end
 
+  # Create a link that will open a dialog box.
+  #
+  # Options:
```
prioux, line 266: Adjust this documentation. There are not 4 options. There are 3 ARGUMENTS and 1 OPTION.

Something like

```ruby
#
# +link_text+ is the text that appears in the page, that can be clicked to open the overlay
# +title+ is the title that will show up as a header at the top of the overlay
# +description+ is the text inside the overlay's body
#
# Options:
#
# size is the maximum height and width of the overlay
```

BrainPortal/app/views/portal/_available_tools_table.html.erb
```diff
@@ -53,7 +53,10 @@
...
+      <%= name %>
+      <%= closable_overlay_content_link(name, "(i)", description) %>
```
prioux, line 58: Remember to re-order the args

BrainPortal/app/views/portal/available.html.erb
```diff
@@ -33,7 +33,7 @@ Not all of them are available to all users.
 <% open_tools     = @tools.select { |tool| (tool.group_id == @everyone_gid) && tool.tool_configs.where(:group_id => @everyone_gid).exists? } %>
 <% restrict_tools = @tools - open_tools %>
 
-<%= build_tabs do |tb| %>
+<%= build_tabs({class: 'available_tabs'}) do |tb| %>
```
prioux, line 36: Why this weird syntax? Everywhere else in the view code we use something like:

```erb
<%= build_tabs( :class => 'available_tabs' ) do |tb| %>
```

prioux (changes requested):
See comments.

BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -261,6 +261,30 @@ def overlay_content_link(name, options = {}, &block)
...
+  #
+  #   +size+: maximum height and width of the overlay
+  def closable_overlay_content_link(link_text, title, description="", options = {size: "50em"})
+    return "" if description.blank?
+
+    link = overlay_content_link(link_text, :enclosing_element => "span") do
+      html = "<div style='overflow: auto; max-height: #{options[:size]}; max-width: #{options[:size]};'>" +
+      "<h1>#{h(title)}</h1><hr/>" +
+      "<p'>#{h(description)}</p>" +
```
prioux, line 279: Why is there is single quote in the P element?

BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -261,6 +261,30 @@ def overlay_content_link(name, options = {}, &block)
...
+  # Create a link that will open a dialog box.
+  #
+  #   +link_text+: clickable text that appears on the page in order to open the overlay
+  #   +title+: title that will show up as a header at the top of the overlay
+  #   +description+: text of the overlay content
+  #
+  # Options:
+  #
+  #   +size+: maximum height and width of the overlay
+  def closable_overlay_content_link(link_text, title, description="", options = {size: "50em"})
```
prioux, line 273: This method doesn't allow us to style the link like the normal overlay_content_link() can.

With the roriginal method, I can say

```
overlay_content_link("info", :class => 'action_link') { "description" }
```

and the options are transformed into attributes that are stored in the "A" element. (Try this in the console)

This new helper is hardcoded to only use the size option, and not propagate the rest of the options hash down the call stack.

It should instead extract the size and pass the rest of options down (using dup).

BrainPortal/app/views/portal/_available_datasets_table.html.erb
```diff
@@ -41,7 +41,10 @@
...
+      <%= closable_overlay_content_link("(i)", name, description) %>
```
prioux, line 46: Instead of the string `"(i)"` use `" &#8505; ".html_safe` and provide the class option "action_link" to add that CSS class to the link (assuming the helper has been modified according to my previous comment). It will make the links much nicer.

BrainPortal/app/helpers/rich_ui_helper.rb
prioux: Maye better is to configure the size to be an actual argument:

```ruby
def closable_overlay_content_link(link_text, title, description="", size="50em", options={})
```

That way you don't need to adjust options before passing it along to the other helper.

natacha-beck:
@prioux should be ok now can you have a look.

PR #1452, opened 2024-12-03
[ADD] Boutiques lib to copy an input before running the cmd line

natacha-beck:
Some tools modify the cached version of the input directory.

This can lead to some issues when the input is used multiple times
on the same Bourreau and the cached version is already modified.

To solve this issue, we can create a copy of the directory in the working
directory and use this copy in the command line.

### DIFF
--- a/BrainPortal/lib/boutiques_input_copier.rb
+++ b/BrainPortal/lib/boutiques_input_copier.rb
@@ -23,7 +23,7 @@
 #
 # Some tools modify the cached version of the input directory.
 #
-# This can lead to some issues when the input is used multiple times
+# This can lead to issues when the input is used multiple times
 # on the same Bourreau and the cached version is already modified.
 #
 # To solve this issue, we can create a copy of the directory in the working
@@ -43,18 +43,18 @@
 #       "value-key": "<-i input>"
 #     },
 #
-# and in the `cbrain:integrator_modules` section look like:
+# and in the `cbrain:integrator_modules` section, it looks like:
 #
 #     "BoutiquesInputCopier": {
-#       "file_input_id": {"select_box_selected": boolean, "select_box_hidden": boolean, "select_box_description": string},
+#       "file_input_id": {"checkbox_selected": boolean, "checkbox_hidden": boolean, "checkbox_description": string},
 #      },
 #
-# select_box_selected:    boolean, true if the parent_input should be copied by default
-# select_box_hidden:      boolean, true if the select_box should be hidden in the form,
+# checkbox_selected:    boolean, true if the parent_input should be copied by default
+# checkbox_hidden:      boolean, true if the select_box should be hidden in the form,
 # if set to true the input_file will be copied in all case.
-# select_box_description: string, description of the select_box input
+# checkbox_description: string, description of the checkbox input
 #
-# If "select_box_hidden" is set to false, a checkbox (select_box) will be added to the form
+# If "checkbox_hidden" is set to false, the checkbox will be added to the form
 # to allow the user to choose if the input should be copied or not.
 #
 # If the checkbox is selected, the input will be copied in the working directory replacing the symlink
@@ -63,7 +63,7 @@
 #     apptool input_copy
 #
 # WARNING: Since this module create a copy of the input and remove
-# the original symlink, it make the task not-restartable, it will fail
+# the original symlink, it makes the task non-restartable. It will fail
 # during the setup phase of the restarted task if the input is already copied.
 #
 module BoutiquesInputCopier
@@ -79,22 +79,21 @@ module BoutiquesInputCopier
   ############################################
 
   def descriptor_for_before_form #:nodoc:
-    descriptor_with_special_input(super.dup)
+    descriptor_with_added_checkboxes(super.dup)
   end
 
   def descriptor_for_form #:nodoc:
-    descriptor_with_special_input(super.dup)
+    descriptor_with_added_checkboxes(super.dup)
   end
 
   def descriptor_for_after_form #:nodoc:
-    descriptor_with_special_input(super.dup)
+    descriptor_with_added_checkboxes(super.dup)
   end
 
   ############################################
   # Bourreau (Cluster) Side Modifications
   ############################################
 
-  # For input in +BoutiquesInputCopier+ section,
   def setup #:nodoc:
     # NOTE: This method is not restartable, if the input is already copied
 
@@ -194,7 +193,7 @@ module BoutiquesInputCopier
     override_invoke_params
   end
 
-    # Utility Methods for the Module
+  # Utility Methods for the Module
   ############################################
 
   def create_checkbox_id(parent_inputid) #:nodoc:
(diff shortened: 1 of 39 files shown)


BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,223 @@
...
+  ############################################
+  # Portal Side Modifications
+  ############################################
+
+  # Adjust the description for the input with the mention of
+  # the fake parent directory information
+  #
+  # For each input in BoutiquesInputCopier section create a new entry in the descriptor
+  # that will be a flag to choose if the input will be copied if the checkbox is selected
+  def descriptor_for_before_form #:nodoc:
```
prioux, line 74: Change this to a new method with an argument (descriptor) and then the other methods that you want to override will call it as needed. Don't make framework methods call each other, each shoudl have their own distinct path through the class hierarchy without cross-over.

```ruby
def extension_of_descriptor(descriptor)
   blah blah
end

def descriptor_for_before_form
   desc = super
   extension_of_descriptor(descriptor)
end

def descriptor_for_form
   desc = super
   extension_of_descriptor(descriptor)
end

def descriptor_for_after_form
   desc = super
   extension_of_descriptor(descriptor)
end
```

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,223 @@
...
+  end
+
+  ############################################
+  # Bourreau (Cluster) Side Modifications
+  ############################################
+
+  # For input in +BoutiquesInputCopier+ section,
+  def setup #:nodoc:
+    # Log revision information
+    basename = Revision_info.basename
```
prioux, line 125: what are these variable basename and commit for?

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,223 @@
...
+    commit   = Revision_info.short_commit
+
+    descriptor = self.descriptor_for_setup
+    # First call the main setup
+    return false if !super
+    return true  if descriptor.custom_module_info('BoutiquesInputCopier').blank?
```
prioux, line 131: you write `descriptor.custom_module_info('BoutiquesInputCopier')` twice, at line 131 and 133; extract ina  variable instead

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,223 @@
...
+      next if to_be_copied.blank? || to_be_copied == "0"
+
+      userfile_id = invoke_params[parent_inputid]
+      next if userfile_id.blank?
+      userfile    = Userfile.find(userfile_id)
+      copy_to                                        = userfile.name + "_" + self.run_id.to_s
+      self.invoke_params[parent_inputid]             = copy_to
+      descriptor.input_by_id(parent_inputid)["type"] = "String"
+
+      rsync_cmd = "sleep 10; rsync -a -L --no-g --chmod=u=rwX,g=rX,Dg+s,o=r --delete #{userfile.name.bash_escape}/* #{copy_to.bash_escape}"
```
prioux, line 154: `sleep 10` ?!?!

prioux: Why use a pattern with `/*` ? This will not copy any files that starts with `.` which might still be an miportant part of the files being copied

MontrealSergiy (commented):
- mixed up with overlay pr
- refer the issue, assign pr to yourself, set tags
- the documentation could be explained
- btw, I do not follow why separate input key is needed, cannot we use old one?
- uppercase value key is not compatible with Pierre new ideas regarding command line templating convention

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,243 @@
...
+    extension_of_descriptor(descriptor)
+  end
+
+  ############################################
+  # Bourreau (Cluster) Side Modifications
+  ############################################
+
+  # For input in +BoutiquesInputCopier+ section,
+  def setup #:nodoc:
+    descriptor = self.descriptor_for_setup
```
prioux, line 130: Why set this variable here before invoking super two lines below? Run the super() first.

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,243 @@
...
+#       "optional": true,
+#       "type": "File",
+#       "command-line-flag": "-i",
+#       "value-key": "<-i input>"
+#     },
+#
+# and in the `cbrain:integrator_modules` section look like:
+#
+#     "BoutiquesInputCopier": {
+#       "parent_input_id": {"default": boolean, "hide": boolean, "description": string},
```
prioux, line 49: What do these all mean? What is `default` ? What is `hide`? What does the `description` do?

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,308 @@
...
+  end
+
+  def need_to_copy(copy_value, config) #:nodoc:
+    to_copy    = false
+
+    # Special case if default-value is true and the field is hidden
+    if config["default-value"] && config["hide"]
+      return true
+    else
+      return copy_value.is_a?(TrueClass) ? true : false
```
prioux, line 88: This TrueClass check is meaningless, it works only if copy_value is actually the Ruby `true` ! You should have verified in the console or in irb:

```
CBRAIN main :182 > 1.is_a?(TrueClass)
 => false 
CBRAIN main :183 > "1".is_a?(TrueClass)
 => false 
CBRAIN main :184 > "true".is_a?(TrueClass)
 => false 
```

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,308 @@
...
+  ############################################
+  # Portal Side Modifications
+  ############################################
+
+
+  # Utility Methods for the Module
+  ############################################
+
+  def copy_name(original_name) #:nodoc:
+    "inputcopier_tmp_" + original_name
```
prioux, line 78: I am afraid this will break a bunch of tools.

If a tool requires an input with a particular naming convention, it will not work anymore. E.g. a tool that requires `sub-12345`.

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,308 @@
...
+#
+#     "BoutiquesInputCopier": {
+#       "parent_input_id": {"default": boolean, "hide": boolean, "description": string},
+#      },
+#
+# default:     boolean, true if the parent_input should be copied by default
+# hide:        boolean, true if the input should be hidden in the form
+# description: string, description of the input
+#
+# The final command line will be:
```
prioux, line 56: This explanation block is still incomplete. E.g. what do you say "the final command is `apptool <-i input_copy>` when that is not a command, but a template of a command, with no explanation as to what exactly is different?

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,308 @@
...
+  # Bourreau (Cluster) Side Modifications
+  ############################################
+
+  # For input in +BoutiquesInputCopier+ section,
+  def setup #:nodoc:
+    original_userfile_ids = {}
+
+    basename = Revision_info.basename
+    commit   = Revision_info.short_commit
+    self.addlog("Creating a copy of the input files in BoutiquesInputCopier.")
```
prioux, line 198: I don't like the order of things happening in this method. super is called too low, AFTER the messages that says "Creating a copy", which means that if there is a crash in any other code it looks like this module is responsible. Not onlhy that, it means the "copy" message will be separated with the other addlog messages by all the other messages produced by the other modules.e.g.

```
Creation copy.... (message from this module)
(messages from other modules)
(messages from other modules)
More messages from this module (if any)

File: BrainPortal/lib/boutiques_input_copier.rb [hunk 11]
```diff
@@ -0,0 +1,308 @@
...
+
+  # Adjust the descriptor for the input with the mention of
+  # the fake parent directory information
+  #
+  # For each input in BoutiquesInputCopier section create a new entry in the descriptor
+  # that will be a flag to choose if the input will be copied if the checkbox is selected
+  def descriptor_with_special_input(descriptor) #:nodoc:
+    parent_input_ids = descriptor.custom_module_info('BoutiquesInputCopier')
+
+    inputs      = descriptor["inputs"]
```
prioux, line 143: Use the methods!!

BrainPortal/lib/boutiques_input_copier.rb
```diff
@@ -0,0 +1,293 @@
...
+  # object method revision_info() won't work.
+  Revision_info=CbrainFileRevision[__FILE__] #:nodoc:
+
+
+  ############################################
+  # Portal Side Modifications
+  ############################################
+
+
+  # Utility Methods for the Module
```
prioux, line 82: General comments

* reorder the methods, the small utility methods at the bottom, the important code at the top
* get_descriptor_id() doesn't `get` anything, and the name is way too generic. Use a better name, e.g. `create_id_for_checkbox()`

Issue #1447, opened 2024-11-20
New Boutiques module to validate strings used as filenames


prioux:
Many tools integrated with boutiques descriptors have inputs of type 'String' that are used to provide a output path or basename for whatever files the tool is going to create. These are also used as part of the templates in the 'output' sections to tell CBRAIN what is to be saved int he database.

For exmaple, in the "inputs" section:

```json
        {
            "description": "The name of the folder to store outputs of XCPD processing.",
            "id": "output_dir",
            "name": "output_dir",
            "optional": false,
            "type": "String",
            "value-key": "[OUTPUT_DIR]",
            "default-value": "xcpd_output"
        }
```

and in the "output-files" section:

```json
        {   
            "name": "XCPD Subject Output Directory",
            "id":  "xcpd_output_dir",
            "description": "Subject level output for XCP-D",
            "optional":  false,
            "path-template": "[OUTPUT_DIR]"
        }
```

The prevent problems, right from the start before launching a task, we should validate them a little bit and make sure they respect the rules that CBRAIN impose for userfiles. A new custom modules need to be implemented which would be provided with the IDs of "String" inputs to check against the CBRAIN userfile naming rules, e.g.

```json
        "cbrain:integrator_modules": {
            "BoutiquesOutputNameValidator": [ "output_dir", "other_id", "other_id_2" ]
        }
```

prioux:
The Ruby method to validate names is

```ruby
   ok = Userfile.is_legal_filename?(name_to_check)
```

PR #1453, opened 2024-12-06
[bugfix] block download files of undetermined (yet) size  #1447

MontrealSergiy:


BrainPortal/app/controllers/userfiles_controller.rb
```diff
@@ -1150,18 +1150,31 @@ def download #:nodoc:
...
+    # It takes few moments to calculate its size, so meanwhile we disallow downloading files of unknown,
+    # and potentially prohibitively large size
+    unsized = userfiles_list.detect { |u| u.size.nil? }
+    if unsized
+       flash[:error] = "Size of the file #{unsized.name} is not yet determined." +
+                       " Please try again latter.\n"
+       respond_to do |format|
+         format.html { redirect_to :action => :index }
+         format.json { render :json => { :error => flash[:error] } }
+      end
```
prioux, line 1166: I think the `end` keyword is not indented properly here? (Could be just github) Check please.

prioux: I think the inside of the `if unsized` block is not indented correctly, on second thought.

Issue #1450, opened 2024-11-27
Add pagination to the BackgroundActivity index page


prioux:
I wanted to avoid this, because I though there would never be so many rows at any given time, but users are sometimes using APIs to register thousands of files one at a time (so, creating a BackgroundActivity::RegisterFile object for each).

PR #1455, opened 2025-01-06
Add pagination to the BackgroundActivity index page #1450

MontrealSergiy:


prioux:
Your code is re-introducing the bug that I fixed in ebe490311800c27643f92578c3082f462925de7b by not telling the pagination object whether or not we are serving an API request. Please fix.

MontrealSergiy:
@prioux Fixed. I rebased on master, as the fix is only on master branch.

Issue #1454, opened 2025-01-06
Code cleanup: remove RawData query module


prioux:
In the early days of CBRAIN we used an ActiveRecord library that wasn't as powerful as the current one.

In order to do some direct column extraction, I created a module that provides these two methods:

```ruby
result1 = ar_query.raw_first_column(:attribute_name)   # array like [ att, att, att, ... ]
result2 = ar_query.raw_rows([ :atttribute_1, :attribute_2])   # array like [ [att1, att2], [att1, att2], ... ]
```

These two methods are completely unnecessary anymore, because ActiveRecord has the method `pluck()` which does almost exactly the same thing.

This issue is about:

1. removing completely the module file `BrainPortal/lib/cbrain_extensions/active_record_extensions/relation_extensions/raw_data.rb`
2. searching the entire codebase and replacing the calls to `raw_first_column()` and `raw_rows()` with `pluck()` instead.

Some code adjustment might be necessary given the `raw_xxx()` methods can be invoked without any arguments, unlike `pluck()`.
  
Checklist over the next few days:

* [x] Adjust CBRAIN codebase
* [x] Adjust CBRAIN plugins ( PR https://github.com/aces/cbrain-plugins-neuro/pull/294)
* [ ] Remove the implementations for raw_rows() and raw_first_column()

MontrealSergiy:
I think first step 2 should be done, both for core and external plugins, then some time latter step 1. 

Otherwise, few important tool will cease working ( including civet and DICOM converters)

prioux:
Right. I didn't think about the plugins.

So yes, first we do 2. We commit and deploy. Once we know for sure the CBRAIN codebase *and* the plugins don't use the raw_X() methods anymore, we perform 1.

prioux:
Important!! You missed a few files in the CBRAIN code base:

```
BrainPortal/lib/tasks/cbrain_plugins.rake
BrainPortal/app/views/tasks/_tasks_display.html.erb
BrainPortal/app/views/userfiles/_quality_control_panel.html.erb
BrainPortal/app/views/users/_users_table.html.erb
BrainPortal/lib/cbrain_extensions/active_record_extensions/relation_extensions/raw_data.rb
```

For your next pull request, make sure to update these files AND remove the `raw_data.rb` module

prioux:
Also, make your PR with `master` as the base branch (make sure you start with the latest master too)

PR #1460, opened 2025-01-15
Final removal of RawData module

MontrealSergiy:


MontrealSergiy:
As described in the issue #1454 this is the last step in module removal. It deletes the module itself as well as few missed uses of that module by in ruby templates and rake task

prioux:
Looks good to me. I will first merge the PR in the plugins repo.

Issue #1431, opened 2024-09-16
Destroying a user might leave behinds the WorkGroups they created


prioux:
The callbacks associated with destroying a user record will prevent the destruction if the user still has any files or tasks.

When the files and tasks have been removed, the user record can be destroyed. However, if the user has created WorkGroups, these are left behind.

The destroy operation should clean up these workgroups.

MontrealSergiy:
@prioux And what should happen to files, dataproviders, etc assigned to that group?

Reset to admin or create a special 'nobody' user?

prioux:
If the user is about to be deleted, there should be no files belonging to the user, or data providers in the project, because all files and data providers are deleted first.

If there are files belonging to other users (in case of a shared project) then we need a new rule.

PR #1473, opened 2025-02-18
bugfix : Destroying a user might leave behinds their WorkGroups #1431

MontrealSergiy:
see #1431 for discussion


BrainPortal/app/models/user.rb
```diff
@@ -595,6 +596,12 @@ def destroy_system_group #:nodoc:
     system_group.destroy if system_group
   end
 
+  def destroy_empty_work_groups
```
prioux, line 599: Add :nodoc: please; aside from that it's ready to merge

Issue #1375, opened 2024-02-16
Rip out all the code that implements VirtualMachine processing


prioux:
This is all, at this point, crud in the codebase. It's never used and the use cases have disappeared with modern clusters and containerization.

Hints to find the code:

```
exception class: NoVmAvailableError 
file: scir_cloud.rb
file: scir_amazon.rb
```

MontrealSergiy:
somehow, time by time, I hear of potential commercial partners interested in a demo with amazon VMs, though, for some reason, Magic Castle is typically used

prioux:
This issue has nothing to do with setting VMs as Bourreaux. It's about the subsystem that launches VMs as task handlers.

MontrealSergiy:
Ok, got it

MontrealSergiy:
I guess is Bourraux VM, self-hosted or whichever is technical detail. It seems that scir_amazon lets one run tasks on AWS EC2, given things configured right, official aws-sdk installed etc. 

Well I guess if AWS to be used you would prefer to redo in GCP style, with batch and command line commands instead of REST API

Amazon Scir a dropdown element in a Bourreau config, I guess one to delete that as well cloud credentials from Bourreau model

MontrealSergiy:
actually now it is meta params, so it is easier

PR #1474, opened 2025-02-24
Rip out all the code that implements VirtualMachine processing, older…

MontrealSergiy:
Rip out all the code that implements VirtualMachine processing, older cloud related attributes in bourreaux and tool configs, ec2 etc..  see #1375 for discussion


prioux:
I think you forgot to add the migration that leads to schema revision 20250224221218

prioux:
You also forgot to create a data migration to clean up old ApplicationRecord that no longer exist

prioux:
You have the scema migration but still don't have a data migration

BrainPortal/db/schema.rb
```diff
@@ -447,14 +447,7 @@
...
 
-  create_table "task_vm_allocations", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_unicode_ci" do |t|
-    t.integer  "vm_id"
-    t.integer  "task_id"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-  end
-
-  create_table "tool_configs", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_unicode_ci" do |t|
+  create_table "tool_configs", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
prioux, line 450: The tool_configs table's definition has changed; this happens when your dev DB is different from the one we use for the standard codebase. Just revert this line (450) with a manual commit.

prioux:
Please add a data migration for your changes. I think maybe you don't know what it means. You now have two schema migrations, which changes the columns or tables of the database. But your pull request also remove some model classes and the database could still have records associated with the removed classes. This would make the system crash, if at any point those rows were accessed. So the database need also its DATA to migrated. Thus the data migration. This will be a migration that only has a `up` direction, not a `down` direction (though you could still roll it back and it should then do nothing).

prioux (changes requested):
The data migration is still missing, see my previous comment.

Also, the migration that destroys objects in the task_vm_allocations table should be removed from this PR because it's useless.

prioux:
Also, please test your data migration thoroughly by creating objects in the console and running it up and down. Make sure the migration really performs what you expect it to do.

prioux:
Please rebase and please add the missing migration, as I explained

prioux:
This PR looks good now. I will test it and merge it if everything is OK.

PR #1475, opened 2025-02-25
allow autolink group description

MontrealSergiy:
This is to automatically transform all links in a group description into a clickable link. 

The use case is to refer to website(s) where files in group were taken etc.

See https://github.com/aces/cbrain-plugins-neuro/issues/296

### DIFF
--- a/BrainPortal/app/helpers/rich_ui_helper.rb
+++ b/BrainPortal/app/helpers/rich_ui_helper.rb
@@ -52,6 +52,20 @@ module RichUiHelper
     "<span style=\"white-space: pre-wrap;\">".html_safe + description + "</span>".html_safe
   end
 
+  def full_description(description, autolink=false)
+
+    if autolink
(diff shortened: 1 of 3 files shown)


BrainPortal/app/helpers/rich_ui_helper.rb
```diff
@@ -52,6 +52,20 @@ def full_description(description)
     "<span style=\"white-space: pre-wrap;\">".html_safe + description + "</span>".html_safe
   end
 
+  def full_description(description, autolink=false)
```
prioux, line 55: Why do you have two methods with the same name?!?!

prioux (changes requested):
Minor changes

BrainPortal/app/helpers/rich_ui_helper.rb
MontrealSergiy: duplicate is deleted

Issue #1037, opened 2020-10-07
Rename Message attribute group_id to "invitation_group_id"


prioux:
The name is too generic. That group_id is for invitation messages only.

anushri20:
Hi, Can I work on this?

prioux:
Be my guest.

saadkhi:
hey @prioux i want to fix it, can you direct me which file i have to look upon and make changes.

prioux:
@saadkhi This issue requires changes in a bunch of places

1. first you need to create a DB migration to rename the column in the database. The table is "messages" and the column is (obviously) "group_id" right now. Database migrations are created by running a Rails job that creates the template for you, e.g. "rails generate migration RenameGroupIdInMessages", which you then need to fill in the code
2. the controllers (in app/controllers) where the attribute group_id is accessed are messages_controller.rb and nh_messages_controller.rb ; care must be taken to ONLY modify the name of the variable for the Messages objects, given group_id is also used in a bunch of other contexts (thus this issue)
3. the model where the attribute group_id is accessed is app/models/messages.rb ; I think it's pretty obvious there.
4. the view templates where modification probably need to be made too are in app/views/messages and app/views/nh_messages ; again care must be taken to figure which parts refer to the group_id we're changing and which parts are group_id from other models.

arvinder004:
hello @prioux can you guide me with the installation

MontrealSergiy:
Dear @arvinder004 ,

Please find the official installation guide at https://github.com/aces/cbrain/wiki/Setup-Guide.

The guide might be a bit dated, so use Ruby 2.7.2 Please take a look at the CBRAIN Rocky Linux installation section, which is more up-to-date and works with Ruby 2.7. Even though it is specific to Rocky Linux, the listed tips are relevant to Ruby 2.7 and several recent OSs.

arvinder004:
Thankyou

Issue #849, opened 2019-08-09
Strip ANSI sequences from captured task outputs before display.


prioux:
Some programs colorize their outputs, so before sending back the stdout and stderr to be displayed in CBRAIN, let's strip those ANSI sequences. Maybe also let's escape or delete any other non-printable control characters, such as as ^H.

natacha-beck:
Closed by #1496

PR #1485, opened 2025-03-23
Refactor message handling to use invitation_group_id

axif0:
- Updated `MessagesController` and related views to replace `group_id` with `invitation_group_id` for better clarity and functionality. 

- Adjusted `message_params` to permit the new attribute and modified the schema to reflect this change. Ensured consistency across the application by updating references in the `nh_messages_controller` and views.

**Related issue**
#1037


BrainPortal/app/models/message.rb
```diff
@@ -74,7 +75,7 @@ def self.send_message(destination, options = {})
     expiry       = options[:expiry]        || options["expiry"]
     critical     = options[:critical]      || options["critical"]      || false
     send_email   = options[:send_email]    || options["send_email"]    || false
-    group_id     = options[:group_id]      || options["group_id"]
+    group_id     = options[:group_id]      || options["group_id"]      # For invitation messages only.
```
prioux, line 78: So we should rename the variable and the key in the options hash to 'invitation_group_id' and make sure that the change is applied in places where this is called (which should only be when making Invitation objects)

BrainPortal/app/views/messages/new.html.erb
```diff
@@ -44,10 +44,10 @@
      </p>
 
      <p>
-     To users of project: <%= group_select(:group_id,
+     To users of project: <%= group_select(:destination_group_id,
```
prioux, line 47: No changes needed in this file, this group_id is again the group of people to send a message to, not an invitation group ID.

BrainPortal/app/views/messages/new_dashboard.html.erb
```diff
@@ -44,8 +44,8 @@
       ] %>
     </p>
 
-    <% # group_id is not the attribute of the Message object %>
-    <%= hidden_field_tag :group_id, current_user.own_group.id %>
+    <% # destination_group_id is not the attribute of the Message object %>
```
prioux, line 47: This should also stay as "group_id"

prioux (changes requested):
The purpose of the pull request is really to distinguish the attribute "group_id" in the Message object from all other uses of "group_id" in other contexts. Here, group_id is fine, because it's the ID of a group of people that a person wishes to contact.

The group_id attribute in Message is not related at all to this system. It is used by the "group invitation" mechanism, where a message contains the group_id of the Group that people are invited to join. That is why I wanted it renamed.

So the pieces that are missing are in the Invitation model, where the group_id is being actually used to make invitations.

The remaining files that need to be adjusted are in

```
app/models/invitation.rb
app/controllers/invitations_controller.rb
app/controllers/nh_invitations_controller.rb
```

In particular, care should be taken to adjust the active record relationship directive `belongs_to :group` in the model, given by convention it makes a link between tables using "group_id <-> Group". Then in the controller code, the method to go throught he relationship shouls also be adjusted (e.g. line 96 of invitations_controller.rb)

prioux:
As an additional comment: yes I'm fully aware this is a tricky change to get right, we're working with some legacy code that originally was not using 'good' names for the invitation framework.

prioux:
Also @axif0 let me know if you decide not to adjust the PR, I will then make the changes myself because I feel bad about creating the original issue without having described the associated technical challenges.

prioux:
(I just made some edits to my messages because some of the sentences were moved from local code comments to the main review comments, and made no sense anymore)

BrainPortal/db/migrate/20240323000000_rename_message_group_id_to_invitation_group_id.rb
```diff
@@ -0,0 +1,9 @@
+class RenameMessageGroupIdToInvitationGroupId < ActiveRecord::Migration[5.0]
```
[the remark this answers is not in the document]
axif0: > I feel you created it manually instead of using the rails generator

Ya, I created it manually. I’ll regenerate it using the Rails generator to ensure the correct timestamp and update the schema accordingly. Thanks for pointing it out!

BrainPortal/db/migrate/20250324164153_rename_message_group_id_to_invitation_group_id.rb
```diff
@@ -0,0 +1,27 @@
+class RenameMessageGroupIdToInvitationGroupId < ActiveRecord::Migration[5.0]
```
prioux, line 1: All these checks are unnecessary. The migration you had in the previous commit was already perfectly fine. There is no need to check for the columns and create it if needed, because whent he migration is run, it is guaranteed that the columns will be present. Just revert to the what it was:

```ruby
  def up
    rename_column :messages, :group_id, :invitation_group_id
  end

  def down
    rename_column :messages, :invitation_group_id, :group_id
  end
```

prioux:
I'll gave to review it. I didn't realize you updated it following my comments, sorry. I'll have a look tomorrow.

prioux:
@axif0 Also you asked for a suggestion of something to do. The issue https://github.com/aces/cbrain/issues/849 is pretty easy. There's only one file to change, BrainPortal/app/model/cluster_task.rb, in the method capture_job_out_err(). The method reads four files and store the content inside two object attributes. Two of them (cluster_stdout and cluster_stderr need some filtering before store. We need to remove ANSI sequences, these are special byte sequences that can be recognized by matching. E.g. the byte sequence ESCAPE (ascii 27), followed by `[`, `0` and `m` ("\e[0m" is another way of expressing all four bytes) is the code to reset color in a terminal. There are hundreds of such sequences, instead of identifying each one, remove them all in one efficient scan of the file. We probably only need to remove the very common ones, for color, underline, bold, background color, etc. Suggestion:

```
  <esc> '['  <any number of digits or semicolons> <any single letter>
```

This would need to be validated with an official ANSI sequence document.

prioux:
Yep, I got a crash at line 144 of BrainPortal/app/views/groups/show.html.erb

So there are other places that need to be adjusted. Please make sure you can invite users and accept the invitation with dummy users in your dev environment.

prioux:
Also line 43 of invitations_controller.rb

prioux:
Also no invitation messages are created because the parameter for the group ID is set with key :invitation_group_id while in the message.rb code, it is still expected to be :group_id

prioux:
Currently I have three required changes:

```diff
diff --git a/BrainPortal/app/controllers/invitations_controller.rb b/BrainPortal/app/controllers/invitations_controller.rb
index 968b4085a..de2970ae5 100644
--- a/BrainPortal/app/controllers/invitations_controller.rb
+++ b/BrainPortal/app/controllers/invitations_controller.rb
@@ -40,7 +40,7 @@ class InvitationsController < ApplicationController
        return
     end
 
-    already_sent_to = Invitation.where(sender_id: current_user.id, active: true, group_id: @group.id).all.map(&:user_id)
+    already_sent_to = Invitation.where(sender_id: current_user.id, active: true, invitation_group_id: @group.id).all.map(&:user_id)
     @users = current_user.visible_users.where("users.id NOT IN (?)", @group.users.map(&:id) | already_sent_to)
     render :partial => "new"
   end
```

```diff
diff --git a/BrainPortal/app/models/message.rb b/BrainPortal/app/models/message.rb
index 6bd6b7aed..6bf05c252 100644
--- a/BrainPortal/app/models/message.rb
+++ b/BrainPortal/app/models/message.rb
@@ -75,7 +75,7 @@ class Message < ApplicationRecord
     expiry       = options[:expiry]        || options["expiry"]
     critical     = options[:critical]      || options["critical"]      || false
     send_email   = options[:send_email]    || options["send_email"]    || false
-    group_id     = options[:group_id]      || options["group_id"]      # For invitation messages only.
+    group_id     = options[:invitation_group_id] || options["invitation_group_id"]
     sender_id    = options[:sender_id]     || options["sender_id"]
 
     # Affects how 'destination' is interpreted. Default: exclude locked users.
```

and

```diff
diff --git a/BrainPortal/app/views/groups/show.html.erb b/BrainPortal/app/views/groups/show.html.erb
index 325348692..8ee4f254e 100644
--- a/BrainPortal/app/views/groups/show.html.erb
+++ b/BrainPortal/app/views/groups/show.html.erb
@@ -141,7 +141,7 @@
 
   <%
     group_members  = @group.users
-    open_invites   = Invitation.where(sender_id: current_user.id, group_id: @group.id, active: true).to_a
+    open_invites   = Invitation.where(sender_id: current_user.id, invitation_group_id: @group.id, active: true).to_a
   %>
 
   <%= show_table(@group, :as => :group, :header => 'Members', :edit_condition => @group.can_be_edited_by?(current_user) && (!current_user.has_role?(:normal_user) || group_members.count > 1 || open_invites.count > 0 )) do |t| %>
```

axif0:
Thank you so much for the detailed explanation! @prioux  I really appreciate your guidance. I'm still learning Ruby and getting familiar with the codebase, so apologies if I miss anything along the way.

prioux:
I think this PR is ready now. I will do a revision.

axif0:
Thank you for the changes. @prioux

prioux:
I'm going to merge as soon as the automated tests are finished.

PR #1489, opened 2025-03-31
Improved task deletion

prioux:
When some tasks have no workdirs but do have archive, we spawn a new BackgroundActivity to delete them. It runs on the Portal.

### DIFF
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -800,9 +800,15 @@ class TasksController < ApplicationController
 
           # MASS DELETE
           if operation == 'delete'
-            # Two sublists, to optimize the delete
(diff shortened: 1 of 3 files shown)


prioux:
@MontrealSergiy @natacha-beck Please have a quick look. No need to run it if you don't want to. I tested it with all four possible task types already multiple times

1. tasks with no workdirs and no archives (deleted immediately by the portal)
2. tasks with no workdirs but with an archive (BackgroundActivity::DestroyTaskWithoutWorkdir object created)
3. tasks with workdirs and with an archive in it (BackgroundActivity::DestroyTask)
4. tasks with workdirs and no archives (also BackgroundActivity::DestroyTask)

An easy way to test is to boot portals and bourreau then find the PID of the Activity workers and kill them. That way you can see the BACs objects beeing created, you can inspect them, and nothing gets deleted. For case 1 I commented out the `t.destroy` in the controller.

PR #1492, opened 2025-04-01
Scrna frame

MontrealSergiy:
this PR an adjustment to recent Html viewer feature, it would offer better html support within  scRNA and FRMI collections


prioux:
Looks good to me.

PR #1496, opened 2025-04-04
Remove ANSI Sequences from cluster_stdout/cluster_stderr

axif0:
Remove ANSI sequences from cluster task stdout and stderr outputs for cleaner logging

Using regex, `/\e\[[0-9;]*[a-zA-Z]/` ANSI escape sequences that control color `\e[31m`, underline `\e[4m` , bold `\e[1m`, background color `\e[41m` .

Related Issue: 
#849 

CC: @prioux


axif0:
I don't find `new task` option in frontend. So that I can actually try with colored/styled output as admin userType. 

image

prioux:
Yep, that's pretty straightforward. Will merge next week.

Issue #1436, opened 2024-10-04
Successful file update message in presence of user error


MontrealSergiy:
Sometimes, a user might attempt to rename a file and inadvertently enter a name that already exists. After submission, the message "File updated successfully" appears, even though the original name is preserved. An appropriate error message or warning would tell user the rename failed.

PS. Rename can be performed either from the file table page (file list) or individual file page (file 'show' page).

MontrealSergiy:
This technically fixes the issue — the error message is now generated.

However, until the file page reload the file still appears with a renamed filename in the table.

prioux:
This fix should go in:

https://github.com/aces/cbrain/blob/00076c82fa94aae43706d48e9cb1673a5eff5f1f/BrainPortal/app/controllers/userfiles_controller.rb#L744

PR #1497, opened 2025-04-06
Enhance error handling for existing userfile names.

axif0:
Added validation to prevent duplicate names and provide user feedback.

Issue: #1436

### DIFF
--- a/BrainPortal/app/models/data_provider.rb
+++ b/BrainPortal/app/models/data_provider.rb
@@ -721,7 +721,10 @@ class DataProvider < ApplicationRecord
                           :user_id => userfile.user_id,
                         )
                         .first
-    return false if target_exists
+    if target_exists
+      userfile.errors.add(:name, "already exists. Please choose a different name.")
+      return false
+    end
     cache_erase(userfile)
     SyncStatus.ready_to_modify_dp(userfile) do
       if impl_provider_rename(userfile,newname.to_s)


MontrealSergiy:
I tested and this fixes the issue https://github.com/aces/cbrain/issues/1436 - the error message is now generated.

However, when rename is performed on the file list page, within the table, rather than file page, the file still appears with a renamed filename in the table, until it is. It would be great to fix this as well. 
Screenshot from 2025-04-07 09-50-25

prioux:
Sorry guys, I'm closing this pull requests because this change is not going the right way. The DataProvider class is not a  place to add errors in a completely different data model (userfiles, in this case). What needs to be done is verify the true/false return status at the places where the provider_rename() method is called. The original code with its simple "return false" was actually correct.

prioux:
Ok I am re-opening this pull request because I read the code too quicky.

It seems the original code ALREADY had many places where messages were 'passed' back to the caller by adding errors to the userfile model. That, to me, is really not correct as far as semantics go, but I guess it's old code and at the time I didn't know better.

I'll have to review this more carefully. For the moment, please leave this aside.

prioux:
Just to expand further: before the rename operation even starts, the userfile is presumed to be valid in every way. It doesn't seem right to add "errors" to it if we decide to not rename it, because after all, it is still perfectly valid with the original name.

prioux:
Uh. I'm not even the person who added the userfile.errors.add() code in data_provider.rb...

https://github.com/aces/cbrain/commit/6634d3df

This was way back, 14 years ago.

prioux:
OK I'm closing this issue, I added a note about where to fix this problem in the original issue, #1436 

It's definitely not within the scope of the data provider class to inform users about the failures to rename a file: it's whatever tried to rename that should handle that.

PR #1498, opened 2025-04-07
Initial implementation of a CPU quota framework

prioux:
A big part of the existing DiskQuota system was re-used. The controller and models were renamed and extended.

This is what I implemented in the past 4 days. Currently, this framework is operation except for the part where quotas are actually enforced. So when deployed, we can create and query quotas, but no user will be prevented from running tasks if their quotas are exceeded.

I will add later the enforcement, because I want to see how the quota objects behave in the real system.

Note that this PR contains two important migrations; if you are switching branches back and forth, remember to rollback the two migration before switching to a branch the doesn't support CPU quotas.

For the moment, I'd like you guys to just review the changes (there are lots of them, but it's also because several files were renamed). Using a git diff command with the "--follow" option will allow you to see the PR without seemingly having entire files appear and disappear. Check the git man pages for more info.


MontrealSergiy:
Few first comments

a) When a quota is submitted with errors, it’s better to stay on the edit page so that the admin is encouraged to correct the errors.

b) In the context of quotas, terms like "past week" (or "past month") can be ambiguous and confusing. Why should a past week, which has already ended, affect the user’s ability to launch? "7-day period" or "weekly allowance" could be clearer, even without specifying whether it relates to the past or recent 7 days. It’s fine to keep "past week," but it would help to explain its meaning in fine print to assist rookie admins in understanding the system's functionality. Since we're not the only ones hosting cbrain, admins of other installations could have a good understanding without diving into the codebase.

c) `limits_are_reasonable` could be renamed to `sanity_check`, "quota_sanity_check," or "quota_is_meaningful" The term "reasonable" usually implies a limit large enough to serve a meaningful purpose (for example, 1 second is not reasonable allocation for nonscientific pipeline).
d) not sure why only one disk quota attribute is provided - migration thing?

prioux:
For a), the show_table helper doesn't seem to have an option to render already in the 'editing' state. It would be convenient. In the meantime, it's not that big of a problem.

For d) I nullify the Disk quota attribute that is not meaningful for CPU quotas (line 268 of cpu_quota.rb) and I also nullify the CPU two quota attributes that are not meaningful for Disk quotas (line 177 of disk_quota.rb)

I'm not going to address b) and c), there's nothing wrong there.

MontrealSergiy:
I think user should see at his cpu quota, even if it is not Bourreau specific.

prioux:
@MontrealSergiy If they can't that's a bug. Users are supposed to be able to see their quota, and also the amount of time they used that applied to them. That's how I envisioned it. Does it not work for you?

MontrealSergiy:
some shows, user/bourreau specific value shows, but not for user without bourreau (i.e. default)

prioux:
@MontrealSergiy Yep, found where that happened, and I just fixed that. See

https://github.com/aces/cbrain/pull/1498/commits/54a70ef49b807e88cdbc7f0884813f7b5197e6f7

Thanks.

MontrealSergiy (approved):
generally, seems working

PR #1501, opened 2025-04-11
Added ability to manage BAC workers from ibc

prioux:
The 'ibc' utility built into the console can now
start and stop BackgroundActivity workers.


prioux:
This PR now includes an admin-only web interface to control the BACs and Bourreaux and Workers etc.

PR #1504, opened 2025-04-17
BTQ tasks can propagate number of cores to Scir

### LINKED ISSUE #1472: Add support for "cpu-cores" from Boutiques descriptor in all SCIR
A boutiques descriptor has a field where the author can provide suggestions for the resources needed to run the tool. It looks like this:

```json
    "suggested-resources": {
        "cpu-cores": 2,
        "ram": 24,
        "walltime-estimate": 25000
    },
```

(RAM is in GB, walltime is in seconds)

Currently, the Boutiques task integrator will read the RAM and walltime values and pass them to the appropriate SCIR layer for the cluster. But the cpu-cores property is not used.

I suggest extending the integrator to also take the cpu-cores property into account. A method at the task level would return the number, so that it can be overridden too, as necessary.

The SCIR libraries also need to be updated to support it (well, I maybe just the SLURM and UNIX ones as a starting point).

prioux:
Support only for SLURM. Closes #1472

### DIFF
--- a/BrainPortal/app/models/boutiques_cluster_task.rb
+++ b/BrainPortal/app/models/boutiques_cluster_task.rb
@@ -416,6 +416,16 @@ class BoutiquesClusterTask < ClusterTask
     nil
   end
 
+  # Returns a number of cores needed for the job; nil means to not
+  # ask for anything particular.
+  def job_number_of_cores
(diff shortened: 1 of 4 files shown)


prioux:
(why did I misspell the title?!?)

PR #1505, opened 2025-04-22
Confirm frame

MontrealSergiy:
I talked about this during the today's (April 22) meeting


MontrealSergiy:
I will reopen when fixed

MontrealSergiy:
fixed and reopened with a cup on memory. I think it can be even smaller, about a dozen or two of users and files, but not entirely sure. Message me on slack if need to discuss it

PR #1508, opened 2025-04-28
ToolConfigurator, fixes #1422

### LINKED ISSUE #1422: Create a container builder and configurator task
Part of a CBRAIN admin job involves configuring new tools, or new tool versions, with new apptainer containers.

The ToolConfig object for a new tool (or not version of an existing tool) often needs to be configured with better values than the default ones (where the tool belongs to admin only, and has no environment variables or cluster parameters, etc). It's a bit of a pain to enter all the config parameters. Often, we just need to copy them from a similar tool, or from a previous version of the same tool.

For the container, CBRAIN has the ability to dynamically build a apptainer image out of a docker image name, and cache it, but a better mechanism is to pre-build the container and store it one of CBRAIN's data providers. Then the ToolConfig can be given the ID of this image directly, and there will never be a need to build the container on any of the clusters.

I'm thinking we could do all that with a new CBRAIN task, a Tool configurator helper. The admin could select an existing blank ToolConfig that they want to configure, and a previous ToolConfig to use as a base, and the task would copy the necessary config parameters. At the same time, if the ToolConfig specifies a docker image name that isn't already in CBRAIN, it would build it, save it to a DP and replace the docker name with the ID of the new apptainer image.

prioux:
Admin-only interface to help speed up the process of setting up a ToolConfig, including copying values from existing ones, and building an Apptainer image.

### DIFF
--- a/BrainPortal/app/helpers/select_box_helper.rb
+++ b/BrainPortal/app/helpers/select_box_helper.rb
@@ -287,6 +287,7 @@ module SelectBoxHelper
   # [tool_configs] the array of ToolConfig objects used to build the select box. Defaults to all tool configs
   #                  accessible by the current_user.
   # [allow_offline] by default the offline tc will be disabled. If this option is set to true the tc will be selectable
+  # [repeat_toolnames] the name of the tool will be repeated in each entry in the selection box, in front of the version name
   #
   # The selection box will partition the ToolConfig objects by 'categories', where there
   # are three such categories:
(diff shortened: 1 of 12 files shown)


MontrealSergiy:
I have exception when visiting bourreau list page `SSH master's "user"  is  not a simple identifier.`

prioux:
That has nothing to do with this PR. It means you configured a Bourreau with a bad remote_user value. Fix it in the console or by opening the 'show' page of each of your bourreaux in turn.

MontrealSergiy:
ok, it was working before (on master) though

MontrealSergiy:
Note that some ports in production nil, will it work?

PR #1510, opened 2025-05-05
New plugins symlinks

### LINKED ISSUE #816: Improve Ruby search path scanning at boot time
When a CBRAIN install has lots of plugins, the search path for Ruby code can get quite big, instead of being constant. This is because we add on path per userfile file type, as seen in `config/application.rb`:

```ruby
    # CBRAIN Plugins load paths: add directories for each Userfile model
    config.eager_load_paths += Dir[ * Dir.glob("#{config.root}/cbrain_plugins/installed-plugins/userfiles/*") ]
```

One reason is that each directory there contains more than just a ruby model, it might contain views and other local support code.

We need to find a way to join all of this together in some sort of single structure, maybe by creating better symlinks in the `installed-plugins` subtree. We'd need to enhance the rake task `cbrain:plugins:install:all`

This is normally not much of a problem, except on VERY VERY slow filesystems, where at boot time the scanning of each directory to find each model file can make the entrire process excruciating. Do we even need to improve this, if the fault is really outside of CBRAIN?

prioux:
This is a significant improvement in the way plugins are integrated.

```
The structure of the "installed_plugins" folder was
changed slightly to flatten the subfolder for userfiles.

Previously, we had something like:

installed-plugins/userfiles/text_file/text_files.rb -> (plugin_loc)

and now we have:

installed-plugins/userfiles/text_file.rb -> (plugin_loc)

This means the boot code for the Rails app no longer globs
into a large set of folders to find its Ruby code files, speeding
it up.

But it also means I had to create a new "views" section
to hold the symlinks to the userfiles's "views" subfolders.

The rake task to installed and cleanup the links has been
adjusted, you WILL need to run 'rake cbrain:plugins:install:all'
to adjust all the symlinks. (And even maybe :clean:all first).

The helper method that builds the userfile view paths has also
been updated.
```

Fixes #816

### DIFF
--- a/Bourreau/config/application.rb
+++ b/Bourreau/config/application.rb
@@ -17,8 +17,8 @@ module CbrainRailsBourreau
     config.eager_load_paths += Dir["#{config.root}/lib"]
     config.eager_load_paths += Dir["#{config.root}/lib/cbrain_task_generators"]
 
-    # CBRAIN Plugins load paths: add directories for each Userfile model
-    config.eager_load_paths += Dir[ "#{config.root}/cbrain_plugins/installed-plugins/userfiles/*" ]
+    # CBRAIN Plugins load paths: where userfiles defined by plugins are located (as links)
+    config.eager_load_paths += Dir[ "#{config.root}/cbrain_plugins/installed-plugins/userfiles" ]
 
     # CBRAIN Plugins load paths: add lib directory for standalone Ruby files
     config.eager_load_paths += Dir["#{config.root}/cbrain_plugins/installed-plugins/lib"]


MontrealSergiy:
It might increase risk of file name intersection among different types?

prioux:
Not really. Since the file names (e.g. xyz.rb) are for Ruby classes, and there can only be one Ruby class of name "Xyz", there can't be two "xyz.rb" files anyway anywhere within the system.

prioux:
I found a few problems with this as I am testing some more. Hang on.

prioux:
So this has been tested on a production server (!) that was idle, and it seems to be right.

One happy consequence of the change is that in development mode, the Rails app won't load any of the model file until it's actually needed by a Ruby statement. You can test it in a console (in dev mode) with a statement like this:

```ruby
in_memory = Userfile.descendants.map(&:to_s)
```

This will show the models currenly loaded; those were loaded most likely because a database table refer to them.

Pick a Userfile subclass that is not listed but you know exist (e.g. XMLFile) and just type the class name in the console:

```
XMLFile
 => XMLFile id: integer, name: string, size: decimal, user_id: integer, ...
```

Then if you run the "in_memory" statement again, you should see the class has been added to the list.

PR #1522, opened 2025-06-12
hide public projects from invite by project #1499

### LINKED ISSUE #1499: When inviting users to a project, the 'quick list based on other work project' is bad
When inviting users, we get a huge panel with checkboxes for all users; at the bottom a utility panel with other projects are listed, and this panel seems to include 'public' projects, which makes no sense. They should not be shown there.

This form has always been horrible, and we should consider changing it to be more like the NH interface version.

MontrealSergiy:


MontrealSergiy:
That could be closed. Pierre already changes the form to follow NH style

Issue #841, opened 2019-07-24
Use new show table context helper to merge new and show pages


prioux:
Now that #771 and #772 has given us the ability to link a set of show table into a single form, we should use this to get rid of all the new.html.erb forms we have and replace them with show.html.erb.

Example:

bourreaux/new.html.erb contains a lot of duplicated comments and inputs when compares to bourreaux/show.html.erb

Views to clean:

* access_profiles
* bourreaux
* custom_filters (? might be tricky)
* data_providers
* groups
* signups
* sites
* tools + tool_configs (? also special)
* users

Issue #754 and PR #755 have already updated each individual show_tables to receive the form builder object.

PR #1526, opened 2025-06-17
[FIX] form when creating a new entry

natacha-beck:
Add a little JS script in order to mimic the click event in case of a new record. 

This avoid the user to click explicitly on the Edit link if he are already here to create a new one.

### DIFF
--- a/BrainPortal/app/views/shared/_show_table_content.html.erb
+++ b/BrainPortal/app/views/shared/_show_table_content.html.erb
@@ -1,5 +1,5 @@
 
-<fieldset id="<%= tb.uniq_id || 0 %>" >
+<fieldset <%= 'class="open_on_display"' if tb.is_for_new_record? %> >
 
   <legend><%= tb.header %>
     <% if tb.editable? %>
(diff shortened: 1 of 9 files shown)


BrainPortal/app/views/shared/_show_table_content.html.erb
```diff
@@ -48,3 +48,12 @@
 
 </fieldset>
 
+<% if tb.object.new_record? %>
```
prioux, line 51: The script tag shoudl specify the language.

prioux:
This code crashes as soon as I try to view these show pages:

1. a bourreau
2. a portal

The expansion of the info box fails for:

1. new tool configs
2. new access profiles

prioux:
Also, what happens on a page where there are multiple independent show_table(), or on a page where there are multiple show_tables but all linked by the same show_table_context ?

prioux:
And it just happens that show_table_context is used in access profiles, and in tool configs.

prioux:
See also #841

prioux:
I'll re-open this so we can have a base for making the fixes.

PR #1533, opened 2025-07-04
[FIX] CPU quota type

### LINKED ISSUE #1528: Adjust the type of the "max_cpu" columns for the Quota model
The quota table has three columns which contains some maximum amounts of CPU time, in seconds.

The columns are currently declared in the schema as ints, which are 32 bits, with a maximum of 4 billion (if unsigned) or 2 billion (if signed).

2 billion seconds is not enough for some quotas (it's about 63 years).

Create a migration to change the types of all three columns to `decimal`.

natacha-beck:
Close [#1528](https://github.com/aces/cbrain/issues/1528)


MontrealSergiy (approved):

MontrealSergiy:
you can set metadata  assignees, labels, type in the issue and the PR

MontrealSergiy:
Just picking : ideally, reverse migration should raise exception if any of values are out of range. Though probably not worth the effort since we just introduced cpu quota

PR #1560, opened 2025-09-30
shorten worker's log for db disconnect

MontrealSergiy:


prioux (changes requested):
This code has error and has not been tested.

PR #1564, opened 2025-10-23
[FIX] when a group is deleted tc group should be adjusted

### LINKED ISSUE #1563: deletion of a project may cause tool page exceptions
the tool config page is correctly shows Project access (Deleted/nonExcisitng) but some tool page bug cases exceptions

I'd just add some safe navigation

And, perhaps delete 'c'  and add hyphen between non and Existing

I am not entire sure yet, as safe navigation might obfuscate the meaning of the availability check methods: the usual CBRAIN logic the admin can see whatever he likes. Perhaps, for admins the group-free configs should be shown in tools index, yet, say in grey or red colour. Not sure though that preservation of the access logic and consistency justifies the effort - perhaps could be some follow up issues.

natacha-beck:
Fix #1563 

When a group is deleted, tool_config associated with this group should be adjusted with an existing group (as it is already done for other object like tool for example)

### DIFF
--- a/BrainPortal/app/models/group.rb
+++ b/BrainPortal/app/models/group.rb
@@ -171,10 +171,14 @@ class Group < ApplicationRecord
       .map    { |a| a.name }
       .reject { |a| a == :resource_usage } # we never modify the resource usage records
     objlist = group_has_many_model_list.inject([]) { |list,modsym| list += self.send(modsym) }
+    # Add tc to the objlist
+    ToolConfig.where(:group_id => self.id).each {|tc| objlist.push(tc)}
     user_id_to_own_group_id = {}
     objlist.each do |obj|
       next if obj.is_a?(ResourceUsage) # again, just to be sure
-      own_group_id = user_id_to_own_group_id[obj.user_id] ||= obj.user.own_group.id
+      own_group_id = obj.is_a?(ToolConfig) ?
+          user_id_to_own_group_id[obj.tool.user_id] ||= obj.tool.user.own_group.id :
+          user_id_to_own_group_id[obj.user_id] ||= obj.user.own_group.id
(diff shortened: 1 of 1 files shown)


BrainPortal/app/models/group.rb
```diff
@@ -171,10 +171,14 @@ def reassign_models_to_owner_group #:nodoc:
...
       .reject { |a| a == :resource_usage } # we never modify the resource usage records
     objlist = group_has_many_model_list.inject([]) { |list,modsym| list += self.send(modsym) }
+    # Add tc to the objlist
+    ToolConfig.where(:group_id => self.id).each {|tc| objlist.push(tc)}
     user_id_to_own_group_id = {}
     objlist.each do |obj|
       next if obj.is_a?(ResourceUsage) # again, just to be sure
-      own_group_id = user_id_to_own_group_id[obj.user_id] ||= obj.user.own_group.id
+      own_group_id = obj.is_a?(ToolConfig) ?
+          user_id_to_own_group_id[obj.tool.user_id] ||= obj.tool.user.own_group.id :
```
prioux: This is a really complex expression...

You have

```ruby
gid = test_for_tc ? 
                    cacheexpression1 ||= extract1 
                  : 
                    cacheexpression2 ||= extract2
```

Couldn't this be all rewritten simpler?

prioux:
I remembered that ToolConfigs don't all have associated Tools, e.g. for a config that defines defaults for an entire Bourreau.

So the code in this PR would crash at `obj.tool`.

I rewrote the adjustement in my own commit in https://github.com/aces/cbrain/commit/a3e871d19ad8e5522ccb026d583c569bf0565672

So I am closing this PR. It was the right place to make the change, but the mods here aren't sufficient.

Issue #1569, opened 2025-11-05
Add automatic bindmounts support in BTQ descriptors


prioux:
Sometimes we want to use a container from a Boutiques descriptor, but the scripts and executables inside the containers are not exactly appropriate for the integration on the CBRAIN side. E.g. we would prefer to have a nice python or bash wrapper that would be the command to run in the descriptor, which would invoke the stuff inside the container. Right when we need to add such a wrapper, we build a new container and just add the wrapper. This a problem for maintenance of both the container's code (under control of the tool's developers) and the wrapper's own maintenance (presumably under our control).

We could make things way simpler by storing the wrapper in some sort of "bin" directory in the plugin's structure, and having in the descriptor a special directive that says CBRAIN must mount that folder in the container at run time. The directive might go into the "custom" section.

Calling it "bin" might not be general enough, for we could need also libs and static data etc. Namming to be discussed.

See also #1568 which would probably be a prerequisite feature.

prioux:
I'm tempted to just establish a convention such that a plugin that has a folder named "cbrain_mnt" means that folder would be mounted as "/cbrain_mnt" within the containers.

Issue #1568, opened 2025-11-05
Add support for bindmounts in ToolConfig


prioux:
Our ToolConfig objects have an attribute where the admin can configure filesystem overlays, using a text box with e.g.

```
file:/some/path/thing.sqs # explicit squashfs file on remote side
userfile:12345            # ID of userfile containing the squashfs file
```

I'd like us to support simple bindmounts too; the ClusterTask class already has support for these when they come from local DPs, adding more would be convenient. I suggest a specification like:

```
bindmount:/bourreau/path/to/thingie:/containerized/path/to/thingie
```

PR #1573, opened 2025-11-13
Bindmount

prioux:
Adds the ability to specify bindmounts in a ToolConfig, for tasks launched with Apptainer.

The configuration is performed in the "Singularity Overlays" text area, where the admin can add lines like:

```
bindmount:/data1/user-data/prioux/misc/HbcdQcContainer:/inside/hbcd1
bindmount:/data1/user-data/prioux/misc/HbcdQcContainer:/inside/hbcd2:ro
```

There are two commits in this PR:

1. A draft work by Sergiy, commit 3d7816d
2. My work finishing the feature, commit bdb1723

### DIFF
--- a/BrainPortal/app/models/tool_config.rb
+++ b/BrainPortal/app/models/tool_config.rb
@@ -460,6 +460,7 @@ class ToolConfig < ApplicationRecord
     return tc_paths if parent1.basename.to_s != "boutiques_descriptors" # check plugins convention
     plugin_path = parent1.parent             # "/path/to/RailsApp/cbrain-plugins/plugin-name"
     plugin_containerized_dir = plugin_path + "container_mnt" # special plugins folder to mount
+    return tc_paths if ! File.directory?(plugin_containerized_dir.to_s)
     tc_paths << [ plugin_containerized_dir.to_s, container_mountpoint + ":ro" ]
     return tc_paths
   end


prioux:
For this pull request, I intend to merge it using a merge commit instead of a squash commit, so that each developer's work is properly credited.

BrainPortal/app/models/cluster_task.rb
```diff
@@ -2382,6 +2382,12 @@ def singularity_commands(command_script)
       sing_opts
     end
 
+    # (4b) Add tool config bindmounts, specified in 'overlay'
+    esc_local_bindmounts = bindmount_paths.inject("") do |sing_opts,(from_path,cont_path)|
```
MontrealSergiy: I'd replace `esc_local_bindmounts` with `esc_bindmounts` `esc_misc_local_bindmounts`, `esc_other_bindmounts`, `esc_config_bidmounts` ... to reduce confusion with `esc_local_dp_bindmounts`, not showstopper though

prioux:
Sure test it, it's easy

prioux:
I just used the diagnostics tool with a plain linux container for it

BrainPortal/app/models/cluster_task.rb
prioux: You say "to reduce confusion with esc_local_dp_bindmounts".

But there is no `esc_local_dp_bindmounts` to be confused with.

There is an "esc_local_dp_mountpoints" which is not the same.

MontrealSergiy: never mind, those long names confused me

prioux:
I am going to add right now the commit for the OTHER feature too, since it's only one method changed. Good design.

prioux:
This PR now closes TWO tickets: #1568 and #1569

natacha-beck:
Tested on my side and it is a really nice addition.

MontrealSergiy:
BTW boutiques support creating files in workfolder (see 'file-template' attribute), though it does not have execution right and only suits for small text files, not executables or libraries

prioux:
> BTW boutiques support creating files in workfolder 'file-template', though it does not have execution right and only suits for small text files, not executables or libraries

I don't understand what that means. Can you explain?

prioux:
In `output-files` ? Never heard of that before. Will check the BTQ doc.

Issue #1577, opened 2025-11-21
CBRAIN skips validation for single file tools


MontrealSergiy:
CBRAIN support a number of validations for input file types, names. However some users discovered a way to circumvent this validation for single-input descriptors.

Many tools ( e.g eegml_alpha_rest ) have just only mandatory input file and users might skip selecting file in the task input form. 

Such task could be submitted and executed, but validation of input file type, name pattern, etc is not performed

This even if type is not appropriate (e.g. BIDS rather than BIDS subject) user can submit task

MontrealSergiy:
When adding ability to "Launching task array by selecting multiple files" for boutiques with a single file input, UI single file is treated similirly to file list, so much of validation is ignored (see commit 1bb303a7df926d9b8c674b502c39db1315cd350e )

I recall there could be issues with validation of long file list with default regex engine, so not yet entirely sure should be it addressed for all files in a list, but defenitely if there is not batch processing user should be required to select single file.

prioux:
I have a prototype code change for this:

```diff
diff --git a/BrainPortal/app/controllers/tasks_controller.rb b/BrainPortal/app/controllers/tasks_controller.rb
index dad744a06..d183dbbd5 100644
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -394,7 +394,7 @@ class TasksController < ApplicationController
       @task.errors.add(:tool_config_id, 'is on an Execution Server that is currently offline')
     end
 
-    unless @task.errors.empty? && @task.valid?
+    if @task.errors.any? || ! @task.valid?
       flash.now[:error] += messages
       initialize_common_form_values
       respond_to do |format|
@@ -406,15 +406,39 @@ class TasksController < ApplicationController
     end
 
     # Create a bunch of tasks and launch them, either in background or in foreground
-    tasklist,messages = create_tasklist_from_initial_task(@task)
+    tasklist,tl_messages = create_tasklist_from_initial_task(@task)
+
+    # Revalidate everything for all the tasks in the new tasklist
+    # This is kind of costly, but we have no choices if special validation
+    # modules have been loaded for after_form().
+    # As soon as any task in the task list is not valid, we stop checking and
+    # return to the form.
+    af_messages = ""
+    tasklist.each do |task|
+      af_messages += task.wrapper_after_form
+      if task.errors.any? || ! task.valid?
[...]

PR #1578, opened 2025-11-25
for single-input tasks, do not treat input file singleton list #1577

MontrealSergiy:
This to restore name and type validation for tools with a single input file.

This handles only single file case, not a bunch of file (some of which could be list). The former might be done  by overwriting input sanitizer rather than after form handler. Yet I am not sure for the case of long lists as regex validation is not linear with standard regular expression engine, could create issue


prioux:
This fix doesn't solve the problem that validations provided by modules are not applied when launching task arrays.

prioux:
This is not at all what we need.

PR #1581, opened 2025-11-30
Spurious params framework

prioux:
This PR adds a new controller directive, 'spurious_params_ban_ip'. In a controller it would look like this:

```ruby
spurious_params_ban_ip :credits => [], :about_us => []
spurious_params_ban_ip :credits  # same as with an explicit => []

spurious_params_ban_ip :login => [ :username, :password ]

spurious_params_ban_ip :new   => []
```

You provide to it the names of actions in the current controller, and a list of allowable parameters. If a GET or POST request is made with a bunch of unknown parameters, we know it's not a genuine client, so internally CBRAIN will ban the client's address. This entire mechanism is meant to be useful only for public-facing pages, not for for anything that happens to a logged-in client.

This PR comes with a bunch of pre-defined ban rules for all the public-facing pages.

Currently, to allow so wiggle room, banning happens when at least 3 spurious params are detected.

FOR CODE REVIEWERS: In a dev environement, you'll never end up actually banning yourself, so feel free to try as many weird params as you want. The messages about them will appear in the Rails logs.

### DIFF
--- a/BrainPortal/app/controllers/nh_sessions_controller.rb
+++ b/BrainPortal/app/controllers/nh_sessions_controller.rb
@@ -35,7 +35,7 @@ class NhSessionsController < NeurohubApplicationController
                          :request_password => [], # NYI
                          :send_password    => [], # NYI
                          :orcid            => [ :code ],
-                         :nh_oidc          => [ :code, :state ]
+                         :nh_oidc          => [ :code, :state, :scope ] # we don't use scope though
 
   def new #:nodoc:
     @orcid_uri      = orcid_login_uri()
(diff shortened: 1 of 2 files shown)


BrainPortal/lib/request_helpers.rb
```diff
@@ -66,5 +67,82 @@ def ban_ip(message)
...
+    return true if ok_params.nil? # no list for this action, so make no further verifs
+
+    # Find out the ones we don't expect
+    extended_keys = ok_params.map(&:to_s) + STANDARD_PARAMS_THAT_DO_NOT_TRIGGER_BANS
+    spurious_keys = params.keys.map(&:to_s) - extended_keys
+    return true if spurious_keys.empty? # everything is good
+
+    # Prepare message
+    num_spurious = spurious_keys.size
+    first_four   = spurious_keys[0..3].join(", ")
```
natacha-beck: Why you do not use `MINIMUM_NUMBER_OF_SPURIOUS_PARAMS` here ?

prioux: Because it's not the same. The constant is the threshold that triggers bans or not. The 0..3 here is just what we report, which is always the first four params no matter what the threshold is.

PR #1614, opened 2026-04-17
[ADD] Boutiques resource manager

natacha-beck:
Added a ResourceManager for Boutiques tool, provides a functionality to specify suggested resource according to option selected by the user. 

>   "BoutiquesResourceManager": {
>      "step": {
>         "demuxalot":
>         {
>             "cpu-cores":         6,
>             "ram":               15,
>             "walltime-estimate": "02:00:00"
>         },
>         "demuxlet":
>         {
>             "cpu-cores":         1,
>             "ram":               150,
>             "walltime-estimate": "04:00:00"
>         }
>      }
>   }

In this case if the step input is selected with demuxlet the task will asked for 1 cpu, 150Go of ram and 4 hours of walltime. 

More than one input can be specified

### DIFF
--- a/BrainPortal/lib/boutiques_resource_manager.rb
+++ b/BrainPortal/lib/boutiques_resource_manager.rb
@@ -1,0 +1,129 @@
+
+#
+# CBRAIN Project
+#
+# Copyright (C) 2008-2026
+# The Royal Institution for the Advancement of Learning
+# McGill University
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+#
+# For example:
+#   "BoutiquesResourceManager": {
+#      "step": {
+#         "demuxalot":
+#         {
+#             "cpu-cores":         6,
+#             "ram":               15,
+#             "walltime-estimate": "02:00:00"
+#         },
+#         "demuxlet":
+#         {
+#             "cpu-cores":         1,
+#             "ram":               150,
+#             "walltime-estimate": "04:00:00"
+#         }
+#      }
+#   }
+#
+# If step input is selected with:
+#     - demuxalot it will set the ram requirement to 15
+#     - demuxlet  it will set the ram requirement to 150
+#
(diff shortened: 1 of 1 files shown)


BrainPortal/lib/boutiques_resource_manager.rb
```diff
@@ -0,0 +1,129 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+#
```
prioux, line 23: The documentation block is missing a paragraph of intro, something like

```
This module provides a tool with the ability to selectively specify resources parameters (walltime, cpus, memory) based on the values of some of the input fields.

etc
```

BrainPortal/lib/boutiques_resource_manager.rb
```diff
@@ -0,0 +1,129 @@
...
+      # Final lookup of resources structure
+      val   = val.to_s
+      res   = resources[val] # {"cpu-cores": 6, "ram": 40, "walltime-estimate": "2-4:00:00" }
+      res
+    end.compact
+  end
+
+  private
+
+  def walltime_in_second(walltime) #:nodoc:
```
prioux: The method should be `walltime_in_seconds` (plural).

BrainPortal/lib/boutiques_resource_manager.rb
```diff
@@ -0,0 +1,129 @@
...
+    # nil or already in second
+    return 0             if !walltime
+    return walltime      if walltime.is_a?(Integer)
+    return walltime.to_i if walltime =~ /\A\d+\z/
+
+    # Should be in format "DD-HH:MM:SS" of "HH:MM:SS" other cb_error
+    unless walltime =~ /\A(?:(\d+)-)?([0-1]\d|2[0-3]):([0-5]\d):([0-5]\d)$\z/
```
prioux, line 136: I would use the `x` modifier with that regex to make things more clear

prioux (changes requested):
Minor changes

BrainPortal/lib/boutiques_resource_manager.rb
prioux: Also instead of using Regexp.last_match, which can change if people insert code between the regex and the time you use it, I would save the match information directly in a variable:

```ruby
matchinfo = walltime =~ /
   long
   regex   # days
   here      
   with    # minutes
   comments
   /x

cb_error "blah" if ! matchinfo

days, hours, mins, secs = matchinfo[1..4]
```

prioux: Ah you know what, don't change the regex. Just do the mods to the documentation block at the top and rename `walltime_in_seconds` and I'll merge.

PR #1624, opened 2026-04-24
list support for Boutiques Input Subdir maker #1623

### LINKED ISSUE #1623: boutiques dir maker with file list
sometimes one might wish put files in the same folder using 

            "BoutiquesInputSubdirMaker": {
                "previous_results": {
                    "dirname": "ReconstructedDir"
                }
            }

Yet if `previous_results` input is list,  invocation file the task setup fails, as in the invocation build by CBRAIN "previous_results" is a string rather than array. It is extension, but also marked as a bug as IMHO the module should at least log some warning explaining that file lists are not supported.

MontrealSergiy:


MontrealSergiy:
found bugs revoking

PR #1629, opened 2026-05-15
Boutiques portal task

### LINKED ISSUE #1628: Boutiques disabled input issue in before_form and after_form
In [ensemblex](https://github.com/aces/cbrain-plugins-genetics/blob/master/boutiques_descriptors/ensemblex_0.0.2.json) for now each input is set to optionnal. It allow a user to launch a task without any input files. 

The issue is the following if they are setup to optional false the before_form and after_form are too strict. 

before_form will not allow the user to go to the form for example if a user select only one wordir it will raise an error. 

In after_form only inactive input are ignored not disabled one.

natacha-beck:
Fixed #1628 

By bypass the verification in before form in case of disabled input files. 

By adding logic to after form in order to have the same patterns as it is for isInactive.


BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -113,10 +115,17 @@ def before_form
         "#{iname} #{ioptional}\n"
       }.join("")
 
+    # List of file that can be disable (requested or optional)
```
prioux, line 118: So this new piece of code is complicated to read. First, the name of the variable. `input_files_can_be_disabled` is not really about how they CAN be disabled, but how they COULD. And it's a true/false value that turns true if any input file is listed as a potentially disabled in any other input. I'm not convinced that useful until we actually verify if it's the case.

For the code, I see a complicated structure of three `any?` embedded in each other. That's unreadable. Also, the input being scanned is accessed using the `[]` operator instead of the method. Here's my attempt to rewrite it, but please double check it too:

```ruby
all_file_ids = needed_files_ids | optional_files_ids # union of all IDs
input_files_can_be_disabled = descriptor.inputs
  .select { |input|  input.value_disables.present? } # subset of inputs that disables other inputs 
  .any?   { |input| (input.value_disables.values.flatten & all_file_ids).present? } # are there at least one of those that disables any of the file inputs
```

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -113,10 +115,17 @@ def before_form
...
+        (needed_files_ids & disabled_inputs).any? || (optional_files_ids & disabled_inputs).any?
+      end
+    end
+
     # return "Warning: you selected more files than this task requires, so you won't be able to assign them all."
     # Not available in case of descriptor qualified to launch multiple task
-    if num_in_files < num_needed_inputs || num_in_files > num_needed_inputs+num_opt_inputs
-      if (! descriptor.qualified_to_launch_multiple_tasks? || num_in_files == 0)
+    if num_in_files < num_needed_inputs || num_in_files > num_needed_inputs + num_opt_inputs
+      if (! descriptor.qualified_to_launch_multiple_tasks? || num_in_files == 0) && !input_files_can_be_disabled
```
prioux, line 128: Again, like I said in my previous comment, we're not going to get the message if there are any POTENTIAL disabling, bot ACTUAL disabling. I don't think that's right.

BrainPortal/app/models/boutiques_portal_task.rb
```diff
@@ -469,6 +480,27 @@ def isInactive(input)
...
 
+  def isDisabled(input)
+    descriptor = self.descriptor_for_after_form
+    descriptor.inputs.any? do |other_input|
+      # An inactive input cannot disable an other one
+      next false if isInactive(other_input)
+      # Disabled via disables-inputs
+      # Eg:
+      #    "disables-inputs": [ "mask" ]
+      (other_input.disables_inputs || []).include?(input.id) ||
```
prioux, line 491: So lines 491 and 500 make up this huge expression like this:

```ruby
(other_input.disables_inputs || []).include?(input.id) || ((other_input['value-disables'] || {})[invoke_params[other_input.id].to_s] || []).include?(input.id)
```

This is unreadable, even with the comments. You are currently within a Ruby `any?` loop, so please use the loop shortcuts to evaluate each clause separately and `break out` of the loop early. e.g.

```
next true if (other_input.disables_inputs || []).include?(input.id)

current_value = invoke_params[other_input.id].to_s
disable_map   = other_input.value_disables || {}
disable_ids   = disable_map[current_value] || []

next true if disable_ids.include?(input.id) # the "next true if" is not really needed here
```

You could also shortcut even earlier with a variant:

```
disable_map   = other_input.value_disables || next false
disable_ids   = disable_map[current_value] || next false
```

but this is tricky if we ever add more conditions within the loop.

prioux (changes requested):
Review my suggestions, it's possible my code examples have bugs.

PR #1636, opened 2026-06-17
Adjusted naming of files uploaded to Zenodo.

### LINKED ISSUE #1635: some files get numerical suffixes at zenodo
Observed: Some task info and log files get numerical suffixes at zenodo
Expected: At least Json files should keep json extension, so average user can open them in browser or appropriate users, associated to json.

Note: we do not upload task itself, just place some script, info data into text files.

To reproduce upload a task to Zenodo. Most likely some files, such as logs or run info,  would have numerical suffixes. For example `runtime_info-3286892-1.json.983607` . 

One workaround considered before could be to use rename files after upload, though there are multiple user complain regarding that endpoint.

The code that might be responsible is within method `upload_text_data_to_deposit`

It creates temporary files for logs in the `/tmp` folder with random number suffix up to 1 million. 

Perhaps, as workaround a numerical prefix or subdirectory could be created.
 

Not sure, maybe there is a way to fix by changes to upload process

prioux:
It seems the library was finding out about the name for the file by calling a mathod on the file handle, and ignoring the separate name string we provided to it. This change makes sure the tmp files have exactly the names we want; they're put in a temp folder named after the deposit ID. Fixes #1635

### DIFF
--- a/BrainPortal/app/controllers/tasks_controller.rb
+++ b/BrainPortal/app/controllers/tasks_controller.rb
@@ -1275,8 +1275,8 @@ class TasksController < ApplicationController
     userfile.sync_to_cache
     content_path = userfile.cache_full_path
     content_name = userfile.name
-    content_path = create_tmp_tar_for_filecollection(userfile) if is_col
-    content_name = userfile.name + ".tar.gz"                   if is_col
+    content_path = create_tmp_tar_for_filecollection(deposit, userfile) if is_col
+    content_name = userfile.name + ".tar.gz"                            if is_col
 
     # Upload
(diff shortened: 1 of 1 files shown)


MontrealSergiy:
I think this workaround should help. Should I test?

MontrealSergiy (approved):
Logic seems fine.

I tested a Diagnostic task, it worked as expected.

I'd use a random number in folder name for files for folders too, not sure what would happen if several  uploads done in the same time but did not try it

prioux:
@MontrealSergiy Thanks for testing it.

PR #1641, opened 2026-07-07
Enhancement: Implement customizable post-task cleanup policies (#1611)

AviJxn:
Closes #1611  this implements the customizable data preservation and cleanup policies discussed, mapping the state action matrix into two intuitive dropdowns in the task control panel.

### Changes
* **Schema**: Added `success_cleanup_policy` and `failure_cleanup_policy` string columns (defaulting to `'keep_all'` so legacy tasks do not break).
* **Controller**: Added the selection dropdowns to `_control.html.erb` and whitelisted the new attributes in `TasksController`.
* **Model Logic**: Added `should_cleanup_component?` to `CbrainTask` to handle the matrix mapping with `erase_workdir_and_outputs` as the default on success.
* **Background Process**: Created `RemoveTaskInputs` and `RemoveTaskOutputs` background activities both include real time checks against active tasks to safely skip erasing any shared or downstream assets.
* **Worker Lifecycle**: Injected an isolated `enqueue_post_task_cleanup_policies` helper hook into `BourreauWorker` to prevent making the monolithic loop any longer.

### DIFF
--- a/Bourreau/app/models/bourreau_worker.rb
+++ b/Bourreau/app/models/bourreau_worker.rb
@@ -503,6 +503,7 @@ class BourreauWorker < Worker
     end # case 'status' is 'New', 'Data Ready', 'Recover*' and 'Restart*'
 
 
+    enqueue_post_task_cleanup_policies(task)
 
     #####################################################################
     # Task notification section
@@ -602,6 +603,40 @@ class BourreauWorker < Worker
       Message.send_message(User.find(user_id), atts.merge(:variable_text => "Number of tasks blocked: #{count}"))
     end
     return true # means quota is exceeded
+  end
+
+  # Enqueues the post-task cleanup policies for a task that has just finished
+  def enqueue_post_task_cleanup_policies(task)
+    return unless task.status.in?(CbrainTask::FINAL_STATUS)
+
+    if task.should_cleanup_component?(:workdir)
+     BackgroundActivity::RemoveTaskWorkdir.new(
+        :user_id            => task.user_id,
+        :remote_resource_id => task.bourreau_id,
+        :status             => 'InProgress',
+        :items              => [ task.id ]
+      ).save
+    end
+
+    if task.should_cleanup_component?(:inputs)
+      BackgroundActivity::RemoveTaskInputs.new(
+        :user_id            => task.user_id,
+        :remote_resource_id => task.bourreau_id,
+        :status             => 'InProgress',
+        :items              => [ task.id ]
+      ).save
+    end
+
+    if task.should_cleanup_component?(:outputs)
+      BackgroundActivity::RemoveTaskOutputs.new(
+        :user_id            => task.user_id,
+        :remote_resource_id => task.bourreau_id,
+        :status             => 'InProgress',
+        :items              => [ task.id ]
+      ).save
+    end
+  rescue => e
+    Rails.logger.error "Post-task cleanup policies hook crashed for task ##{task.id}: #{e.message}"
   end
 
 end
--- a/BrainPortal/app/models/background_activity/remove_task_inputs.rb
+++ b/BrainPortal/app/models/background_activity/remove_task_inputs.rb
@@ -1,0 +1,48 @@
+#
+# CBRAIN Project
+#
+# Copyright (C) 2008-2026
+# The Royal Institution for the Advancement of Learning
+# McGill University
+#
+
+# Removes the cached inputs of a CBRAIN task safely from the local cache.
+# Must be run on a Bourreau only.
+class BackgroundActivity::RemoveTaskInputs < BackgroundActivity::TerminateTask
+
+  Revision_info = CbrainFileRevision[__FILE__] #:nodoc:
+
+  def process(item)
+    super(item)
+    
+    cbrain_task = CbrainTask.where(:bourreau_id => CBRAIN::SelfRemoteResourceId).find_by(id: item)
+    return [false, "Task not found"] unless cbrain_task
+
+    input_userfile_ids = cbrain_task.params[:interface_userfile_ids] || []
+    
+    input_userfile_ids.each do |userfile_id|
+      userfile = Userfile.find_by(id: userfile_id)
+      next unless userfile
+
+      # Safety check: skip if another active task relies on this exact input
+      file_in_use = CbrainTask.active
+                              .where.not(id: cbrain_task.id)
+                              .any? { |t| t.params[:interface_userfile_ids]&.include?(userfile_id) }
+
+      if file_in_use
+        self.addlog("Skipped cache erase for '#{userfile.name}' (ID: #{userfile_id}) - asset is currently in use.")
+        next
+      end
+
+      userfile.cache_erase
+      self.addlog("Deleted cache of input file '#{userfile.name}' (ID: #{userfile_id}) via post-task cleanup policy.")
+    end
+
+    [true, nil]
+  end
+
+  def prepare_dynamic_items
+    populate_items_from_task_custom_filter
+  end
+
+end
--- a/BrainPortal/app/models/background_activity/remove_task_outputs.rb
+++ b/BrainPortal/app/models/background_activity/remove_task_outputs.rb
@@ -1,0 +1,52 @@
+#
+# CBRAIN Project
+#
+# Copyright (C) 2008-2026
+# The Royal Institution for the Advancement of Learning
+# McGill University
+#
+
+# Removes the cached outputs of a CBRAIN task safely from the local cache.
+# Must be run on a Bourreau only.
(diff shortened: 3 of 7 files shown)


AviJxn:
Hey @prioux  it looks like the CI pipeline is dropping on a strict framework sequence error. The workflow runs rake db:seed before executing pending migrations, which triggers a crash on my new cleanup policy migration (`BrainPortal/db/migrate/20260629232300_add_cleanup_policies_to_cbrain_tasks.rb`). The core implementation files are entirely isolated and clean.

prioux:
@AviJxn That's because your pull request includes a migration, but it doesn't include the effect of that migration on the full schema. In your work environment, when you apply your migration, it will also update the file db/schema.rb which describe the final effect of the DB. That file is what is loaded in CI. During a CI check, no migrations are applied, because the DB is initialized right from the start with the final schema.

BrainPortal/app/models/background_activity/remove_task_inputs.rb
```diff
@@ -0,0 +1,48 @@
...
+    input_userfile_ids = cbrain_task.params[:interface_userfile_ids] || []
+    
+    input_userfile_ids.each do |userfile_id|
+      userfile = Userfile.find_by(id: userfile_id)
+      next unless userfile
+
+      # Safety check: skip if another active task relies on this exact input
+      file_in_use = CbrainTask.active
+                              .where.not(id: cbrain_task.id)
+                              .any? { |t| t.params[:interface_userfile_ids]&.include?(userfile_id) }
```
prioux, line 30: You can't be sure that t.params[:interface_userfile_ids] will contain an array of IDs that are numeric or string. This is a known inconsistent part of CBRAIN. So your comparison with `.include?()` would fail. Generally, when I need to write such code, I turn everything into strings. So I'd end up with something ugly like this:

```ruby
.any? { |t| (t.params[:interface_userfile_ids] || []).map(&:to_s)
            .include?(userfile_id.to_s)
      }
```

BrainPortal/app/models/background_activity/remove_task_inputs.rb
```diff
@@ -0,0 +1,48 @@
...
+    return [false, "Task not found"] unless cbrain_task
+
+    input_userfile_ids = cbrain_task.params[:interface_userfile_ids] || []
+    
+    input_userfile_ids.each do |userfile_id|
+      userfile = Userfile.find_by(id: userfile_id)
+      next unless userfile
+
+      # Safety check: skip if another active task relies on this exact input
+      file_in_use = CbrainTask.active
```
prioux, line 28: You can restrict the search to only the active tasks on the current server, we don't really care if a file is being used on a different one.

BrainPortal/app/models/background_activity/remove_task_inputs.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      userfile.cache_erase
+      self.addlog("Deleted cache of input file '#{userfile.name}' (ID: #{userfile_id}) via post-task cleanup policy.")
+    end
+
+    [true, nil]
+  end
+
+  def prepare_dynamic_items
```
prioux, line 44: This is leftover from another BAC, it shouldn't be here.

BrainPortal/app/models/background_activity/remove_task_outputs.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  Revision_info = CbrainFileRevision[__FILE__] #:nodoc:
+
+  def process(item)
+    super(item)
+    
+    cbrain_task = CbrainTask.where(:bourreau_id => CBRAIN::SelfRemoteResourceId).find_by(id: item)
+    return [false, "Task not found"] unless cbrain_task
+
+    output_userfiles = Userfile.where(task_id: cbrain_task.id).to_a
```
prioux, line 21: This is non-sense; there is no such relation between userfiles and tasks. You can't find the output of tasks using a relation, you need to query the params array of a task.

Task models needs to be extended with a method that returns their output IDs. For BoutiquesTasks, these would be recognized by looking up the keys `_cbrain_output_IDHERE` in params, where IDHERE is the ID of the output-files object in the Boutiques descriptor for the task.

prioux: The fact that this code is submitted and would crash (because `Userfile.where(:task_id => x)` is illegal) tells me you never actually ran your code, did you?

BrainPortal/app/models/background_activity/remove_task_outputs.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      userfile.cache_erase
+      self.addlog("Deleted cache of output file '#{userfile.name}' (ID: #{userfile_id}) via post-task cleanup policy.")
+    end
+
+    [true, nil]
+  end
+
+  def prepare_dynamic_items
```
prioux, line 48: Also spurious.

Bourreau/app/models/bourreau_worker.rb
```diff
@@ -503,6 +503,7 @@ def process_task(task) # when entering this method, +task+ is a partial object,
     end # case 'status' is 'New', 'Data Ready', 'Recover*' and 'Restart*'
 
 
+    enqueue_post_task_cleanup_policies(task)
```
prioux, line 506: All the code including in "bourreau_worker.rb" is misplaced. The role of the BourreauWorker is to manage the life cycle of the task, e.g. how each state is turned into another, but not the actual work being performed during each state.

So the code for cleaning up inputs and outputs should instead be integrated into the class ClusterTask.

BrainPortal/app/models/background_activity/remove_task_outputs.rb
```diff
@@ -0,0 +1,52 @@
...
+# CBRAIN Project
+#
+# Copyright (C) 2008-2026
+# The Royal Institution for the Advancement of Learning
+# McGill University
+#
+
+# Removes the cached outputs of a CBRAIN task safely from the local cache.
+# Must be run on a Bourreau only.
+class BackgroundActivity::RemoveTaskOutputs < BackgroundActivity::TerminateTask
```
prioux, line 11: Again, this is a bad name for that class, we are not removing `outputs`.

BrainPortal/app/models/background_activity/remove_task_inputs.rb
```diff
@@ -0,0 +1,48 @@
...
+# CBRAIN Project
+#
+# Copyright (C) 2008-2026
+# The Royal Institution for the Advancement of Learning
+# McGill University
+#
+
+# Removes the cached inputs of a CBRAIN task safely from the local cache.
+# Must be run on a Bourreau only.
+class BackgroundActivity::RemoveTaskInputs < BackgroundActivity::TerminateTask
```
prioux, line 11: This is a bad name for that class, we are not removing inputs ! We are removing the cached copies of inputs.

BrainPortal/db/migrate/20260629232300_add_cleanup_policies_to_cbrain_tasks.rb
```diff
@@ -0,0 +1,6 @@
+class AddCleanupPoliciesToCbrainTasks < ActiveRecord::Migration[5.0]
+  def change
+    add_column :cbrain_tasks, :success_cleanup_policy, :string, default: 'keep_all'
+    add_column :cbrain_tasks, :failure_cleanup_policy, :string, default: 'keep_all'
+  end
+end
```
prioux, line 6: Missing final NL

BrainPortal/app/models/background_activity/remove_task_inputs.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    [true, nil]
+  end
+
+  def prepare_dynamic_items
+    populate_items_from_task_custom_filter
+  end
+
+end
```
prioux, line 48: Missing final NL

prioux (changes requested):
There are many issues here. See comments.

Also, there are missing elements that I would ideally have liked to have in this PR:

1. provide an abstract CbrainTask method that returns the IDs of outputs; the default would be [] and subclasses would override this, thus giving the framework an easy access to the list of outputs of a task. BoutiquesTasks would fetch from the params array and return the `_cbrain_output_XXX` arrays for instance.
2. the decision about which input file's cache to clean or not would benefit from using the kind of logic we use in the special module BoutiquesInputCacheCleaner ; that module use the timestamps of the synchronization subsystem to make some decisions
3. we could entirely skip cleaning any input or output that are less than 1mb; we don't really care about leaving those in the caches, in the end. The point is to provide cleanup of LARGE pieces of data that are only used once (either as inputs or as produced outputs)

PR #1645, opened 2026-07-09
I18n by chunk

xmpham94:
access profiles page translated in french


natacha-beck:
@xmpham94 
You sould commit on my branch here: https://github.com/natacha-beck/cbrain/tree/i18n_by_chunk

Not on master.

PR #1659, opened 2026-09-17
Added code to prevent floods of email

prioux:
When robots ask for new accounts.


prioux:
Note to reviewers: in development mode, the Rails cache system is disabled, so basically the safety feature never triggers. Which is normal.

To test this, run in production mode. Fill the signup form. Click submit. Check the production logs, and see that eveything is ok. Click back in your browser. Change a field, and submit again. The logs now should have a warning message about duplication and no 'email' was sent. BTW, you need to file the 'system from email address' in your portal too.

